PR #10, opened 2018-03-12
[#155791674] Remove Email Templates management from Super Admin

noremmie:


MothOnMars:
The deletions look reasonable, but we should also remove the super admin routes: https://github.com/GSA/search-gov/blob/master/config/routes.rb#L211

noremmie:
Thanks, @MothOnMars - I missed the routes.

MothOnMars:
LGTM 👍

PR #12, opened 2018-03-15
[#155844997] Set ActionMailer settings from secrets.yml

noremmie:
@MothOnMars I'll summarize my comments in bullet list form.

* we no longer configure ActionMailer with any SMTP settings from MandrillAdapter (which is going away soon)
* we no longer specify any ActionMailer configuration in environment files such as `config/environments/test.rb`
* we are instead pulling ActionMailer configuration from secrets.yml
  * this makes configuring ActionMailer as deployment-friendly as possible because our deployment recipes already generate secrets.yml
  * the only ActionMailer configuration option that doesn't work with this scheme is the `logger` setting which we would have to set in code somewhere if we ever want to use something other than the default
  * see the "Configuration options" section here: http://api.rubyonrails.org/classes/ActionMailer/Base.html
* I tested this extensively on demo3, and everything worked as expected - even the `force_to` behavior (this behavior prevents us from accidentally emailing real users from demo environments - all emails sent from demo environments go to one configured email address)

Here's what I plan on using for the `email` section of `secrets.yml` in production:

```
email:
  action_mailer:
    delivery_method: :sendmail
```

Our production and demo hosts all have a `sendmail` utility that already uses the correct SMTP settings for sending emails via AWS SES. When search-gov is configured this way, it calls out to that `sendmail` utility to do the email sending. I could have also set the delivery_method to `:smtp` and provided our SES SMTP connection config in a `smtp_settings` section, and I successfully tested that style in a demo environment too. I prefer the `:sendmail` way because it's simpler.

The corresponding cookbooks PR is [here](https://github.com/GSA/usasearch-cookbooks/pull/266), and it includes an appropriate update for `secrets.yml.dev`.


MothOnMars:
Overall, LGTM. (Gotta love those PRs that are mostly a sea of red.) My only nit is that there are no tests for `config/initializers/email_interceptor.rb`. I realize that it's based on config/initializers/email_interceptor.rb, which *also* had no tests, so I leave it to your discretion. (I'm also thinking we have cuke tests that use the emailer, so we may be covered by those.)

noremmie:
Thanks, @MothOnMars. I struggled for a while to think of a way to add tests for the new code. I thought maybe I could create some sort of `EmailConfigatron` object that takes in the relevant secrets.yml section and calls the appropriate `ActionMailer::Base` methods. But I still wouldn't be able to test the fact that I passed the correct config values into the `EmailConfigatron`. I'll take your "LGTM" and continue to noodle appropriate initializer tests in the future.

FWIW I tested the bejesus out of the various configuration options in a demo environment, and I'm going to roll this out in an extremely careful manner.

MothOnMars:
@noremmie , I can live with the lack of tests. At least we'll be running the initializer code as part of the app setup, so we'd have some warning if it were going to explode.

PR #16, opened 2018-03-22
James/deep collection to sgov 156126489

jamesisaacs:


app/models/site_search.rb
```diff
@@ -21,3 +20,4 @@ def domains_scope_options
   def populate_additional_results
   end
 end
+
```
MothOnMars: looks like an extra line break snuck in

app/controllers/searches_controller.rb
```diff
@@ -159,6 +159,7 @@ def search_consumer_urls
   end
 
   def docs_search_klass
+    @search_options[:affiliate].search_engine = 'SearchGov' if @search_options[:document_collection]&.too_deep_for_bing?
     return I14ySearch if gets_i14y_results?
```
MothOnMars: Shouldn't this be:
```ruby
def docs_search_klass
  return I14ySearch if gets_i14y_results? || @search_options[:document_collection]&.too_deep_for_bing?
```
?

app/models/site_search.rb
jamesisaacs: Thanks.

app/controllers/searches_controller.rb
jamesisaacs: The search engine needed to be set in addition to sending I14ySearch as the search class.

dawnpm:
@MothOnMars is this ready to go? thank you.

MothOnMars:
LGTM 👍

PR #19, opened 2018-03-29
[#155791760] Transition Mandrill template emails to search-gov ERB template emails

noremmie:


noremmie:
@jamesisaacs this seems like a big change, but it's mostly removing a lot of Mandrill-related stuff and replacing `MandrillAdapter` usage with the equivalent `Emailer` usage.

I also renamed the `new_user_email_notification` template to `user_email_notification` because a similar name change happened in our Mandrill account after we had moved that email into Mandrill.

noremmie:
@jamesisaacs, the codeclimate issue is about the fact that `Emailer` now has 22 methods which is higher than the recommended number of 20 methods. Given the tight deadline on this work and given that the number of methods in that class is largely a function of the number of email templates we use, I would prefer to wait until after this is released to address codeclimate.

PR #22, opened 2018-04-03
Configure Elasticsearch Watchers to send email via SES

noremmie:
@jamesisaacs, here's a rundown of the changes being made in this PR. The search-gov app's `Emailer` class doesn't need to specify any BCC info anymore because the app has been sending all its email via the system `sendmail` utility which already BCCs the correct email address. So I removed the BCC functionality from `Emailer`. `Emailer` still needs to send an email to the "usasearchoutbound@" email address though in the case where the app tries to send an email but can't find the template. So I renamed that to `ADMIN_EMAIL_ADDRESS` which I'm now providing in `config/secrets.yml`.

I updated the `Watcher` class to use the `email` action instead of the `webhook` action. Since I can't configure Elasticsearch Watchers to send email via the system `sendmail` utility, I have to specify the BCC email address in the Watcher itself. So you'll see that I used `Emailer::ADMIN_EMAIL_ADDRESS` for that which is the correct address to use.

I added two `email_templates` records and rigged `Watcher` to load those records to provide email subjects and bodies. These new email templates are just like the other email templates except that they're evaluated by Elasticsarch in order to merge in Watcher values instead of evaluated by ERB.


noremmie:
I've added the `do-not-merge` tag only because I want to merge this myself (after getting a favorable review) in a way that preserves both commits separately.

noremmie:
I forgot to mention one more important thing. After deploying this I plan to run the migration that adds the two new `email_templates` records, and then I'll run something like this to update all the existing Elasticsearch Watchers:

```
Watcher.all.each { |w| ES.client_reader.watcher.put_watch(id: watcher.id, body: watcher.body) }
```

PR #23, opened 2018-04-05
Mct/indexing tweaks

MothOnMars:


spec/support/shared_vcr_setup.rb
```diff
@@ -27,7 +27,6 @@
 
   config.default_cassette_options = {
     record: :new_episodes,
-    re_record_interval: 2.months,
```
MothOnMars: @noremmie , this commit keeps the 2-month interval for specs (which I am still tempted to raise), but removes the re-recording requirement for cucumber tests. This seems safe, as the API calls in the cuke tests *should* all be covered by the specs, and it allows us to manage a smaller number of changes when re-recording is necessary.

app/models/searchgov_url.rb
```diff
@@ -27,7 +27,7 @@ class SearchgovUrl < ActiveRecord::Base
   class SearchgovUrlError < StandardError; end
 
   def fetch
-    self.last_crawled_at = Time.now
+    self.update_attributes(last_crawled_at: Time.now)
```
MothOnMars: This change is intended to prevent race conditions, where multiple processes looking for `unfetched` urls (`"where last_crawled_at = nil"`) might pick up the same record in the time it takes for the fetch process to run (which could be > 30 seconds).

app/models/sitemap_indexer.rb
```diff
@@ -10,21 +10,20 @@ def index
     sitemap = Sitemaps.discover(domain)
     Rails.logger.info "[Searchgov SitemapIndexer] #{log_info.merge(sitemap_entries_found: sitemap.entries.count).to_json}"
     sitemap.entries.each{ |entry| process_entry(entry) }
-    SearchgovUrl.fetch_new
```
MothOnMars: This `fetch_new` business will be completely rejiggered when I build a sane queuing system. I'm removing it from the bulk indexing task and the sitemap indexer because a) it makes those processes run much longer than they should, and b) it runs the risk of multiple processes fetching URLs for the same domains, which could result in  URLs for within a domain being fetched too quickly, and getting us blocked.

PR #24, opened 2018-04-06
[#156559132] Correctly iterate over assets hash

nickmarden:


nickmarden:
@noremmie this is one-half of the reason why assets were being served from the origin server. To be useful this requires the changes in https://github.com/GSA/usasearch-cookbooks/pull/272

nickmarden:
@noremmie I made a [comment](https://www.pivotaltracker.com/story/show/156559132/comments/188049944) about the "top-level" CSS and JS not using the asset host. I did more reading and found out why that was, and I've updated this PR to fix the underlying issue. Please take another look.

In case it's not obvious, the problem was that

`Rails.application.config.action_controller.asset_host = 'something'`

was not working due to load order issues whereas

`ActionController::Base.asset_host = 'something'` 

does work in the initializer. See https://github.com/rails/rails/issues/16209

PR #26, opened 2018-04-09
[#156210698] improve xls parsing performance

MothOnMars:


spec/models/application_document_spec.rb
```diff
@@ -76,5 +76,13 @@
         expect(parsed_content).not_to match(/\uFFFD/)
       end
     end
+
+    context 'when an XLS file contains a million empty rows' do #because that's a thing.
```
noremmie: The fixture file used here only has 124704 rows, not a million rows :trollface:

MothOnMars: :P

PR #28, opened 2018-04-18
[#156201423] restore check for redirections outside domain

MothOnMars:


MothOnMars:
Good call - I rebased, & CC is now happy. (Note - some refactoring of that class is on its way, because I agree w/ CC that it is getting overstuffed...)

PR #29, opened 2018-04-19
[#156888108] Bump newrelic_rpm gem and set up newrelic.yml for dev use

noremmie:
@MothOnMars this change upgrades the `newrelic_rpm` gem to its latest version and updates `config/newrelic.yml` to be more suitable for dev use. That file gets overwritten in production, so I think it makes sense to only provide settings there for a development environment.

I've removed the `ssl` setting because according to the [changelog](https://github.com/newrelic/rpm/blob/master/CHANGELOG.md#v500) and [a recent New Relic blog post](https://discuss.newrelic.com/t/support-for-non-ssl-connections-eol/53570), it's no longer supported, and all connections made to New Relic are now SSL no matter what.

I'll follow up with other stories to get this gem upgraded for all of our apps so that we can make sure those apps are also only talking to New Relic using SSL.

I tested this on demo3 by deploying the app there and then manually enabling `monitor_mode`. I saw app metrics from demo3 in New Relic after that change. I also tested this in my local dev environment and was able to see metrics from that environment in New Relic as well (my "server" shows up there as "69e64100db9d").


noremmie:
@MothOnMars I take back what I said about only providing settings in `newrelic.yml` for a development environment. It turns out that we want to explicitly disable New Relic in the test environment too which I've done in a follow-up commit.

config/newrelic.yml
```diff
@@ -1,10 +1,9 @@
...
-  app_name: <%= Rails.application.secrets.newrelic['app_name'] %>
-  license_key: <%= Rails.application.secrets.newrelic['license_key'] %>
-  developer_mode: <%= Rails.application.secrets.newrelic['developer_mode'] || 'false' %>
-  log_level: <%= Rails.application.secrets.newrelic['log_level'] || 'info' %>
-  monitor_mode: <%= Rails.application.secrets.newrelic['monitor_mode'] ? 'true' : 'false' %>
-  ssl: true
+development:
+  app_name: USA Search
+  license_key: FILL_IN_LICENSE_KEY
+  developer_mode: true
```
MothOnMars: @noremmie , per the NR docs, "Developer mode is removed and no longer supported as of Ruby agent version 4.1.0."

https://docs.newrelic.com/docs/agents/ruby-agent/features/developer-mode

`monitor_mode: true` should be the only necessary line to enable the dev monitoring. That's how I tested locally.

I'm also thinking we might want to set the dev app name to something like `USA Search - DEVELOPMENT` or even `USA Search - YOUR NAME` to avoid any confusion when looking at the metrics.

noremmie:
Thanks, @MothOnMars. I've addressed your feedback in a "squashme" commit.

nickmarden:
@noremmie @MothOnMars before merging this could you please look at my comment on https://github.com/GSA/usasearch-cookbooks/pull/279 - I'm not sure this change is actually necessary.

README.markdown
```diff
@@ -316,14 +316,16 @@ Odie documents will take days, and should run as low priority. But fetching and
...
 # Performance
-We use NewRelic to monitor our site performance, especially on search requests. If you are doing something around search, make
+We use New Relic to monitor our site performance, especially on search requests. If you are doing something around search, make
 sure you aren't introducing anything to make it much slower. If you can, make it faster.
 
-With NewRelic, you can launch ‘developer mode’ on your development machine.
+With New Relic, you can launch ‘developer mode’ on your development machine.
 
-1. Edit newrelic.yml and change this line to true:
+1. Edit newrelic.yml setting `app_name` to something that includes your name (it will show up in New Relic), setting `monitor_mode` to true, and etting `license_key` to your New Relic license key or the production license key
```
MothOnMars: typo: "and etting `license_key`"

MothOnMars:
LGTM, pending your response to @nickmarden 's comments.

noremmie:
@MothOnMars, the conclusion of the discussion over at [usasearch-cookbooks pull #279](https://github.com/GSA/usasearch-cookbooks/pull/279) is that @nickmarden prefers to let the search-gov version of `config/newrelic.yml` be the source of truth for New Relic config. So I'm going to revert at least some of the changes I've made to that file in this PR and open a new corresponding cookbooks PR. I'm going to need to make other changes too, so I'm going to close this PR out for now.

Also, I still think the New Relic gem bump in this PR is worthwhile to pursue, so I'll be including that change as well.

PR #41, opened 2018-04-27
update rails-html-sanitizer due to XSS vulnerability

MothOnMars:
- https://nvd.nist.gov/vuln/detail/CVE-2018-3741


MothOnMars:
@noremmie , I expect this will pass after https://github.com/GSA/search-gov/pull/41 is merged.

PR #42, opened 2018-04-27
[#154643827] create SearchgovDomain model

MothOnMars:


db/migrate/20180408135739_create_searchgov_domains.rb
```diff
@@ -0,0 +1,15 @@
...
+      t.string :domain, null: false
+      t.boolean :clean_urls, null: false, default: true
+      t.string :status
+      t.integer :urls_count, null: false, default: 0
+      t.integer :unfetched_urls_count, null: false, default: 0
+
+      t.timestamps null: false
+    end
+
+    add_index :searchgov_domains, :domain, length: 100, unique: true
```
noremmie: Is there any reason why this shouldn't use `length: 255` since the`domain` column is that big?

spec/support/fetchable_behavior.rb
```diff
@@ -115,4 +115,16 @@
       it { is_expected.to eq true }
     end
   end
+
+  describe '#indexed?' do
+    subject(:indexed) { record.indexed? }
+
+    context 'when the last_crawl_status = "OK"' do
```
noremmie: Minor nit: it might be nice to have a test for when `#indexed?` returns false.

app/models/searchgov_url.rb
```diff
@@ -15,15 +15,26 @@ class SearchgovUrl < ActiveRecord::Base
...
   validates :url_extension,
     exclusion: { in: BLACKLISTED_EXTENSIONS,  message: "is not one we index" },
     allow_blank: true
 
   before_validation :escape_url
+  before_validation :set_searchgov_domain, on: :create
   before_destroy :delete_document
 
+  belongs_to :searchgov_domain
+  counter_culture :searchgov_domain, column_name: 'urls_count'
```
noremmie: I read up on the `counter_culture` gem since I had never heard of it. I think I get what's going on here. This is what's responsible for keeping `searchgov_domains.urls_count` and `searchgov_domains.unfetched_urls_count` up to date 👍.

db/migrate/20180408135739_create_searchgov_domains.rb
MothOnMars: Yes, MySQL barfs when you try to create a unique index on 255 char. I considered a) making the domains column only 100 char, but *technically* a domain name can be up to 255 char. I figured that the chances of our domains even hitting 100 char are slim enough that I'm not worried about false positives on a unique check for two different 100+ char domains with the same first 100 characters.

spec/support/fetchable_behavior.rb
MothOnMars: can do

MothOnMars:
@noremmie , apologies for those extra `.orig` files. I pushed a minor late-game amendment to this commit that I clearly didn't review properly. Fixing...

db/migrate/20180408135739_create_searchgov_domains.rb
noremmie: Ah... that makes sense. I had checked the index on `searchgov_urls.url` and saw that it was 255 characters, but that index isn't unique. So it wasn't right for me to use that as a comparison 👍.

MothOnMars:
I've removed those `.orig` files and added the spec. There will be a follow-up story to add the super admin page. I was doing those in tandem, but I decided I should split them up to simplify the review & testing. I will also need to create the domain records & update the existing searchgov url records with the association, which I plan to do after this is merged and before the super admin pages are set up.

noremmie:
Sounds good, and LGTM 👍

Issue #30, opened 2018-04-25
Changing a Collection's time scope should not send a searcher back to Page 1 results


nickmarden:
As a searcher,
When I change my Collections time scoping from the default "Any Time" to another time option,
I should not start seeing results from page 1.

Given that I'm on a Collection,
I should continue to see results only from within the Collection I was on
And the results should be filtered and sorted based on my specifications.

Background:
Reported by a customer on Feb 7 2018. The customer noticed that dc= was getting stripped away when she changed the time settings on the Consumer Information Collection of ftc_dev:

https://search.usa.gov/search/docs?affiliate=ftc_prod&dc=5047&query=amazon.com

---

Additional notes from @MothOnMars: 

The time filter links should be using the `/search/docs` route, instead of `/search`:
Example:

link is: https://search.usa.gov/search?affiliate=ftc_prod&query=amazon.com&sort_by=r&tbs=m
link should be: https://search.usa.gov/search/docs?affiliate=ftc_prod&dc=5047&query=amazon.com&sort_by=r&tbs=m

Unfortunately, even when the correct link is used, it appears that the time filter and sort parameters are not used. The same results are returned for the following two searches, whereas the second search should filter out any results older than a month :

https://search.usa.gov/search/docs?affiliate=ftc_prod&dc=5047&query=amazon.com
https://search.usa.gov/search/docs?affiliate=ftc_prod&dc=5047&query=amazon.com&sort_by=r&tbs=m

So there are actually two bugs here:
* one in our construction of the filter links for docs
* one preventing the time filter params from being used in collection searches.

viclim:
@nickmarden
 I will take a look at this during this weekend.

Issue #3, opened 2018-03-07
Separate ELK ES client config from custom index ES client config


nickmarden:
Right now the [ES module](https://github.com/GSA/search-gov/blob/master/lib/es.rb) is designed in such a way that all accesses to Elasticsearch go to the same cluster endpoint(s).

We would like to be able to send requests related to [analytics](https://github.com/GSA/search-gov/blob/79b80c4fac0ecbbff0a0ed2c70642a727fa7ac82/app/models/rtu_top_n.rb) to one ES endpoint and requests related to our custom document indices (i.e. anything using [Indexable](https://github.com/GSA/search-gov/blob/79b80c4fac0ecbbff0a0ed2c70642a727fa7ac82/lib/indexable.rb)) to another ES endpoint. This separation would allow us to upgrade our Elasticsearch usage from ES 1.x to ES 5.x or 6.x independently: first for analytics, then for our custom indices; or vice versa.

PR #43, opened 2018-04-29
[WIP][#30] Fix i14y querying

viclim:
Fix https://github.com/GSA/search-gov/issues/30.

- Use `/search/docs` as the path for `I14ySearch`.
- Includes `tbs` and `sort_by` into `@search_options`. This will be used to send to i14y service.
- Disable re-recording of `application_document_title` cassettes. Why do these need to be rerecorded every 2 months anyway?

@nickmarden

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -105,6 +105,9 @@ RSpec.configure do |config|
       options[:re_record_interval] = nil if /bing.?v5/i === name
       options[:re_record_interval] = nil if /google|gss/i === name
 
+      # disable re-recording application_document_title cassettes which relies on tika server.
+      options[:re_record_interval] = nil if /application_document_title/ === name
+
       VCR.use_cassette(name, options, &example)
     end
   end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -171,6 +171,7 @@ class ApplicationController < ActionController::Base
       @search_params.merge!(tbs: @search.tbs) if @search.tbs
       @search_params.merge!(since_date: @search.since.strftime(I18n.t(:cdr_format))) if permitted_params[:since_date].present? && @search.since
       @search_params.merge!(until_date: @search.until.strftime(I18n.t(:cdr_format))) if permitted_params[:until_date].present? && @search.until
+      @search_params.merge!(dc: permitted_params[:dc]) if permitted_params[:dc].present?
       @search_params.merge!(permitted_params.slice(:contributor, :publisher, :sort_by, :subject))
     end
   end
--- a/app/controllers/searches_controller.rb
+++ b/app/controllers/searches_controller.rb
@@ -115,7 +115,11 @@ class SearchesController < ApplicationController
   end
 
   def set_docs_search_options
-    @search_options = search_options_from_params :dc
+    @search_options = search_options_from_params :dc,
+                                                 :since_date,
+                                                 :sort_by,
+                                                 :tbs,
+                                                 :until_date
     document_collection = @affiliate.document_collections.find_by_id(@search_options[:dc])
     @search_options.merge!(document_collection: document_collection)
   end
--- a/app/helpers/searches/paths_helper.rb
+++ b/app/helpers/searches/paths_helper.rb
@@ -22,11 +22,11 @@ module Searches::PathsHelper
   end
 
   def path_for_i14y_search(search, search_params, extra_params = {})
-    i14y_params = search_params.slice(:affiliate, :m)
+    i14y_params = search_params.slice(:affiliate, :m, :dc)
     i14y_params[:query] = search.query
     i14y_params.merge! extract_current_search_filter_params(search)
     i14y_params.merge! extra_params
-    search_path i14y_params
+    docs_search_path i14y_params
   end
 
   def path_for_blended_search(search, search_params, extra_params = {})
--- a/app/models/search.rb
+++ b/app/models/search.rb
@@ -39,7 +39,6 @@ class Search
   def run
     @error_message = (I18n.translate :too_long) and return false if @query.length > MAX_QUERYTERM_LENGTH
     @error_message = (I18n.translate :empty_query) and return false unless query_present_or_blank_ok? and !query_blacklisted?
-
     response = search
     handle_response(response)
     populate_additional_results
--- a/spec/helpers/searches/paths_helper_spec.rb
+++ b/spec/helpers/searches/paths_helper_spec.rb
@@ -11,11 +11,37 @@ describe Searches::PathsHelper do
                                         sort_by: 'date',
                                         tbs: 'm',
                                         per_page: 20,
+                                        dc: 1,
                                         query: 'test query') }
 
-      it 'returns the correct search path' do 
+      it 'returns the correct search path' do
+        expected_params = { affiliate: affiliate, query: 'test query', sort_by: 'date', tbs: 'm', dc: 1 }
+        expect(helper.path_for_filterable_search(i14y_search, {affiliate: affiliate, dc: 1}, {})).to eq(docs_search_path expected_params)
+      end
+    end
+
+    context 'Blended search' do
+      let(:blended_search) { BlendedSearch.new(affiliate: affiliate,
+                                               sort_by:   'date',
+                                               tbs:       'm',
+                                               query:     'test query') }
+
+      it 'returns the correct search path' do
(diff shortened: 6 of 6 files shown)


app/models/i14y_search.rb
```diff
@@ -2,12 +2,14 @@ class I14ySearch < FilterableSearch
   include SearchInitializer
   include Govboxable
   I14Y_SUCCESS = 200
-  attr_reader :collection
+  attr_reader :collection,
+              :dc
```
nickmarden: @MothOnMars does the i14y search interface support the `dc` attribute? I'm concerned that @viclim might be sending an attribute that will be silently dropped on the floor without the expected behavior.

spec/spec_helper.rb
```diff
@@ -105,6 +105,9 @@
       options[:re_record_interval] = nil if /bing.?v5/i === name
       options[:re_record_interval] = nil if /google|gss/i === name
 
+      # disable re-recording application_document_title cassettes which relies on tika server.
+      options[:re_record_interval] = nil if /application_document_title/ === name
```
nickmarden: @MothOnMars could you weigh on whether you want this change?

nickmarden:
@viclim I'm not sure that this PR fixes the issues that #30 describes. In that issue it is stated that the two issues to be solved are:

> So there are actually two bugs here:
>
> one in our construction of the filter links for docs
> one preventing the time filter params from being used in collection searches.

To make sure I understand this PR better, could you test the following and report the results here (or fix your PR as needed):

1. Whether the "path bug" (`/search` instead of `/search/docs`) also occurs for *non*-i14y searches as well? (the initial bug report seemed to indicated that this path bug was not i14y-specific)
2. Whether the new parameters you are passing to i14y in this PR are actually used by i14y?
3. Whether your PR branch preserved time filter params for both i14y and non-i14y docs searches?

app/models/i14y_search.rb
viclim: @nickmarden I think you're right about `dc` being a useless attribute. [As shown here](https://github.com/GSA/search-gov/blob/master/app/models/i14y_search.rb#L53), all the affiliate's handles are being sent to the i14y service.

So, what's the relationship between a DocumentCollection and an I14yDrawer?

viclim:
@nickmarden 

1. The only time when the filter bar is visible under `/search/docs` is [when `search` is a `FilterableSearch`](https://github.com/GSA/search-gov/blob/master/app/helpers/searches/filters_helper.rb#L5), and `I14ySearch` is the only `FilterableSearch` [used in `/search/docs`](https://github.com/GSA/search-gov/blob/master/app/controllers/searches_controller.rb#L163). So I don't think the path bug occurs for non-i14y searches.

2. It's not being used. Will need to correct that. We need a way to know [which `handle` to use for which document collection though](https://github.com/GSA/i14y/blob/master/app/controllers/api/v1/collections.rb#L69).

3. Only i14y doc search preserved time filter params. Non-i14y docs search has its own link generator, which doesn't include time filter params in the output.

dawnpm:
We definitely need the dc parameter to be retained and used. Otherwise there won’t be a way for us to scope searches within the searchgov index for a given collection.

dawnpm:
Regarding which drawer handle to use, couldn’t we leverage the method that @mothonmars established when we introduced i14y-supported collection searches?

nickmarden:
> We definitely need the dc parameter to be retained and used. Otherwise there won’t be a way for us to scope searches within the searchgov index for a given collection.

@dawnpm what is that comment in response to?

nickmarden:
> Only i14y doc search preserved time filter params. Non-i14y docs search has its own link generator, which doesn't include time filter params in the output.

@viclim I am having a verb problem with "[o]nly i14y doc search preserved time filter params". Did you mean "only i14y doc search _should preserve_ time filter params", or did you really mean (in the past tense) "before this issue was reported, i14y doc search was already preserving time filter params"? 

The first interpretation seems reasonable; the second one seems impossible because we wouldn't be having this discussion if i14y doc search *already preserved* these params.

Sorry for being a grammar nit but I'm trying to make sure I understand your point #3.

app/models/i14y_search.rb
dawnpm: @viclim a `DocumentCollection` is an override for the default search scope set in the domains list. The folders specified in a `DocumentCollection` may or may not be held in an i14y drawer, they may be over in the SearchGov index.

The `DocumentCollection` id can be sent in as a param, to tell the system which override to use....

`search.usa.gov/search?affiliate=foo&query=bar` -> this uses the Domains list as the scope for the query 
`search.usa.gov/search/docs?affiliate=foo&dc=12345&query=bar` -> this uses the list of URL path prefixes listed in the `DocumentCollection 12345`

In either of these cases, the source index may be 
`BingV6` (if `search_engine=BingV6`) 
OR 
`SearchGov + i14y Drawer(s)` (if `search_engine=SearchGov` AND `gets_i14y_results=true`) 
OR 
`SearchGov` (if `search_engine=SearchGov` AND `gets_i14y_results=false`) 
OR 
`i14y Drawer(s)` (if `search_engine=BingV6` AND `gets_i14y_results=true`)

SearchGov is itself an i14y drawer, but related to the affiliate as a search engine, not as an i14y drawer.

viclim:
>>Only i14y doc search preserved time filter params. Non-i14y docs search has its own link generator, which doesn't include time filter params in the output.
>
>@viclim I am having a verb problem with "[o]nly i14y doc search preserved time filter params". Did you mean "only i14y doc search should preserve time filter params", or did you really mean (in the past tense) "before this issue was reported, i14y doc search was already preserving time filter params"?
> 
> The first interpretation seems reasonable; the second one seems impossible because we wouldn't be having this discussion if i14y doc search already preserved these params.
> 
> Sorry for being a grammar nit but I'm trying to make sure I understand your point 3.

It's the first interpretation.

nickmarden:
@MothOnMars after a bunch of discussion I believe we arrived at the following:
- The only `/search/docs` search type that supports filtering (i.e. `is_a? FilterableSearch`) is `I14ySearch`, so @viclim added support to remember `since_date`, `until_date`, `sort_by` and `tbs` to those searches
- The path for filterable i14y searches [is now](https://github.com/GSA/search-gov/pull/43/files#diff-8b22f219477bae569f2d5e85ff41bf1eR29)`/search/docs` instead of `/search`

Could you take a look at the PR? Let me know if you want me to deploy it somewhere for smoke testing.

viclim:
@nickmarden @dawnpm 
> Regarding which drawer handle to use, couldn’t we leverage the method that @MothOnMars established when we introduced i14y-supported collection searches?

You're absolute right about this. I didn't notice [this](https://github.com/GSA/search-gov/blob/master/app/models/i14y_search.rb#L84), which append the `URL prefixes` from `DocumentCollection` to the `query` params, and send it to the i14y API.

dawnpm:
> The path for filterable i14y searches is now /search/docs instead of /search

Does this mean that i14y searches on /search will not be filterable? We want i14y searches to be filterable on _both_ /search and /search/docs.

nickmarden:
@dawnpm I think we've gotten to the heart of this issue and why the path bug exists to begin with.

@viclim I think that [the change you made](https://github.com/GSA/search-gov/pull/43/files#diff-8b22f219477bae569f2d5e85ff41bf1eR29) to force i14y searches to link to `/search/docs` is actually incorrect. We want the link to i14y searches to be `/search/docs` _only if the previous i14y search that we are currently handling was a `/search/docs` search_. 🍝 

In other words asking "what path should we generate for the search?" (e.g. `path_for_i14y_search`) cannot be answered in the helper only by looking at the _type_ of search that is being made, but also by looking in the context of the current SERP in which the results are being shown.

I *think* that's the case anyway. @MothOnMars your thoughts?

@viclim if I'm right, I look forward to an elegant solution to that problem :trollface:

spec/spec_helper.rb
MothOnMars: @nickmarden @viclim , the short answer is yes, but we should do this a different way. 

The backstory on re-recording cassettes is that we originally decided to re-record everything monthly in case there were any external API changes that we weren't aware of. That was later bumped to bi-monthly, and I recently turned off re-recording entirely for cucumber tests. At this point, I'm inclined to turn off re-recording for everything but BingV6-related specs, but I would like to do that in a separate commit (on its way).

MothOnMars:
Closing this up. I have a different fix for this.

PR #44, opened 2018-04-30
[#157199079] only re-record BingV6 cassettes

MothOnMars:


MothOnMars:
@noremmie , I should note that we're making calls to Bing in some of the other specs & features, but I think re-recording for this set of low-level specs should cover us for any unexpected Bing changes.

spec/support/shared_vcr_setup.rb
```diff
@@ -29,6 +29,7 @@
     record: :new_episodes,
     match_requests_on: [:method, :uri_with_paperclip_normalization],
     clean_outdated_http_interactions: true,
+    re_record_interval: nil,
```
nickmarden: Isn't this the default behavior?

MothOnMars: Good point. I'll remove that line.

PR #45, opened 2018-04-30
[#156888108] Bump newrelic_rpm gem and straighten out New Relic configs

noremmie:
This is a new take on #29. This is meant to be deployed only after the changes in [this cookbooks PR](https://github.com/GSA/usasearch-cookbooks/pull/280) are being used for deploys.


config/secrets.yml.dev
```diff
@@ -64,10 +64,9 @@ dev_settings: &DEV_SETTINGS
     adapter: :net_http_persistent
   newrelic:
     <<: *NEWRELIC_SECRETS
-    app_name: USA Search
-    developer_mode: false
+    app_name: USA Search (PUT YOUR NAME HERE)
+    enabled: false
```
MothOnMars: Shouldn't this be `agent_enabled`?
https://docs.newrelic.com/docs/agents/ruby-agent/configuration/ruby-agent-configuration

noremmie: `Rails.application.secrets.newrelic['enabled']` gets used as the `monitor_mode` New Relic configuration value in `config/newrelic.yml` which is what we've been using all along to turn New Relic metrics on or off for an environment. The reason I didn't call it `Rails.application.secrets.newrelic['monitor_mode']` was because when @nickmarden did the original work to extract all the New Relic config into `secrets.yml`, he called the relevant Chef attribute `enabled` instead of `monitor_mode` presumably because the former was more clear. I also think it's more clear this way (i.e., we can tell folks to set `enabled` to true in `secrets.yml` rather than telling them to set `monitor_mode` to true).

MothOnMars: I see. I would expect all of the newrelic config keys to correspond to the standard ones used by New Relic, so this is less clear to me than `monitor_mode` or `agent_enabled` would be. But I can live with it.

MothOnMars:
LGTM 👍

PR #47, opened 2018-05-02
[#155000779] fix time-based searching for collections

MothOnMars:


app/helpers/searches/paths_helper.rb
```diff
@@ -22,11 +22,11 @@ def path_for_filterable_search(search, search_params, extra_params = {})
   end
 
   def path_for_i14y_search(search, search_params, extra_params = {})
-    i14y_params = search_params.slice(:affiliate, :m)
+    i14y_params = search_params.slice(:affiliate, :m, :dc)
     i14y_params[:query] = search.query
     i14y_params.merge! extract_current_search_filter_params(search)
     i14y_params.merge! extra_params
-    search_path i14y_params
+    url_for i14y_params
```
nickmarden: [TIL](https://apidock.com/rails/ActionView/Helpers/UrlHelper/url_for#473-Use-the-current-URL-with-changes), wow.

MothOnMars: Riiiiight? You and me both.

MothOnMars:
@nickmarden , FYI, I'm doing some final local testing before I assign this (to you...) for official review. I like to open unassigned PRs early to get feedback from Code Climate.

spec/helpers/searches/paths_helper_spec.rb
```diff
@@ -1,22 +1,28 @@
 require 'spec_helper'
 
 describe Searches::PathsHelper do
-
   fixtures :affiliates, :i14y_drawers, :i14y_memberships
   let(:affiliate) { affiliates(:power_affiliate) }
 
-  describe '#path_for_filterable_search' do
+  describe '#path_for_filterable_search', type: :request do
```
MothOnMars: @nickmarden , it feels a little dirty to make this an integration test, but I wasn't able to find a way to stub the current path used by the `url_for` method (I barked up a number of trees before finding this solution). If you have such a way, I'm all ears.

app/helpers/searches/paths_helper.rb
MothOnMars: @viclim ,thanks for the work you did on this story. I didn't actually mean to steal it, but in the process of poking the code to determine how it *should* work, I figured out how it *could* work.

MothOnMars: @nickmarden , the return value is just the path, i.e.:
`/search/docs?affiliate=764544198&query=test+query&dc=1&sort_by=date&tbs=m`

PR #48, opened 2018-05-03
[#157196153] add spec for sitemaps with whitespace

MothOnMars:


MothOnMars:
@noremmie , this is just a spec for an issue I fixed in the custom branch we're using for the `sitemaps_parser` gem: https://github.com/MothOnMars/sitemaps/tree/discovery_fixes. I've been adding specs to our repo to cover us in the event that we switch our sitemap parsing gem.

PR #50, opened 2018-05-07
[#157330193] create SearchgovDomainIndexerJob

MothOnMars:


app/models/searchgov_domain.rb
```diff
@@ -11,6 +11,10 @@ def delay
     end
   end
 
+  def index_urls
+    SearchgovDomainIndexerJob.perform_later(self, delay)
```
MothOnMars: We're passing the value of the `delay` to the job, to avoid having to call that method and hit `robots.txt` each time the job is run.

app/models/searchgov_url.rb
```diff
@@ -66,18 +66,6 @@ def document_id
     Digest::SHA256.hexdigest(url_without_protocol)
   end
 
-  def self.fetch_new(delay: 10)
```
MothOnMars: This (old, busted) method is essentially being replaced by the new job.

MothOnMars:
@noremmie , I deliberately kept this job very simple for now, so it doesn't include any logic about checking for current indexing/crawling activity, etc. That will come later, I wanted to get this usable so I can tackle the upcoming NOAA indexing.

MothOnMars:
"I don't think we need to answer these questions right away. I think it would be safe for us to see how this works in production at first and then add some safeguards later to protect against undercrawling and overcrawling."
- That's my thinking too. I left lengthier responses to your q's in https://www.pivotaltracker.com/story/show/157330193/comments/189294220.

README.markdown
```diff
@@ -294,7 +294,7 @@ action on your local machine, assuming you have installed the Redis server.
 
 1. Launch the Sinatra app to see the queues and jobs
 
-    % resque-web
+    % resque-web ./config/initializers/setup_resque.rb
```
MothOnMars: This allows the "Scheduled" & "Delayed" tabs to show up.

lib/tasks/update_https.rake
```diff
@@ -22,44 +22,36 @@ namespace :usasearch do
     end
   end
 
-  def info(msg)
```
MothOnMars: @noremmie & @nickmarden - 

TIL, the hard way: https://kevinjalbert.com/defined_methods-in-rake-tasks-you-re-gonna-have-a-bad-time/

I ran into an issue where, deep in the bowels of some resque code I was working on, `@redis.info` was being called, and `@redis.info` was defined as...this rake task method:
```
> show-method @redis.info
From: /Users/Moth/src/GSA/search-gov/lib/tasks/update_https.rake @ line 25:
Owner: Object
Visibility: private
Number of lines: 5

def info(msg)
  STDOUT.puts(msg) unless Rails.env.test?
end
```

I didn't realize that methods within rake tasks aren't namespaced(!). At any rate, I moved the `puts` silencing to the test via a new `quiet_puts` helper method.  I'm leaving the other methods in this task for now, as their names are less likely to collide with others. That said, I have learned my lesson and will no longer put methods in my rake tasks.

Note: the Redis team is aware of this pitfall, and will avoid it in future versions:
```
Passing 'info' command to redis as is; administrative commands cannot be effectively namespaced and should be called on the redis connection directly; passthrough has been deprecated and will be removed in redis-namespace 2.0 (at /Users/Moth/.rvm/gems/ruby-2.3.5@searchgov-rails42/gems/resque-1.27.4/lib/resque/data_store.rb:59:in `method_missing')
```

nickmarden: lol, ouch.

PR #52, opened 2018-05-10
[#157487414] set correct scheme for sitemap urls

MothOnMars:
- add SearchgovDomain#index_sitemap method
- add SearchgovDomain#scheme method

### DIFF
--- a/app/models/searchgov_domain.rb
+++ b/app/models/searchgov_domain.rb
@@ -15,6 +15,20 @@ class SearchgovDomain < ActiveRecord::Base
     SearchgovDomainIndexerJob.perform_later(self, delay)
   end
 
+  def scheme
+    begin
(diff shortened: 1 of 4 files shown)


app/models/searchgov_domain.rb
```diff
@@ -11,6 +11,20 @@ def delay
     end
   end
 
+  def scheme
+    begin
+      response = open("http://#{domain}/", allow_redirections: :safe, 'User-Agent' => DEFAULT_USER_AGENT)
```
MothOnMars: @noremmie , I am still half-considering storing this in the `SearchgovDomains` table, but for now I want this checked dynamically. I'll be using the `searchgov_domains.status` column as a kind of kill-switch, meaning that any SearchgovDomain with an error status will be deactivated for things like sitemap indexing, crawling, etc.

PR #56, opened 2018-05-15
[#157608938] make redirections check stricter

MothOnMars:


MothOnMars:
@noremmie , I'm still using that gem for domain validation:
https://github.com/GSA/search-gov/blob/master/app/models/searchgov_domain.rb#L35

PR #61, opened 2018-05-16
[#157050583] New passwords should be different than old passwords

ethanewing:


MothOnMars:
@ethanewing , it would be good to add a Cucumber test for this scenario to https://github.com/RapidRiverSoftware/search-gov/blob/master/features/users.feature#L97. If you're not familiar with Cucumber, their [docs](https://docs.cucumber.io/) are great, but you should  be able to add to the existing scenario pretty easily.

The other scenario that we should have called out in this story is resetting a password. See https://github.com/RapidRiverSoftware/search-gov/blob/master/features/password_reset.feature for existing tests. You will find that the requirement of "a user who doesn't remember their current password should not be able to re-use that password" will complicate matters.

app/models/user.rb
```diff
@@ -247,7 +247,7 @@ def confirm_current_password
   end
 
   def new_password_differs_from_current
-    if password == current_password
+    if valid_password?(password)
```
nickmarden: How is `valid_password?` the correct test here?

ethanewing: As far as I can tell, `valid_password?(password)` tests that the variable `password`, when encrypted with the user's salt, matches the encrypted password in the database. It's a method that's connected to the Devise gem. So calling this method gets around the issue of having to compare a plain text password supplied in the form with the encrypted one in the database. Maybe I misunderstood the issue/goal of this fix and that's not behavior we want to support.

features/password_reset.feature
```diff
@@ -46,3 +46,16 @@ Feature: Password Reset
     And I press "Email me a link to reset my password"
     Then I should see "No user was found with that email address"
     And I should be on the password reset page
+
+  Scenario: Trying to reset password to the password currently in the database
```
MothOnMars: As cucumber tests are meant to echo the actual user experience, rather than anything going on under the hood, I would rename this scenario to something less technical, such as `Trying to re-use the same password`.

features/password_reset.feature
```diff
@@ -46,3 +46,16 @@ Feature: Password Reset
...
+    Given I am on the login page
+    And I follow "Forgot your password?"
+    When I fill in "email" with "affiliate_admin@fixtures.org"
+    And I press "Email me a link to reset my password"
+    Then "affiliate_admin@fixtures.org" should receive an email
```
MothOnMars: I believe you can omit these lines:
```
    Then "affiliate_admin@fixtures.org" should receive an email
    When I open the email
    And I should see "https://localhost:3000/password_resets" in the email body
```
Those are tested in the preceding scenarios, and should not be necessary to run this feature.

app/models/user.rb
```diff
@@ -254,4 +263,5 @@ def require_email_verification
     self.requires_manual_approval = !has_government_affiliated_email?
     true
   end
+
```
MothOnMars: looks like an extra line break snuck in

spec/models/user_spec.rb
```diff
@@ -114,10 +114,22 @@
...
+              expect(user.valid_password?('newpass123!')).to be true
+            end
+          end
+
+          context 'when the new password is the same as the current password' do
+            before { user.update_attributes(password: 'goodpass1!', current_password: 'goodpass1!') }
+
+            it 'fails' do
+              expect(user.errors[:password]).to eq ['is invalid: new password must be different from current password']
+              expect(user.valid_password?('goodpass1!')).to be true
```
MothOnMars: You can omit this line, as the preceding line is what you're actually testing.

spec/models/user_spec.rb
```diff
@@ -372,7 +384,7 @@
 
     context "has matching email verification token and requires manual approval" do
       before do
-        expect(adapter).to receive(:push_user).exactly(3).times
+        expect(adapter).to receive(:push_user).exactly(2).times
```
MothOnMars: @ethanewing , let's chat about this tomorrow.

spec/models/user_spec.rb
```diff
@@ -114,10 +114,22 @@
         before { user.require_password_confirmation = true }
 
         context 'when the current password is correct' do
-          before { user.update_attributes(password: 'newpass123!', current_password: 'goodpass1!') }
 
-          it 'updates the password' do
-            expect(user.valid_password?('newpass123!')).to be true
+          context 'when the new password is different from the current password' do
+            before { user.update_attributes(password: 'newpass123!', current_password: 'goodpass1!') }
```
MothOnMars: I forgot to mention that you can call `update` instead of `update_attributes`. They are equivalent, but Rails 4 lets us use `update`. And again, bang methods are preferred in tests, so you can use `update!` here and in the following spec.

app/models/user.rb
MothOnMars: @ethanewing , please remove the extra line break

spec/models/user_spec.rb
MothOnMars: @ethanewing ^

Issue #59, opened 2018-05-16
configure rspec to run specs in random order


MothOnMars:
Currently, our specs run in the same order each time, which can result in unexpected failures when tests are run out of order. This prevents us from running our tests in parallel (https://github.com/GSA/search-gov/issues/55), and generally slows down development when unexpected errors occur.

This behavior can be see by running `rspec spec/ --order rand`.

The default `spec_helper.rb` file includes information on configuring random ordering:
```
  # Run specs in random order to surface order dependencies. If you find an
  # order dependency and want to debug it, you can fix the order by providing
  # the seed, which is printed after each run.
  #     --seed 1234
  config.order = :random

  # Seed global randomization in this process using the `--seed` CLI option.
  # Setting this allows you to use `--seed` to deterministically reproduce
  # test failures related to randomization by passing the same `--seed` value
  # as the one that triggered the failure.
  Kernel.srand config.seed
```

We should:

- [ ] add those configuration options & notes to our `spec_helper.rb` to ensure specs are run in random order
- [ ] fix any specs that are order-dependent

(Note: One reason I've found for some order-dependency is fixtures that are loaded in one test, and then used by subsequent tests that don't explicitly load those fixtures.)

PR #63, opened 2018-05-21
(WIP) Feature/config random order rspec

HalahAb:


spec/models/api_google_docs_search_spec.rb
```diff
@@ -174,6 +178,10 @@
     end
 
     context 'when Google response contains empty results' do
+
+      before { I18n.locale = :es }
```
MothOnMars: @HalahAb , nothing in this test seems to require that the language be set to `es`...? I would expect that the only time these `before` and `after` lines are required would be for specs with a context like `when the site locale is es`.

spec/models/web_results_post_processor_spec.rb
```diff
@@ -68,7 +68,7 @@
                          :published_at => DateTime.parse("2011-09-26 21:33:06"),
                          :guid => '80799 at www.whitehouse.gov',
                          :rss_feed_url_id => rss_feed_urls(:white_house_blog_url).id)
-        15.times do |x|
```
MothOnMars: I think this change is not necessary. See my comments in https://rrsoft.slack.com/archives/G8APZCEH3/p1531181738000177

spec/models/api_google_docs_search_spec.rb
HalahAb: I think it was causing new cassettes files to be generated, because when ran in order the previous example didn't switch back to en and the original cassette was in es for this context. I think I was trying to avoid that, but I've added the modified cassettes now.

HalahAb:
Resolves https://github.com/GSA/search-gov/issues/59

PR #65, opened 2018-05-21
[#157751525] create SitemapIndexerJob

MothOnMars:


MothOnMars:
@noremmie , this story just adds the job; https://www.pivotaltracker.com/story/show/156020227 will handle scheduling. I'd like to see how this works in prod in terms of timing before setting up a schedule.

PR #66, opened 2018-05-23
Mct/create searchgov domain 157816176

MothOnMars:


spec/spec_helper.rb
```diff
@@ -70,7 +70,6 @@
     TestServices::delete_es_indexes
     TestServices::create_es_indexes
     require "#{Rails.root}/db/seeds/template.rb"
-    ActiveJob::Base.queue_adapter = :test
```
MothOnMars: I moved this setting to `config/environments/test.rb` to be consistent with how we set the queue adapter in other environments.

MothOnMars:
Yeah, I'm trying not to overthink some of this code. I considered throwing a `rescue` in there, but I'm still not sure I want `check_status` to raise an error at all.

noremmie:
648fdf2 LGTM too 👍

PR #68, opened 2018-05-29
Mct/super admin 157223800

MothOnMars:


MothOnMars:
Per my discussion with @noremmie ,  the URL searches will always be scoped by the `searchgov_urls.searchgov_domain_id`, such as: `where searchgov_urls.searchgov_domain_id = 1 AND searchgov_urls.url like %foo%'`. The fuzzy matching on `url` is not ideal, but those searches will not require a table scan due to the foreign key scoping. If we see any performance issues due to the fuzzy matching, we should consider a) backing that table with elasticsearch or b) investigating [MySQL's full text search](https://dev.mysql.com/doc/refman/5.5/en/fulltext-search.html).

(For future reference, Active Scaffold only lets you specify a "starts with" query if you are searching by a single field. For multi-field searches such as the one on `last_crawl_status` and/or `url`, it does not allow you to customize the drop-down options or ordering).

PR #69, opened 2018-05-29
[#157825658] DRY up SearchEngineAdapter#run and delegate ImageSearch diagnostics to its search_instance which does the actual search

noremmie:


noremmie:
@MothOnMars, the motivation for this change is that `SearchEngineAdapter.run` is exactly like `SearchOnCommercialEngine.search` except that the latter picks the search engine response diagnostics off of the response and attaches it to the Search model so that they can be logged later by `SearchImpression.log`. The fact that `SearchEngineAdapter.run` does its own thing means that any search engine response diagnostics don't get logged for its searches. This is causing a certain small percentage of our image searches to not have response diagnostics associated with them which is throwing off our own count of how many requests we're sending to Bing.

This update just changes `SearchEngineAdapter` to include the `SearchOnCommercialEngine` module and implements the `SearchEngineAdapter#run` method in terms of `SearchOnCommercialEngine.search`. It also adds error handling for the case when `SearchEngineAdapter.run` returns false.

The only other change is that `ImageSearch` has its own `diagnostics` method, but it has never been used to actually set any diagnostics from a search engine response. `ImageSearch` should instead be delegating that method to its underlying search instance just like it does for other things like the search results.

app/engines/search_engine_adapter.rb
```diff
@@ -24,18 +26,11 @@ def initialize(klass, options)
...
-    end
-  rescue SearchEngine::SearchError => error
-    Rails.logger.warn "Error getting image search results from #{@search_engine.class} endpoint: #{error}"
-    false
+    @search_engine_response = search
   end
 
   def results
-    @results || (paginate(post_process_results(@search_engine_response.results)) if @search_engine_response.results)
+    @results || (paginate(post_process_results(@search_engine_response.results)) if @search_engine_response && @search_engine_response.results)
```
MothOnMars: I think you can clean this up with the "safe operator":
`if @search_engine_response&.results`

spec/models/image_search_spec.rb
```diff
@@ -17,6 +17,46 @@
     end
   end
 
+  describe '#diagnostics' do
+    subject(:image_search) { ImageSearch.new(affiliate: affiliate, query: 'corgis', cr: use_commercial_results) }
+    let(:use_commercial_results) { nil }
+    let(:affiliate) { affiliates(:basic_affiliate) }
+    let(:search_engine) { nil }
+    before { allow(affiliate).to receive(:search_engine).and_return(search_engine) }
+    before { allow_any_instance_of(underlying_search_class).to receive(:diagnostics).and_return(:underlying_diagnostics) }
```
MothOnMars: Nit: this line is pretty long and could be broken up for readability

app/engines/search_engine_adapter.rb
noremmie: Fixed!

spec/models/image_search_spec.rb
noremmie: This is also fixed, and I also broke up the `subject` business above per our chat 👍

noremmie:
Thanks, @MothOnMars. I have picked your nits 😀

app/engines/search_engine_adapter.rb
MothOnMars: I'm curious - what necessitated the reversion?

spec/models/image_search_spec.rb
```diff
@@ -18,12 +18,17 @@
   end
 
   describe '#diagnostics' do
-    subject(:image_search) { ImageSearch.new(affiliate: affiliate, query: 'corgis', cr: use_commercial_results) }
+    subject(:image_search) do
```
MothOnMars: 👍

app/engines/search_engine_adapter.rb
noremmie: In the case of an error, `SearchEngineAdapter.run` returns false making `@search_engine_response` false, and false doesn't have a `.results` method.

MothOnMars: Ah, gotcha. Thanks, that makes sense.

MothOnMars:
👍 LGTM, pending CircleCI

PR #71, opened 2018-05-31
(WIP) Feature/Circleci setup workflow and parallelise cucumber tests

HalahAb:


lib/tasks/simplecov_check_coverage.rb
```diff
@@ -0,0 +1,15 @@
+require 'simplecov'
+result_hash = SimpleCov.result.to_hash
+puts "#{result_hash.keys}"
+puts "NOW CHECKING COVERAGE"
+puts "Coverage = #{SimpleCov.result.covered_percent}"
```
MothOnMars: @HalahAb , this is looking good so far. One CircleCI feature you may find useful, if you haven't already been using it, is the "Rerun job with SSH" button. You can throw a `binding.pry` line into your code, then run the job with SSH so that you can tinker with the code directly on the CircleCI server, look around the environment, etc.

HalahAb: @MothOnMars Thanks Martha! oh yeah that would be very useful, I saw that button but I'm not sure I can do anything with it, it's telling me I require write permissions to trigger builds.

MothOnMars: Ah, right. Phooey.

.circleci/config.yml
```diff
@@ -4,23 +4,14 @@
...
-      - image: circleci/mysql:5.6
-        environment:
-          MYSQL_ALLOW_EMPTY_PASSWORD: true
-          MYSQL_ROOT_HOST: "%"
-
-      - image: redis:3.2
-
-    working_directory: ~/search-gov
-
     steps:
```
MothOnMars: I noticed that sometimes `steps:` is followed by one line break, and sometimes two. For consistency, let's just use one line break throughout the file.

.circleci/config.yml
```diff
@@ -55,34 +92,49 @@ jobs:
...
+          key: elasticsearch
 
       - run:
           name: Prepare database
           command: |
             bundle exec rake db:create
             # Configure MySQL 5.6 for ROW_FORMAT=DYNAMIC storage
             cat .circleci/database-configuration.sql | bundle exec rails db
-            bundle exec rake db:structure:load
+            bundle exec rake db:structure:load
```
MothOnMars: looks like some extra spaces snuck in here

.circleci/config.yml
```diff
@@ -103,17 +155,145 @@ jobs:
...
+            sudo apt-get update && sudo apt-get install mysql-client 
+
+      - restore_cache:
+          key: search-gov-{{ checksum "Gemfile.lock" }}
+
+      - run: bundle --path vendor/bundle
+
+      - restore_cache:
+          key: elasticsearch
+
```
MothOnMars: remove extra line break

.codeclimate.yml
```diff
@@ -38,3 +38,4 @@ exclude_paths:
 - app/api/search_consumer/
 -  "**/vendor/**/*"
 - app/assets/images/
+- scripts/
```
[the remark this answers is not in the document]
MothOnMars: Note: I realized that any failures in that directory are probably related to my original reporting code. I'll run take a look at the failures....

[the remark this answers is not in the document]
HalahAb: Hey Martha, I added the scripts directory and a skip over it here because the coverage script was being considered as part of the total coverage. I tried adding it in another path under lib>tasks>check_coverage and tried to specifically ignore that path only but that seemingly didn't work, it seemed to be related to a config issue I couldn't figure out. The other alternative would be to save it under the same path but write a test for it to account for it's coverage. Would that be a better option?

MothOnMars:
@HalahAb , this is working well. I'm happy to see the CircleCI failure when coverage is below 100%, and I'm *really* happy to see the coverage output. Once those (minor) stylistic issues are cleaned up, this should be good to go.

.codeclimate.yml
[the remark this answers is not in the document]
MothOnMars: Ah, I see. In that case, you could skip the check for that specific file, rather than the whole directory.

HalahAb:
Hey @MothOnMars I updated the pull request with the fixes you suggested. I’ve removed the exclude on the scripts path I had in place earlier, and it seems that it was not needed to begin with. I tried to verify this by testing a few different things.

The issue that led me to have it in there was likely to do with the fact that I had the coverage_check file under lib > tasks> check_coverage, and the lib directory couldn't be excluded from coverage. so it ending up checking that file for coverage.

Then I guess I assumed that I had to explicitly exclude the file path from being checked for coverage when I moved it to a 'scripts' folder. Right now, it seems to be working fine without needing to be explicitly exclude the path or the file. I just wanted to double check that was ok.

.circleci/config.yml
MothOnMars: @HalahAb , I think I was unclear. `steps:` should be followed by just one line break throughout the file, i.e.:
```
  steps:
    - restore_cache
```

HalahAb: ah so sorry I misunderstood! I just pushed again!

PR #73, opened 2018-06-01
results api instrux update

dawnpm:
We've had questions about the endpoint for accessing Garfield results and what the highlighting character is, so I'm updating the instructions to provide these answers proactively.

While I'm in here, I'm also removing mention of Google Site Search.

As I create this, I see that my CircleCI tests on the first two commits, but I don't know why that might be.


noremmie:
@dawnpm I fixed up the cucumber test that these changes broke (the test was expecting a Google reference that no longer exists), and I've also updated some of the language around highlighting. We use one special character at the beginning of the quoted text and a different special character at the end of the quoted text. They both look the same in a browser. How do you feel about the differences  between these two screenshots?

url_encoded_highlighting_character

raw_highlighting_characters

There's one other thing about the highlighting language. [That wikia reference](http://unicodesymbols.wikia.com/wiki/%EE%80%80) doesn't mention anything about highlighting and actually says that's an unknown characters. I'm confused that might lead to more questions. I can't find any other useful links. How would you feel about me removing the "and you can learn more..." clause?

dawnpm:
@noremmie I added the direct character references.

PR #75, opened 2018-06-05
lock down searchgov urls route

MothOnMars:


MothOnMars:
@noremmie , this change ensures that users can't access `admin/searchgov_urls` directly, which could open us up to the abusive DB queries we discussed in https://github.com/GSA/search-gov/pull/68.

dawnpm:
I disagree with implementing this lockdown until we see what the query patterns are like. It's incredibly useful for the program team to be able to view indexing issues across domains.

MothOnMars:
@dawnpm , we can already see what the query patterns are like. For example, the following search would require a scan of the entire `searchgov_urls` table, because it requires a full-text search on both `searchgov_urls.last_crawl_status` and `searchgov_urls.url`:
```
Started GET "/admin/searchgov_urls?utf8=%E2%9C%93&search%5Burl%5D%5Bopt%5D=%25%3F%25&search%5Burl%5D%5Bfrom%5D=state.gov%2Fj%2Finl%2F&search%5Burl%5D%5Bto%5D=&search%5Blast_crawl_status%5D%5Bopt%5D=%25%3F%25&search%5Blast_crawl_status%5D%5Bfrom%5D=missing&search%5Blast_crawl_status%5D%5Bto%5D=&commit=Search" for 66.44.31.180 at 2018-06-04 21:06:14 +0000
Processing by Admin::SearchgovUrlsController#index as JS
  Parameters: {"utf8"=>"✓", "search"=>{"url"=>{"opt"=>"%?%", "from"=>"state.gov/j/inl/", "to"=>""}, "last_crawl_status"=>{"opt"=>"%?%", "from"=>"missing", "to"=>""}}, "commit"=>"Search"}
```

The default search on that page is a "contains" search, which cannot use the DB index  and therefore requires a scan of the full multi-million line table. That search is not ideal but much safer when scoped by the domain ID, which is the behavior on the search domains page. A "starts with" or "=" search is safer, but there's no guarantee that each user will only use those options, and there's no easy way to remove "contains" as an option from the drop-down. I could omit the drop-down and limit the search type to `value =`, but that would severely limit the search functionality based on the domain.

I can see the utility in having a cross-domain search, but we would need a separate story for that. Unfortunately, ActiveScaffold doesn't provide an easy way to allow certain search types in a nested view (by domain) but restrict them in an un-nested view (such as by allowing only "value starts with" or "value = " type searches). If we need a full-text search across domains on either the URL field or the `last_crawl_status` field, we would need to investigate backing that with Elasticsearch or by using a different DB index. If any reports are needed before that's implemented, we can pull those from the database.

dawnpm:
Ok @MothOnMars , your points are convincing. Thanks for the discussion.

PR #76, opened 2018-06-06
[#158132420] increase SearchgovUrl doc size limit to 15mb

MothOnMars:


noremmie:
LGTM! And I think it would be fine to accept this from prod given that it's not customer-facing 👍

PR #79, opened 2018-06-07
[#156343590] improve expanding URL regex

MothOnMars:
@noremmie , I ran this regex against all the existing searchgov URL records, and found zero false positives.


MothOnMars:
Speaking of grokking regexes, this is my new favorite site:
https://regex101.com/r/sJ9gM7/46

jamesisaacs:
@MothOnMars +1 I found that site too not long ago.

PR #80, opened 2018-06-07
[#158173466] Add sort_by=date support to /api/v2/search/i14y

noremmie:
@MothOnMars it turns out that when the `/api/v2/search/i14y` endpoint was originally created, it was wired up to the wrong search options validator class. The `Api::NonCommercialSearchOptions` is just like the `Api::SearchOptions` class except that it adds the `sort_by` option which `I14yApiSearch` already knows how to properly consume.

### DIFF
--- a/spec/controllers/api/v2/searches_controller_spec.rb
+++ b/spec/controllers/api/v2/searches_controller_spec.rb
@@ -13,7 +13,8 @@ describe Api::V2::SearchesController do
       query_or: 'alternative',
       query_quote: 'barack obama',
       filetype: 'pdf',
-      filter: '2'
+      filter: '2',
+      sort_by: 'date',
     }
   end
   let(:query_params) do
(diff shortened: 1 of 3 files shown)


spec/controllers/api/v2/searches_controller_spec.rb
```diff
@@ -365,6 +365,17 @@
       it { is_expected.to respond_with :success }
     end
 
+    context 'when the search options include sort_by' do
```
MothOnMars: This strikes me as an unnecessarily specific test. I'm thinking it could be something more like:
 ```
when the search options are valid
    it passes the correct options to its ApiI14ySearch object
```

 and test all the expected options.

MothOnMars:
https://github.com/GSA/search-gov/pull/80/commits/845f9f72b0b8aedb7fd199b7fd314e61c4eafa41 LTGM 👍

PR #83, opened 2018-06-08
[#158225026] refactor SitemapIndexer to create & update but not fetch

MothOnMars:
urls
- raise mass assignment error in dev & test environments

### DIFF
--- a/app/jobs/searchgov_domain_indexer_job.rb
+++ b/app/jobs/searchgov_domain_indexer_job.rb
@@ -2,9 +2,9 @@ class SearchgovDomainIndexerJob < ActiveJob::Base
   queue_as :searchgov
 
   def perform(searchgov_domain:, delay:)
-    searchgov_domain.searchgov_urls.unfetched.first&.fetch
+    searchgov_domain.searchgov_urls.fetch_required.first&.fetch
 
-    if searchgov_domain.searchgov_urls.unfetched.any?
+    if searchgov_domain.searchgov_urls.fetch_required.any?
       SearchgovDomainIndexerJob.set(wait: delay.seconds).
         perform_later(searchgov_domain: searchgov_domain, delay: delay)
     end
--- a/app/jobs/sitemap_indexer_job.rb
+++ b/app/jobs/sitemap_indexer_job.rb
@@ -2,6 +2,6 @@ class SitemapIndexerJob < ActiveJob::Base
   queue_as :searchgov
 
   def perform(searchgov_domain:)
-    searchgov_domain.index_sitemap
+    SitemapIndexer.new(searchgov_domain: searchgov_domain).index
   end
 end
--- a/app/models/searchgov_url.rb
+++ b/app/models/searchgov_url.rb
@@ -13,7 +13,7 @@ class SearchgovUrl < ActiveRecord::Base
                                 application/vnd.openxmlformats-officedocument.spreadsheetml.sheet
                               )
 
-  attr_accessible :last_crawl_status, :last_crawled_at, :url
+  attr_accessible :last_crawl_status, :last_crawled_at, :url, :lastmod
   attr_reader :response, :document
   attr_readonly :url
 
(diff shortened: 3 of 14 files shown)


MothOnMars:
@noremmie , this refactoring is intended to:
1) resolve the timeout errors that have been occurring in the sitemap indexing jobs
2) separate concerns between gathering the new & updated sitemap URLs vs. indexing those URLs.

The `SitemapIndexer` now only creates & updates `SearchgovUrl` records, but leaves the indexing to the `SearchgovDomainIndexerJob`.

db/migrate/20180608190543_add_lastmod_to_searchgov_url.rb
```diff
@@ -0,0 +1,5 @@
+class AddLastmodToSearchgovUrl < ActiveRecord::Migration
+  def change
+    add_column :searchgov_urls, :lastmod, :datetime
```
MothOnMars: Achtung!

This column is being added to the `searchgov_urls` table, so this migration could take a while.

app/jobs/searchgov_domain_indexer_job.rb
```diff
@@ -2,9 +2,9 @@ class SearchgovDomainIndexerJob < ActiveJob::Base
   queue_as :searchgov
 
   def perform(searchgov_domain:, delay:)
-    searchgov_domain.searchgov_urls.unfetched.first&.fetch
+    searchgov_domain.searchgov_urls.fetch_required.first&.fetch
```
MothOnMars: The new `SearchgovUrl.fetch_required` scope includes both new URLs and URLs with a `lastmod` value later than `last_crawled_at`.

app/jobs/sitemap_indexer_job.rb
```diff
@@ -2,6 +2,6 @@ class SitemapIndexerJob < ActiveJob::Base
   queue_as :searchgov
 
   def perform(searchgov_domain:)
-    searchgov_domain.index_sitemap
+    SitemapIndexer.new(searchgov_domain: searchgov_domain).index
```
MothOnMars: @noremmie , you and I talked about this change a few days ago; the SearchgovDomain#index_sitemap method now enqueues a SitemapIndexerJob.

app/models/searchgov_url.rb
```diff
@@ -33,7 +33,9 @@ class SearchgovUrl < ActiveRecord::Base
   counter_culture :searchgov_domain, column_name: 'urls_count'
   counter_culture :searchgov_domain,
     column_name: proc {|url| !url.fetched? ? 'unfetched_urls_count' : nil },
-    column_names: { ['searchgov_urls.last_crawl_status is null'] => 'unfetched_urls_count' }
+    column_names: { ['searchgov_urls.last_crawled_at IS NULL'] => 'unfetched_urls_count' }
```
MothOnMars: This tweak is to ensure that `unfetched` here means the same thing as it does in `Fetchable`'s `unfetched` scope: 
```ruby
scope :unfetched, -> { where('ISNULL(last_crawled_at)') }
```

config/environments/development.rb
```diff
@@ -37,6 +37,9 @@
     # Expands the lines which load the assets
     config.assets.debug = true
   end
+
+  # Raise exception on mass assignment protection for Active Record models
```
MothOnMars: http://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#upgrading-from-rails-3-1-to-rails-3-2-config-environments-test-rb

db/migrate/20180608190543_add_lastmod_to_searchgov_url.rb
noremmie: Thanks for the heads up.

How would you feel about naming the new column `last_modified_at` so that it is more consistent with the other multi-word columns in this table?

* `last_crawled_at`
* `last_crawl_status`
* `load_time`
* `created_at`
* `updated_at`
* `searchgov_domain_id`

MothOnMars: @noremmie , I've thought about this more and I would like to keep this consistent with the `lastmod` sitemap attribute. I've tested this on demo1, and it's working as desired. Let me know if you'd like further testing on this; otherwise, I'll go ahead and merge it.

MothOnMars: (P.S. My first reply didn't get posted immediately. Despite the timestamps, several hours did pass between those two comments!)

noremmie: Oh - I see. That makes sense.

PR #85, opened 2018-06-11
[#157205231] pass URL "changed" value to I14y

MothOnMars:
- add support for Dublin Core data
- return Time objects instead of strings for WebDocument timestamps


MothOnMars:
@noremmie , this PR populates the `changed` field for searchgov docs with the either the `lastmod` value from a sitemap, or the modification date from the HTML metadata (either Open Graph or Dublin Core) -- whichever is more recent. It also tweaks the `WebDocument` API to return a `Time` value rather than a string for the document timestamps, because those can come in a hodgepodge of formats.

app/models/html_document.rb
```diff
@@ -73,4 +77,12 @@ def main_html
       element&.text.present?
     end&.to_html
   end
+
+  def dublin_core_date
```
MothOnMars: Note: I'm only using the DC data in `HtmlDocument` but not `ApplicationDocument` because I've found that application documents almost always have both a created and a modified timestamp, and any other metadata such as Dublin Core or Open Graph is redundant.

app/models/web_document.rb
```diff
@@ -18,6 +18,14 @@ def language
     @language ||= extract_language&.downcase
   end
 
+  def created
+    timestamp(extract_created)
+  end
+
+  def changed
```
MothOnMars: I'm using "changed" here to mirror the I14y API, which takes `created` and `changed`.

spec/models/html_document_spec.rb
```diff
@@ -85,18 +86,73 @@
         expect(description).to eq 'My OG Description'
       end
     end
+
+    context 'when a Dublin Core description is available' do
+      let(:raw_document) { doc_with_dc_data }
+
+      it 'returns the open graph description' do
```
noremmie: I think this should say "returns the Dublin Core description" instead of "returns the open graph description".

noremmie:
54741f5 LGTM too 👍

PR #86, opened 2018-06-19
Mct/speedier sitemaps

MothOnMars:


MothOnMars:
@noremmie , this is the latest round of refactoring sitemap processing (another will come after Ethan finishes the Sitemaps model). In order to speed up the sitemap indexing jobs, the `SitemapIndexer` now processes a single sitemap at a time, rather than all the sitemaps for a domain (that may seem like a no-brainer, but I had been using the handy `discover` method of the `Sitemaps` gem to find and process all the sitemaps in tandem -- no more.)

This change still requires bumping the Resque timeout (https://github.com/GSA/search-gov/pull/89), as sitemaps with 50k URLs still take quite a while to process. I will also need to keep a DNM label on this until I've finished the [manage searchgov domain states](https://www.pivotaltracker.com/story/show/157986614) story is finished, because I need to prevent each sitemap indexing process from kicking off a new domain indexing job.

MothOnMars:
@noremmie https://github.com/GSA/search-gov/pull/86/commits/a748b095507bbbf47827315e6e72247c4baa39f2 is a follow-up to the addition of the sitemaps table: https://github.com/GSA/search-gov/pull/92
This will be mergeable once https://github.com/GSA/search-gov/pull/108 is merged, because the `SitemapIndexer` can call `SearchgovDomain#index_urls` any number of times without enqueuing duplicate domain indexer jobs.

PR #87, opened 2018-06-20
(WIP) fix flakey cucumber test:  "Add/edit/remove Collection"

HalahAb:


Gemfile.lock
```diff
@@ -859,6 +864,7 @@ DEPENDENCIES
   virtus (~> 1.0.5)
   web-console (~> 2.0)
   webmock (~> 3.1.1)
+  webrick (~> 1.3, >= 1.3.1)
```
MothOnMars: @HalahAb , what does adding `webrick` accomplish? (I'm not opposed to it; I just want to make sure we have a good reason to add new gems)

features/step_definitions/debugging_steps.rb
```diff
@@ -10,3 +10,7 @@
 When /^I write the page body to (.*)$/ do |path|
   File.open(path, 'w') { |f| f.puts(page.body) }
 end
+
+And /^(?:|I )confirm alert message is present "([^"]*)"$/ do |msg|
```
MothOnMars: I don't think we need this extra debugging step. The following step definitions should accomplish the same thing:

```ruby
features/step_definitions/web_steps.rb

When /^(?:|I )press "([^"]*)" and confirm "([^"]*)"$/ do |button, msg|
  accept_confirm msg do
    click_button(button)
  end
end

When /^(?:|I )follow "([^"]*)" and confirm "([^"]*)"$/ do |link, msg|
  accept_confirm msg do
    click_link(link)
  end
end
```

features/support/env.rb
```diff
@@ -10,12 +10,14 @@
 require 'cucumber/rails'
 require 'capybara/rails'
 require 'email_spec/cucumber'
+require 'capybara-screenshot/cucumber'
```
MothOnMars: With the addition of cabybara-screenshot, we should update the [`show me the page` debugging step definition](https://github.com/GSA/search-gov/blob/master/features/step_definitions/debugging_steps.rb#L1) to call `screenshot_and_open_image` instead of `save_and_open_page`. (Also: make sure that works ;) )

features/step_definitions/web_steps.rb
```diff
@@ -53,6 +53,11 @@ def with_scope(locator)
   click_button(button)
 end
 
+#Alternative method to click_button(button), used when button can not be found
+When /^(?:|I )find the "([^"]*)" button and click it$/ do |button|
```
MothOnMars: Actually, let's reword this slightly for readability:
`When /^(?:|I )find and click the "([^"]*)" button$/`

Gemfile.lock
HalahAb: I was prompted to install it after upgrading to capybara 3.3.1 I think its running the capybara.server, but I've since reverted back to version 2.18. I couldn't reconcile the differences that lead to the test breaks, and I didn't want to take up too much time trying. 2.18 seems to be working fine for now.

MothOnMars:
Test results:
✅ cuke tests passed 20x locally
✅ all tests passed 5x on CircleCI
✅ `show me the page` opens a screenshot with CSS

Merging! 🎆

PR #88, opened 2018-06-20
[#155989343] Use original i14y query instead of spelling suggestion for govbox searches

noremmie:
The fact that we do a govbox search using i14y's spelling suggestion for the user's original search term instead of doing a govbox search using the user's original search term is causing trouble. The fix is to just use the user's original search term for the govbox search (and update the appropriate expectations in the specs, of course).


MothOnMars:
LGTM 👍

PR #92, opened 2018-06-21
Feature/create sitemaps model

ethanewing:
Preliminary pull request.


app/concerns/fetchable.rb
```diff
@@ -47,6 +47,7 @@ module Fetchable
     scope :unfetched, -> { where('ISNULL(last_crawled_at)') }
 
     before_validation :normalize_url
+    before_validation :truncate_error_message
```
MothOnMars: The `last_crawl_status` field will not always be an error message. (When things go right, its value is `'OK'`.) So I would rename this method to `truncate_last_crawl_status`.

app/concerns/fetchable.rb
```diff
@@ -90,4 +91,14 @@ def downcase_scheme_and_host_and_remove_anchor_tags
   def url_extension
     Addressable::URI.parse(url).extname.downcase.from(1)
   end
+
+  def set_searchgov_domain
+    self.searchgov_domain = SearchgovDomain.find_or_create_by(domain: URI(url).host) unless url.nil?
+  end
+
+  def truncate_error_message
+    if self.last_crawl_status && self.last_crawl_status.length > 255
```
MothOnMars: This method can be reduced to one line. I'll give you a hint: Rail's [ActiveSupport library](http://guides.rubyonrails.org/active_support_core_extensions.html) adds some handy additional methods to core Ruby classes like `String`. Also, Ruby 2.3 introduced the '[safe navigation operator](http://mitrev.net/ruby/2015/11/13/the-operator-in-ruby/)' to help when calling methods on a object that might be `nil`, i.e.:
```ruby
> 'foo'.reverse
"oof"
> nil.reverse
NoMethodError: undefined method `reverse' for nil:NilClass
> nil&.reverse
nil

File: db/migrate/20180611171416_create_real_sitemaps.rb [hunk 3]
```diff
@@ -0,0 +1,12 @@
+class CreateRealSitemaps < ActiveRecord::Migration
+  def change
+    create_table :real_sitemaps do |t|
```
nickmarden: Did we intend to change the name of the table to `real_sitemaps`? That seems like a weird choice. It's in the database, so of course it's real...

app/models/real_sitemap.rb
```diff
@@ -0,0 +1,13 @@
+class RealSitemap < ActiveRecord::Base
```
MothOnMars: We still want to use `Sitemap` as the name of the class. You need to rename the migration file & migration class (`CreateSitemapsTable`?), not the ActiveRecord class.

db/structure.sql
```diff
@@ -1,8 +1,8 @@
--- MySQL dump 10.13  Distrib 5.6.35, for osx10.11 (x86_64)
+-- MySQL dump 10.13  Distrib 5.7.22, for osx10.13 (x86_64)
 --
--- Host: localhost    Database: usasearch_development
+-- Host: 127.0.0.1    Database: usasearch_test
```
nickmarden: I have no idea what's going on here, but this `structure.sql` contains some bad info...

- We don't dump the `usasearch_test` DB
- We aren't using MySQL 5.7.x

app/models/real_sitemap.rb
```diff
@@ -0,0 +1,13 @@
...
+  include Fetchable
+
+  attr_readonly :url
+
+  belongs_to :searchgov_domain
+
+  validates :url, presence: true, uniqueness: true,
```
MothOnMars: As I mentioned in our chat:

"You probably don’t need these validations in the Sitemap class, because you’re getting the validations in `Fetchable` automatically:
  ```
validates :url, presence: true, uniqueness: true,
   length: {maximum: 2000}, format: {with: /\Ahttps?:\/\/\w+\.gov\/\w+\.(xml|gz|txt)/, message: 'invalid url'}

```
Fetchable validations:
 ```
validates_length_of :url, maximum: 2000
validates_presence_of :url
validates_url :url, allow_blank: true
```
I should also note that the `validates_url` method comes courtesy of the `validate_url` gem, which takes care of the URL format validation (and don’t be surprised if some unexpected URLs are marked as “valid” - URLs are notoriously tricky to validate: https://mathiasbynens.be/demo/url-regex)"

db/migrate/20111103131805_create_sitemaps.rb
```diff
@@ -4,7 +4,7 @@ def self.up
       t.string :url
       t.references :affiliate
       t.timestamp :last_crawled_at
-
```
MothOnMars: As this file isn't related to your changes, I would leave it as the version on master (even if that means leaving some extra whitespace):

`git co master db/migrate/20111103131805_create_sitemaps.rb` 

As a general rule, migration files are never altered once they are used; they essentially serve as historical documentation of DB changes over time. (Some organizations do periodically delete old migration files, but we do not.)

db/migrate/20180611171416_create_real_sitemaps.rb
MothOnMars: Nope. I commented on this above.

@ethanewing , another good rule of thumb is to consider how readable & understandable your code will be to future developers. If you came across a class named `RealFoo`, you would probable wonder, 'what makes this different from `Foo`? Is there a `FakeFoo` class?`. Confusion would ensue.

app/models/real_sitemap.rb
nickmarden: @MothOnMars I do think that `validates_url` isn't going to suffice completely - we still need to validate that the URL ends in `\.(xml|gz|txt)`, right?

MothOnMars: Nope - I specifically told Ethan not to validate the extension, because I've seen XML sitemaps with other extensions, such as `.aspx`. I'm trusting our super admins (and eventually our admins) not to put junk into that table. If we see too much Garbage In, I may want to lock down the validations at that point.

nickmarden: Oh, OK. I was context-free on that part of the conversation. I think you can see how I inferred the requirement from this validation though.

MothOnMars: Yeah, that's the validation I told him to remove.

db/structure.sql
MothOnMars: @ethanewing , let's chat about this in our meeting.

spec/models/sitemap_spec.rb
```diff
@@ -0,0 +1,72 @@
+require 'spec_helper'
+
+describe Sitemap do
+  let!(:url) { 'http://www.agency.gov/boring' }
```
MothOnMars: P.S. I'm guessing you snagged `http://www.agency.gov/boring` from one of my specs. I was apparently feeling snarky that day; let's use something more, ah, boring, such as: `http://www.agency.gov/doc1`.

spec/models/sitemap_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+  let!(:html) { read_fixture_file("/html/page_with_og_metadata.html") }
```
MothOnMars: Copy pasta?

app/models/real_sitemap.rb
ethanewing: So I was finding when running certain specs in `sitemap_spec.rb` that a test that made sure the URL 'https://foo/bar' was not valid would not pass, because a Sitemap object would still end up being created despite that (I think) we don't want that URL to be considered valid.  (I suppose I'm misunderstanding what we want out of our sitemap model, though.)

And now that I'm looking through that spreadsheet it looks like my validation doesn't come close to accommodating all those, so it was a fool's errand on my part to try to write a validation in the first place.

db/structure.sql
MothOnMars: @ethanewing , please install [MySQL 5.6](https://dev.mysql.com/doc/refman/5.6/en/installing.html)

MothOnMars: @ethanewing - don't worry about switching to 5.6 for now. That can happen after we wrap up this story.

spec/models/sitemap_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+  let!(:valid_attributes) { { url: url } }
+  let!(:sitemap) { Sitemap.new(valid_attributes) }
+  let!(:existing_domain) { SearchgovDomain.create!(domain: 'existing.gov') }
+
+  it { is_expected.to have_readonly_attribute(:url) }
+
+  describe 'schema' do
```
MothOnMars: Now that I think about it, we can move these schema specs to `Fetchable`, because these should also apply to all the classes that include `Fetchable`. The duplicate specs can be removed (if they are present) from `searchgov_url_spec.rb` and `indexed_document_spec.rb`.

app/concerns/fetchable.rb
```diff
@@ -90,4 +91,14 @@ def downcase_scheme_and_host_and_remove_anchor_tags
   def url_extension
     Addressable::URI.parse(url).extname.downcase.from(1)
   end
+
+  def set_searchgov_domain
+    self.searchgov_domain = SearchgovDomain.find_or_create_by(domain: URI(url).host) unless url.nil?
```
MothOnMars: `unless url.nil?` should not be necessary

spec/models/sitemap_spec.rb
```diff
@@ -0,0 +1,72 @@
+require 'spec_helper'
+
+describe Sitemap do
+  let!(:url) { 'http://www.agency.gov/boring' }
+  let!(:html) { read_fixture_file("/html/page_with_og_metadata.html") }
+  let!(:valid_attributes) { { url: url } }
+  let!(:sitemap) { Sitemap.new(valid_attributes) }
+  let!(:existing_domain) { SearchgovDomain.create!(domain: 'existing.gov') }
```
MothOnMars: As this variable is only used in one of your specs, it should be moved to that context.

spec/models/sitemap_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+      let!(:existing) { Sitemap.create!(valid_attributes) }
+
+      it { is_expected.to validate_uniqueness_of(:url).on(:create) }
+
+      it 'is case-sensitive' do
+        expect(Sitemap.new(url: 'https://www.agency.gov/BORING.xml')).to be_valid
+      end
+    end
+
+    context 'when validating url presence (not being nil)' do
```
MothOnMars: You can omit this, as this is covered by the validation spec in the `'a record with a fetchable url'` shared examples:
    `it { is_expected.to validate_presence_of :url }`

spec/models/sitemap_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+      let!(:attempt) { Sitemap.new }
+
+      it { is_expected.not_to be_valid }
+    end
+
+    context 'when validating last_crawl_status when it is > 255 characters' do
```
MothOnMars: As this `before_validation` check is in `Fetchable`, this spec should be in the shared examples for `'a record with a fetchable url'`.

It's best to use a single condition for a context; you can always nest multiple contexts. That keeps your specs flexible, as you may want to add additional nested contexts within the same parent context. And you should make sure that the conditions for your context are set in a `let` or `before` block, rather than in the spec itself. In this case, I would do something like:
```ruby
context 'when the last_crawl_status is > 255 characters' do
  let(:sitemap) { Sitemap.new(valid_attributes.merge(last_crawl_status: 'x' * 300) }

  it 'truncates the last_crawl_status' do
    expect(sitemap.valid?).to eq true
    expect(sitemap.last_crawl_status).to eq 'x' * 255
  end
end
```
(Note: normally I wouldn't use two `expect` lines in a single spec, but in this case I would need to call `sitemap.valid?` anyway before checking the `last_crawl_status` value, so making separate specs would be somewhat redundant.)

spec/models/sitemap_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+  let!(:url) { 'http://www.agency.gov/boring' }
+  let!(:html) { read_fixture_file("/html/page_with_og_metadata.html") }
+  let!(:valid_attributes) { { url: url } }
+  let!(:sitemap) { Sitemap.new(valid_attributes) }
+  let!(:existing_domain) { SearchgovDomain.create!(domain: 'existing.gov') }
+
+  it { is_expected.to have_readonly_attribute(:url) }
+
+  describe 'schema' do
+    it { is_expected.to have_db_column(:url).of_type(:string).with_options(null: false, limit: 2000) } # causing problems for I'm not sure why
```
MothOnMars: I'm not seeing any issues with this spec...?

app/concerns/fetchable.rb
```diff
@@ -90,4 +91,14 @@ def downcase_scheme_and_host_and_remove_anchor_tags
...
+  end
+
+  def truncate_last_crawl_status
+    if self.last_crawl_status && self.last_crawl_status.length > 255
```
MothOnMars: make this a one-liner per earlier comment

spec/models/sitemap_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+    it { is_expected.to have_db_column(:last_crawl_status).of_type(:string) }
+    it { is_expected.to have_db_column(:last_crawled_at).of_type(:datetime) }
+  end
+
+  describe 'associations' do
```
MothOnMars: let's create another set of shared examples in `fetchable_behavior.rb` such as `a record that belongs to SearchgovDomain`. Then we can remove those duplicate specs for SearchgovUrl and Sitemap.

app/models/sitemap.rb
```diff
@@ -0,0 +1,13 @@
+class Sitemap < ActiveRecord::Base
+  include Fetchable
+
+  attr_readonly :url
+
+  belongs_to :searchgov_domain
+
+  validates :url, presence: true, uniqueness: true,
+   length: {maximum: 2000}, format: {with: /\Ahttps?:\/\/(\w+\.)?\w+\.gov(\/\w+)*(\/|\.\w+)?\z/, message: 'invalid url'}
+  validates :last_crawl_status, length: {maximum: 255}
```
MothOnMars: let's move this validation to `Fetchable`, and make sure there's a corresponding spec in `a record that behaves like a fetchable URL`

app/models/sitemap.rb
```diff
@@ -0,0 +1,14 @@
...
+  include Fetchable
+
+  attr_readonly :url
+
+  belongs_to :searchgov_domain
+
+  validates_associated :searchgov_domain, on: :create
+  validates_presence_of :searchgov_domain, on: :create
+
+  before_validation :set_searchgov_domain, on: :create
```
MothOnMars: Sorry, I didn't catch this earlier. Because this is a `before_validation` callback, it makes logical sense to place this line before the validation lines in the code.

spec/support/fetchable_behavior.rb
```diff
@@ -96,6 +99,40 @@
         expect(described_class.create!(valid_attributes.merge(url: url)).url).to eq("http://www.nps.gov/hey/I/am/usagov/and/love/extra/slashes.shtml")
       end
     end
+
+    context "when URL doesn't have a protocol" do
+      let(:url) { "www.nps.gov/sdfsdf" }
+
+      it "should prepend it with https://" do
```
MothOnMars: This isn't your fault, as this spec was copied, but let's tweak this to [avoid using should](http://www.betterspecs.org/#should):
`it 'prepends it with "https://"'`

spec/support/fetchable_behavior.rb
```diff
@@ -136,3 +173,45 @@
     end
   end
 end
+
+shared_examples_for 'a record with an indexable url' do
+  describe 'validations' do
+    it { is_expected.to allow_value("http://some.site.gov/url").for(:url) }
```
MothOnMars: We really only need one spec for a valid URL; the following two spec lines can be removed.

MothOnMars: This spec should actually be in the `shared_examples_for 'a record with a fetchable url'` section.

spec/support/fetchable_behavior.rb
```diff
@@ -136,3 +173,45 @@
...
+
+      it "is not valid" do
+        expect(record).not_to be_valid
+        expect(record.errors.full_messages.first).to match(/extension is not one we index/i)
+      end
+    end
+  end
+end
+
+shared_examples_for 'a record with a searchgov_domain' do
```
MothOnMars: Please rename this to:
`shared_examples_for 'a record that belongs to a searchgov_domain'`

`a record with` seems to indicate a `has_one` relationship.

spec/support/fetchable_behavior.rb
```diff
@@ -136,3 +173,45 @@
...
+  describe 'associations' do
+    it { is_expected.to belong_to(:searchgov_domain) }
+
+    context 'on creation' do
+      context 'when the domain already exists' do
+        let!(:existing_domain) { SearchgovDomain.create!(domain: 'very_unique.gov') }
```
MothOnMars: For consistency, I would use `existing.gov` rather than `very_unique.gov`

spec/support/fetchable_behavior.rb
```diff
@@ -136,3 +173,45 @@
...
+        let!(:sitemap) { described_class.create!(url: 'https://very_unique.gov/foo.xmpl') }
```
MothOnMars: typo: `xmpl`

spec/support/fetchable_behavior.rb
```diff
@@ -136,3 +173,45 @@
...
+
+        it 'sets the searchgov domain' do
+          expect(sitemap.searchgov_domain).to eq(existing_domain)
+        end
+      end
+
+      context 'when the domain has not been created yet' do
+        it 'creates the domain' do
+          expect{ described_class.create!(url: 'https://brandnewdomain.gov/loop.xmpl') }.
```
MothOnMars: see above comment re. typo & URL path

app/models/searchgov_domain.rb
```diff
@@ -8,6 +8,7 @@ class SearchgovDomain < ActiveRecord::Base
   after_create { SearchgovDomainPreparerJob.perform_later(searchgov_domain: self) }
 
   has_many :searchgov_urls, dependent: :destroy
+  has_many :sitemaps
```
MothOnMars: whoops - this relationship should also be set to `dependent: :destroy` (🤦‍♀️ )

spec/support/fetchable_behavior.rb
```diff
@@ -136,3 +173,41 @@
...
+        expect(record).not_to be_valid
+        expect(record.errors.full_messages.first).to match(/extension is not one we index/i)
+      end
+    end
+  end
+end
+
+shared_examples_for 'a record that belongs to a searchgov_domain' do
+  describe 'associations' do
+    it { is_expected.to belong_to(:searchgov_domain) }
```
MothOnMars: @ethanewing , I realized we are missing an association spec in `searchgov_domain_spec.rb` for the `has_many` relationship to `sitemaps`. I'll add that:
```ruby
it { is_expected.to have_many(:sitemaps).dependent(:destroy) }
```
Strict TDD would have helped, because you would have written the spec before adding the `has_many` relationship in the `SearchgovDomain` model.

spec/models/sitemap_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+  let(:valid_attributes) { { url: url } }
+
+  it { is_expected.to have_readonly_attribute(:url) }
+
+  describe 'schema' do
+    it { is_expected.to have_db_column(:url).of_type(:string).with_options(null: false, limit: 2000) }
+  end
+
+  describe 'validations' do
+    context 'when the sitemap\'s domain is invalid' do
```
MothOnMars: No need to escape the apostrophe -- you can simply use double quotes:
```ruby
context "when the sitemap's domain is invalid" do
```
I'll clean this up, and move this spec to the `shared_examples_for 'a record that belongs to a searchgov_domain'` section.

spec/support/fetchable_behavior.rb
```diff
@@ -1,26 +1,27 @@
...
       record = described_class.new(valid_attributes.merge(url: ('x' * 2001) ))
       expect(record).not_to be_valid
       expect(record.errors[:url]).to include("is too long (maximum is 2000 characters)")
     end
 
-    context 'when the url extension is blacklisted' do
-      let(:movie_url) { "http://www.nps.gov/some.mov" }
-      let(:record) { described_class.new(valid_attributes.merge(url: movie_url)) }
+    context 'when last_crawl_status is > 255 characters' do
+      let(:temp) { described_class.create!(valid_attributes.merge(last_crawl_status: 'x' * 300)) }
```
MothOnMars: @ethanewing , I just noticed this. Any reason you named this variable `temp`?

Also, remember to avoid using `create` unless absolutely necessary, as persisting objects to the database slows down tests. I'll clean this up to simply validate but not save the record.

ethanewing: I think it was just a bad name I gave it. I have an old habit of naming variables that are only used once or within a small scope "temp" and I forgot to change it to something more descriptive. I'll be better about that and the unnecessary `create`s

MothOnMars: Ah, that makes sense, but certainly something to avoid in the future. (In the Ruby world, it's SOP to use almost obnoxiously descriptive method & variable names.)

spec/support/fetchable_behavior.rb
```diff
@@ -136,3 +173,41 @@
...
+        let!(:sitemap) { described_class.create!(url: 'https://existing.gov/foo') }
+
+        it 'sets the searchgov domain' do
+          expect(sitemap.searchgov_domain).to eq(existing_domain)
+        end
+      end
+
+      context 'when the domain has not been created yet' do
+        it 'creates the domain' do
+          expect{ described_class.create!(url: 'https://brandnewdomain.gov/loop') }.
```
MothOnMars: I also noticed that in the transfer of this spec from `searchgov_url_spec.rb` to `fetchable_behavior.rb`, the url changed from `https://brandnewdomain.gov/foo` to `https://brandnewdomain.gov/loop`. I will fix that; I'm guessing that was a text-editor-induced typo?

ethanewing: I think so!

PR #97, opened 2018-06-27
Feature/super admin sitemap

ethanewing:
Issue 95: Super Admins should be able to add and view sitemaps for searchgov domains

### DIFF
--- a/app/controllers/admin/searchgov_domains_controller.rb
+++ b/app/controllers/admin/searchgov_domains_controller.rb
@@ -1,13 +1,11 @@
 class Admin::SearchgovDomainsController < Admin::AdminController
   active_scaffold :searchgov_domain do |config|
     config.label = 'Search.gov Domains'
-    config.actions = %i[create list search export nested show]
+    config.actions = %i[create list search export nested]
     config.create.columns = [:domain]
     config.columns = %i[id domain status urls_count unfetched_urls_count created_at]
     config.nested.add_link(:searchgov_urls, label: "URLs", page: false)
-
-    config.nested.add_link(:sitemaps, label: "Sitemaps", page: false)
-    config.show.columns = [:domain, :status, :urls_count, :unfetched_urls_count, :created_at, :updated_at]
+    config.nested.add_link(:sitemaps, page: false)
   end
 
   def after_create_save(record)
--- a/app/controllers/admin/searchgov_domains_controller.rb
+++ b/app/controllers/admin/searchgov_domains_controller.rb
@@ -1,10 +1,17 @@
 class Admin::SearchgovDomainsController < Admin::AdminController
   active_scaffold :searchgov_domain do |config|
     config.label = 'Search.gov Domains'
-    config.actions = %i[create list search export nested]
+    config.actions = %i[create list search export nested show]
     config.create.columns = [:domain]
     config.columns = %i[id domain status urls_count unfetched_urls_count created_at]
     config.nested.add_link(:searchgov_urls, label: "URLs", page: false)
+
+    # config.nested.add_link(:sitemaps, label: "Sitemaps", page: false)
+    config.show.columns = [:domain, :status, :urls_count, :unfetched_urls_count, :created_at, :updated_at]
+    config.show.columns.add_subgroup 'Sitemaps' do |group|
+      # group.columns = [:url, :created_at, :updated_at]
+      group.add :sitemaps
+    end
   end
 
   def after_create_save(record)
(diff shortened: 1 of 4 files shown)


app/controllers/admin/searchgov_domains_controller.rb
```diff
@@ -1,10 +1,13 @@
 class Admin::SearchgovDomainsController < Admin::AdminController
   active_scaffold :searchgov_domain do |config|
     config.label = 'Search.gov Domains'
-    config.actions = %i[create list search export nested]
+    config.actions = %i[create list search export nested show]
     config.create.columns = [:domain]
     config.columns = %i[id domain status urls_count unfetched_urls_count created_at]
     config.nested.add_link(:searchgov_urls, label: "URLs", page: false)
+
+    config.nested.add_link(:sitemaps, label: "Sitemaps", page: false)
```
MothOnMars: @ethanewing , I think you can actually remove `label: "Sitemaps"` from these arguments. I added the custom `label: "URLs"` above because I didn't want the label to appear as `Searchgov Urls`. I *believe* ActiveScaffold will automatically set the `Sitemaps` label in this case.

app/controllers/admin/searchgov_domains_controller.rb
```diff
@@ -1,10 +1,13 @@
 class Admin::SearchgovDomainsController < Admin::AdminController
   active_scaffold :searchgov_domain do |config|
     config.label = 'Search.gov Domains'
-    config.actions = %i[create list search export nested]
+    config.actions = %i[create list search export nested show]
```
MothOnMars: You can skip adding the `show` link; I was hoping to include the sitemaps in that view, but now that we're displaying sitemaps separately, the `show` view doesn't show anything that's not already displayed in the list view.

app/controllers/admin/searchgov_domains_controller.rb
```diff
@@ -1,10 +1,13 @@
...
     config.create.columns = [:domain]
     config.columns = %i[id domain status urls_count unfetched_urls_count created_at]
     config.nested.add_link(:searchgov_urls, label: "URLs", page: false)
+
+    config.nested.add_link(:sitemaps, label: "Sitemaps", page: false)
+    config.show.columns = [:domain, :status, :urls_count, :unfetched_urls_count, :created_at, :updated_at]
```
MothOnMars: See above re. `show`

app/controllers/admin/sitemaps_controller.rb
```diff
@@ -0,0 +1,10 @@
+class Admin::SitemapsController < Admin::AdminController
+  active_scaffold :sitemap do |config|
+    config.label = "Sitemaps"
+    config.actions = [:create, :delete, :list]
+    config.columns = [:url, :created_at, :updated_at]
+    config.create.columns = [:url]
+    config.create.link.label = "Create Sitemap"
```
MothOnMars: You can skip the custom label here to be consistent with the implementation in other pages/sections in super admin, which just display the default 'Create New' label.

features/admin.feature
```diff
@@ -222,6 +225,14 @@ Feature:  Administration
     And I should not see "Delete"
     And I should see "search.gov"
 
+    When I follow "Sitemaps"
+    Then I should see "search.gov/sitemap.xml"
+    And I follow "Delete" and confirm "Are you sure you want to delete this sitemap?"
```
MothOnMars: After this line, I would add:
`Then I should not see "search.gov/sitemap.xml"`
to confirm the sitemap was deleted.

features/admin.feature
```diff
@@ -234,6 +245,7 @@ Feature:  Administration
     And I fill in "Domain" with "www.state.gov"
     And I press "Create"
     Then I should see "www.state.gov has been created"
+    And I follow "Close"
```
MothOnMars: ?

features/admin.feature
```diff
@@ -222,6 +225,14 @@ Feature:  Administration
     And I should not see "Delete"
     And I should see "search.gov"
 
+    When I follow "Sitemaps"
+    Then I should see "search.gov/sitemap.xml"
+    And I follow "Delete" and confirm "Are you sure you want to delete this sitemap?"
+    Then I follow "Create Sitemap"
```
MothOnMars: The `Then` line I mentioned above concludes the deletion test. Because you're now starting a creation test, I would change `Then` to `When`:
`When I follow "Create New"`

(Note that you will need to use `Create New` instead of `Create Sitemap`, as we're not using a custom create label.

ethanewing: The reason I had originally changed the label to "Create Sitemap" instead of going with the default "Create New" was because when running this scenario, which tests that new `searchgov_domains` can be created and that new `sitemaps` can be created, cucumber was giving an "ambiguity error" because there were two lines that said "When I follow 'Create New'".  Is there another way around this then changing the label of the button to create sitemaps, besides testing that feature in a different scenario?

PR #100, opened 2018-07-02
Feature/background fetch super admin

ethanewing:
Not finished: searchgov_url_fetcher_job still WIP


features/admin.feature
```diff
@@ -226,7 +226,8 @@ Feature:  Administration
     Then I should see "search.gov/oops"
     And I follow "Fetch"
     And I wait for ajax
-    Then I should see "404"
+    Then I should see "Your URL has been added to the fetching queue"
+    And I follow "Close" within "div.info-message.message"
```
MothOnMars: @ethanewing , remember to keep cucumber features as 'human-readable' as possible. For `within` steps, you'll need to define one a 'human-readable' page area. See https://github.com/GSA/search-gov/blob/master/features/step_definitions/within_steps.rb

In this case, you might say something like `within the first URLs row`. I suspect that `div.info-message.message` isn't the best CSS id for that; let me know if you need help finding a better identifier.

spec/jobs/searchgov_url_fetcher_job.rb
```diff
@@ -0,0 +1,30 @@
+require 'spec_helper'
+
+describe SearchgovUrlFetcherJob do
+  subject(:perform) { SearchgovUrlFetcherJob.perform_now(args[:searchgov_url]) }
```
MothOnMars: Because you aren't using [keyword arguments](https://robots.thoughtbot.com/ruby-2-keyword-arguments) for this job, you can just say:
`subject(:perform) { SearchgovUrlFetcherJob.perform_now(searchgov_url) }`
...and remove the `args` variable.

spec/jobs/searchgov_url_fetcher_job.rb
```diff
@@ -0,0 +1,30 @@
...
+  let!(:searchgov_url) { SearchgovUrl.create(url: 'https://agency.gov/') }
+  let(:args) do
+    { searchgov_url: searchgov_url }
+  end
+
+  it_behaves_like 'a searchgov job'
```
MothOnMars: 🌟 👍

spec/jobs/searchgov_url_fetcher_job.rb
```diff
@@ -0,0 +1,30 @@
+require 'spec_helper'
+
+describe SearchgovUrlFetcherJob do
+  subject(:perform) { SearchgovUrlFetcherJob.perform_now(args[:searchgov_url]) }
+  let!(:searchgov_url) { SearchgovUrl.create(url: 'https://agency.gov/') }
```
MothOnMars: are you sure you need the bang method here?

spec/jobs/searchgov_url_fetcher_job.rb
```diff
@@ -0,0 +1,30 @@
...
+  describe '#perform' do
+    it 'must have one parameter' do
+      expect{ SearchgovUrlFetcherJob.perform_now }.
+        to raise_error(ArgumentError)
+    end
+
+    it 'fetches a searchgov_url' do
+      # byebug
+      perform
+      expect(searchgov_url.last_crawl_status).to_not be(nil)
```
MothOnMars: We don't really need the record to be fetched in our test, nor do we need the test to know anything about what `fetch` does (such as changing the `last_crawl_status`). A cleaner (and faster) way to test this would be to use a mock, i.e.:
```ruby
expect(searchgov_url).to receive(:fetch)
perform
```

spec/jobs/searchgov_url_fetcher_job.rb
```diff
@@ -1,7 +1,7 @@
 require 'spec_helper'
 
 describe SearchgovUrlFetcherJob do
-  subject(:perform) { SearchgovUrlFetcherJob.perform_now(args[:searchgov_url]) }
+  subject(:perform) { SearchgovUrlFetcherJob.perform_now(searchgov_url) }
   let!(:searchgov_domain) { SearchgovDomain.create(domain: 'agency.gov', status: '200') }
   let(:searchgov_url) { SearchgovUrl.create(url: 'https://agency.gov/') }
   let(:args) do
```
MothOnMars: Per my earlier comment, you can remove the `args` variable, which you are no longer using.

spec/jobs/searchgov_url_fetcher_job.rb
```diff
@@ -1,7 +1,7 @@
 require 'spec_helper'
 
 describe SearchgovUrlFetcherJob do
-  subject(:perform) { SearchgovUrlFetcherJob.perform_now(args[:searchgov_url]) }
+  subject(:perform) { SearchgovUrlFetcherJob.perform_now(searchgov_url) }
   let!(:searchgov_domain) { SearchgovDomain.create(domain: 'agency.gov', status: '200') }
```
MothOnMars: I don't see this variable used anywhere...?

features/step_definitions/within_steps.rb
```diff
@@ -60,7 +60,8 @@
   'in the twitter govbox' => '#twitter_govbox',
   'in the site header' => '.l-site-header',
   'in the active site main navigation' => '.l-site-nav.main .active',
-  'in the active site sub navigation' => '.l-site-nav.sub .active'
+  'in the active site sub navigation' => '.l-site-nav.sub .active',
+  'in the SearchgovDomain URLs table' => 'tbody.messages #as_admin__searchgov_urls-messages'
```
MothOnMars: @ethanewing , that selector appears to be for just the section that holds the flash messages for that table. As we may want to use the `in the SearchgovDomain URLs table` selector more generically, I'd try the `as_admin__searchgov_urls-active-scaffold` div.

spec/jobs/searchgov_url_fetcher_job.rb
ethanewing: Did you mean to remove the `args` variable from the `SearchgovUrlFetcherJob.perform_now(args[:searchgov_url])` line, from this spec altogether, or both (I've so far only removed it from the `*.perform_now` statement)?

spec/jobs/searchgov_url_fetcher_job.rb
ethanewing: I believe the `args` variable is necessary for the shared example, `it_behaves_like 'a searchgov job'`, to work correctly?

spec/jobs/searchgov_url_fetcher_job.rb
ethanewing: I've removed this variable.  It seemed to be necessary before when I was trying to test the fetching of a record, but now that I've changed that to what you recommended it isn't needed anymore.

spec/jobs/searchgov_url_fetcher_job.rb
MothOnMars: Ah, you are correct. I was thinking you could just pass the `searchgov_url` record as the argument, rather than a hash, because you are not using keyword arguments.

However...let's change that. I am not always consistent about using keyword args, but I do like how explicit they are about complaining when you are missing an argument. Using keyword args would also be consistent with the other job arguments:
```
$ gg 'def perform' app/jobs/
app/jobs/searchgov_domain_indexer_job.rb:  def perform(searchgov_domain:, delay:)
app/jobs/searchgov_domain_preparer_job.rb:  def perform(searchgov_domain:)
app/jobs/sitemap_indexer_job.rb:  def perform(searchgov_domain:)
```
So, instead of:
`def perform(searchgov_url)`

You'd use:
`def perform(searchgov_url:)`

And the way to initiate that job would be:
`SearchgovUrlFetcherJob.perform_now(searchgov_url: searchgov_url)`

It's a little more wordy than just `SearchgovUrlFetcherJob.perform_now(searchgov_url)`, but it gives us more flexibility in case we need to add  more args down the line. Sorry for making this specification late in the game; I was thinking I didn't mind the inconsistency with the other jobs, but I do think we should make this change.

spec/jobs/searchgov_url_fetcher_job.rb
MothOnMars: bump

ethanewing: I think I do, as the tests don't seem to pass if I don't have it.  Maybe there's a better way of organizing it?

spec/jobs/searchgov_url_fetcher_job.rb
ethanewing: Gotcha.  So I now have it that `perform` uses a keyword argument.  Since the `args` variable is still necessary, would it be cleaner to write `SearchgovUrlFetcherJob.perform_now(args)` in `subject` declaration in the spec?

MothOnMars: yep

spec/jobs/searchgov_url_fetcher_job.rb
MothOnMars: The change you made is what I was looking for 👍

spec/controllers/searchgov_urls_controller_spec.rb
```diff
@@ -1,26 +1,27 @@
...
+        allow(SearchgovDomainPreparerJob).to receive(:perform_later)
+      end
 
-    it 'enqueues a searchgov_url_fetcher_job to the searchgov queue' do
-      # ActiveJob::Base.queue_adapter = :test
-      expect{ post :fetch, params }.
-        to have_enqueued_job(SearchgovUrlFetcherJob).on_queue("searchgov")
+      it 'enqueues a searchgov_url_fetcher_job to the searchgov queue' do
+        expect{ post :fetch, params }.
+          to have_enqueued_job(SearchgovUrlFetcherJob).on_queue("searchgov")
```
MothOnMars: @ethanewing , TDD is helpful for bug fixes as well. This test doesn't actually test this change, because it's not checking the job arguments: https://github.com/GSA/search-gov/commit/e7b3adb.

spec/controllers/searchgov_urls_controller_spec.rb
```diff
@@ -1,26 +1,27 @@
...
 
-  describe '#fetch' do
-    # before do
-    #   # post :fetch, association: 'searchgov_urls', parent_scaffold: 'admin/searchgov_domains',
-    #   # searchgov_domain_id: searchgov_url.searchgov_domain.id, id: searchgov_url.id
-    #   # post "/admin/searchgov_domains/#{searchgov_url.searchgov_domain.id}/searchgov_urls/#{searchgov_url.id}/fetch?_method=post&association=searchgov_urls&parent_scaffold=admin%2Fsearchgov_domains&searchgov_domain_id=#{searchgov_url.searchgov_domain.id}"
-    #   post :fetch, params
-    #   p response.body
-    # end
+  context 'when logged in as an admin' do
```
MothOnMars: @ethanewing , you should be able to use the shared `'super admin logged in'` context for this:
https://github.com/GSA/search-gov/blob/master/spec/support/shared_sites_controller_shared_behavior.rb#L46
https://relishapp.com/rspec/rspec-core/docs/example-groups/shared-context

spec/controllers/searchgov_urls_controller_spec.rb
ethanewing: Would that test to make sure it only takes named arguments go in `searchgov_url_fetcher_job_spec.rb` since that test is specifically about the functionality of the `perform` function within the model?  I've already got a test in that spec that tests that the `perform` is invalid if it isn't given an argument -- should I just add another that tests that it's invalid if it's not given a named argument?

MothOnMars: This test is covering the bug you fixed with https://github.com/GSA/search-gov/commit/e7b3adb. It ensures we're passing the args correctly to the job. You can do that by specifying the args in your test:
```ruby
expect{ post :fetch, params }.
  to have_enqueued_job(SearchgovUrlFetcherJob).on_queue("searchgov").
  with(searchgov_url: searchgov_url)
```

It couldn't hurt to also make your `searchgov_url_fetcher_job_spec.rb` spec more specific, by testing that it requires a`searchgov_url` keyword argument.

spec/controllers/searchgov_urls_controller_spec.rb
```diff
@@ -0,0 +1,27 @@
+require 'spec_helper'
+
+describe Admin::SearchgovUrlsController do
+  fixtures :users
+
+  let(:admin) { users(:affiliate_admin) }
+  let(:searchgov_url) { SearchgovUrl.create(url: 'agency.gov/test') }
```
MothOnMars: Try to use bang methods in your tests, to ensure that your setup steps don't fail silently: `SearchgovUrl.create!(url: 'agency.gov/test')`

spec/jobs/searchgov_url_fetcher_job_spec.rb
```diff
@@ -0,0 +1,23 @@
+require 'spec_helper'
+
+describe SearchgovUrlFetcherJob do
+  subject(:perform) { SearchgovUrlFetcherJob.perform_now(args) }
+  let!(:searchgov_url) { SearchgovUrl.create(url: 'https://agency.gov/') }
```
MothOnMars: See above comment re. bang methods

spec/jobs/searchgov_url_fetcher_job_spec.rb
```diff
@@ -15,6 +15,11 @@
         to raise_error(ArgumentError)
     end
 
+    it 'must have a named argument' do
+      expect { SearchgovUrlFetcherJob.perform_now searchgov_url }.
+        to raise_error(ArgumentError)
```
MothOnMars: The more specific you can be in your specs, the better. Here, you can specify the error message as well as the error class:
`expect....to raise_error(ArgumentError, 'missing keyword: searchgov_url')`

spec/controllers/searchgov_urls_controller_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+  let(:admin) { users(:affiliate_admin) }
+  let(:searchgov_url) { SearchgovUrl.create!(url: 'agency.gov/test') }
+  let(:params) do
+    { association: 'searchgov_urls', parent_scaffold: 'admin/searchgov_domains',
+    searchgov_domain_id: searchgov_url.searchgov_domain.id, id: searchgov_url.id }
+  end
+
+  include_context 'super admin logged in' do
+    describe '#fetch' do
+
```
MothOnMars: @ethanewing - I just noticed this. Please remove this extra line break

spec/controllers/searchgov_urls_controller_spec.rb
```diff
@@ -0,0 +1,23 @@
+require 'spec_helper'
+
+describe Admin::SearchgovUrlsController do
+  fixtures :users
+
+  let(:admin) { users(:affiliate_admin) }
```
MothOnMars: @ethanewing , now that you're using the shared context, this line can be removed

spec/controllers/searchgov_urls_controller_spec.rb
```diff
@@ -3,7 +3,6 @@
 describe Admin::SearchgovUrlsController do
   fixtures :users
```
MothOnMars: whoops - I missed this too! You can remove this line and the line break following.

ethanewing: My specs are failing when I comment that line out, unless I transfer that line (`fixtures :users`) into the shared context itself.

MothOnMars:
@ethanewing , please move `spec/controllers/searchgov_urls_controller_spec.rb` to `spec/controllers/admin/searchgov_urls_controller_spec.rb`

PR #101, opened 2018-07-03
re-record Bing cassettes

MothOnMars:
- stop re-recording BingV6 cassettes
- set re_record_interval for Bing-related API specs to 4 months


noremmie:
LGTM. I'll set my doomsday alarm to four months from now 👍

PR #103, opened 2018-07-04
[#158131076] follow client-side redirects for searchgov urls

MothOnMars:
- fix HtmlDocument#redirect_url method for quoted paths

### DIFF
--- a/app/models/html_document.rb
+++ b/app/models/html_document.rb
@@ -19,8 +19,8 @@ class HtmlDocument < WebDocument
   def redirect_url
     refresh = html.css('meta[http-equiv]').detect{|node| /refresh/i === node['http-equiv'] }
     if refresh
-      new_path = refresh['content'].match(/.*URL=('|")?(?<path>[^('|")]*)/i)[:path]
-      URI.join(url, new_path).to_s
+      new_path = refresh['content'].match(/.*URL=['"]?(?<path>[^'"]*)/i)[:path]
+      URI.join(url, URI.encode(new_path)).to_s
     end
   end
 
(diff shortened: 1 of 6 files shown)


app/models/html_document.rb
```diff
@@ -18,8 +18,10 @@ def keywords
   # Returns client-side redirect url
   def redirect_url
     refresh = html.css('meta[http-equiv]').detect{|node| /refresh/i === node['http-equiv'] }
-    new_path = refresh.try(:[],'content')&.gsub(/.*url=/i,'')
-    URI(url).merge(URI(new_path)).to_s if new_path
+    if refresh
+      new_path = refresh['content'].match(/.*URL=('|")?(?<path>[^('|")]*)/i)[:path]
```
noremmie: I think this regex is more complicated than it needs to be, and the character class towards the end that is meant to only exclude single and double quotes is also excluding parentheses and pipes. Parentheses and pipes inside a character class lose their special meaning. Here you can see the regex not capturing the full path:

```
irb(main):001:0> "URL='./new().html'".match(/.*URL=('|")?(?<path>[^('|")]*)/i)[:path]
=> "./new"
```

Using just a character class and no groups is simpler and more correct:

```
irb(main):002:0> "URL='./new().html'".match(/.*URL=['"]?(?<path>[^'"]*)/i)[:path]
=> "./new().html"
```

It would be good to see some additional specs to show that the regex is doing what's intended.

MothOnMars: You have revealed my ignorance; I thought `(a|b`) meant `'a' or 'b'`, not `'a' or '|' or 'b'`. I will make those changes, then check "Regexes for Dummies" out of the library.

Issue #105, opened 2018-07-04
RssFeedUrl#last_crawl_status value should be truncated before validation


MothOnMars:
We're seeing a lot of errors in our logs for `last_crawl_status` error messages that are too long for that column in the database:
```
Mysql2::Error: Data too long for column 'last_crawl_status' at row 1: UPDATE `rss_feed_urls` SET `last_crawl_status` = 'Mysql2::Error: Data too long for column \'last_crawl_status\' at row 1: UPDATE `rss_feed_urls` SET `last_crawl_status` = \'redirection forbidden: http://www.technology.wv.gov/_layouts/feed.aspx?xsl=1&web=%2Fnews&page=37e69109-afb9-4fb4-b956-bd41f735669e&wp=91e42068-5876-4e0f-8606-2161b12e7530 -> http://technology.wv.gov/_layouts/feed.aspx?xsl=1&web=%2Fnews&page=37e69109-afb9-4fb4-b956-bd41f735669e&wp=91e42068-5876-4e0f-8606-2161b12e7530\', `updated_at` = \'2018-04-17 18:02:33\' WHERE `rss_feed_urls`.`id` = 18790', `updated_at` = '2018-04-17 18:02:33' WHERE `rss_feed_urls`.`id` = 18790
```

We should truncate that value to 255 characters before validation.

**Bug steps:**
Attempt to save or create an `rss_feed_url` record with a `last_crawl_status` longer than 255 characters
```ruby
too_long_status = 'x' * 500
RssFeedUrl.create!(rss_feed_owner_type: 'Affiliate',
  url: 'https://www.usa.gov/rss/updates.xml',
  last_crawl_status: too_long_status)
```

**Expected Results:**
- `last_crawl_status` is truncated to 255 characters, and the record is successfully saved to the database

**Actual Results:**
- the above MySQL error is raised

MothOnMars:
@ethanewing , I'm thinking this would be a good opportunity to extend ActiveRecord itself to add a `truncate_value` method that could be used generically within other ActiveRecord classes, taking a field name & length as arguments. Something like (untested code to follow...):
```ruby
def truncate_value(field, length_limit)
  self.send(:field) = self.send(field)&.truncate(limit)
end
```

Then any class with a value that requires truncation could do use that in a callback:
`before_validation :truncate_value, :foo, 255`

See https://stackoverflow.com/a/2329394/1020168 for info on extending ActiveRecord.

PR #106, opened 2018-07-09
Feature/rss feed truncate validation

ethanewing:
Resolves https://github.com/GSA/search-gov/issues/105

### DIFF
--- a/app/models/rss_feed_url.rb
+++ b/app/models/rss_feed_url.rb
@@ -1,5 +1,4 @@
 class RssFeedUrl < ActiveRecord::Base
-  include ActiveRecordExtension
   OK_STATUS = 'OK'
   PENDING_STATUS = 'Pending'
   STATUSES = [OK_STATUS, PENDING_STATUS]
@@ -11,7 +10,7 @@ class RssFeedUrl < ActiveRecord::Base
 
   before_validation NormalizeUrl.new(:url), on: :create
   before_validation do
-    truncate_value(:last_crawl_status, 255)
+    truncate_value(:last_crawl_status, 255) if last_crawl_status.length > 255
   end
 
   validates_presence_of :rss_feed_owner_type, :url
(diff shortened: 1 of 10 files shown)


app/models/rss_feed_url.rb
```diff
@@ -10,6 +9,9 @@ class RssFeedUrl < ActiveRecord::Base
   before_destroy :blocking_destroy_news_items
 
   before_validation NormalizeUrl.new(:url), on: :create
+  before_validation do
+    truncate_value(:last_crawl_status, 255) if last_crawl_status.length > 255
```
MothOnMars: No need for the `if last_crawl_status.length > 255`.

spec/models/rss_feed_url_spec.rb
```diff
@@ -115,6 +115,20 @@
         end
       end
     end
+
+    context 'when the last_crawl_status is greater than 255 characters' do
+      before do
+        rss_feed_url.update(last_crawl_status: 'x' * 500)
```
MothOnMars: I would change this so that you are not making any database calls, which always slow down tests. Instead of updating a database record, you could change the context to:
let(:rss_feed_url) { RssFeedUrl.new(last_crawl_status: 'x' * 500) }

Then your test can be changed to:
`expect{ rss_feed_url.valid? }.to change{ rss_feed_url.length }.from(500).to(255)`

I like to use `to change` matchers because it ensures that your test setup is correct.

spec/models/rss_feed_url_spec.rb
```diff
@@ -115,6 +115,20 @@
...
+      end
+
+      it 'should truncate the value to 255 characters' do
```
MothOnMars: Remember to avoid using `should`: http://www.betterspecs.org/#should

spec/models/rss_feed_url_spec.rb
```diff
@@ -115,6 +115,20 @@
...
+        expect(rss_feed_url.last_crawl_status.length).to eq(255)
+      end
+
+      it 'should be valid' do
```
MothOnMars: I think we can skip this test, as we are using truncation rather than a length validation for `last_crawl_status`.

PR #108, opened 2018-07-12
Mct/domain states 157986614

MothOnMars:


MothOnMars:
@noremmie , this commit is intended to track domain indexing activity, and to put the kibosh on attempts to trigger multiple indexing jobs for the same domain. At the moment, the states are limited to `idle` and `indexing`, but that will eventually include states such as `crawling` or `reindexing`.

Basic usage:
```ruby
sd = SearchgovDomain.find_or_create_by(domain: 'search.gov')
sd.activity #=> 'idle'

sd.index_urls #this triggers a state transition to 'indexing'
# Enqueued SearchgovDomainIndexerJob...

sd.index_urls
# www.bop.gov is already being indexed
```
This allows any process that creates URLs to call `index_urls` safely. A warning will be logged, but no duplicate indexing job will be enqueued. When the domain indexing job is complete, it sets the state back to `idle`. (Note: I'm deliberately not doing any error handling in the indexer job to set the state back to `idle` if something goes wrong. Once I have a handle on all the hiccups therein, I'll make those changes, but for now I feel it's sufficient for that job to bark loudly via Airbrake.)

MothOnMars:
@noremmie , I should have mentioned that. The AASM bang methods *without* bang simply set the relevant attribute, but do not commit the change. The bang methods actually update the record.
https://github.com/aasm/aasm#bang-events

noremmie:
That makes perfect sense, @MothOnMars. Thanks!

PR #111, opened 2018-07-17
[#157884284] force HTML encoding to UTF-8

MothOnMars:


spec/models/html_document_spec.rb
```diff
@@ -271,6 +271,14 @@
       it 'omits the invalid bytes' do
         expect(parsed_content).to eq 'invalid bytes'
       end
+
+      context 'when the html can be force-encoded to UTF-8' do
+        let(:raw_document) { open('https://www.ssa.gov/espanol/beneficios/ssi/').read }
```
noremmie: How would you feel about adding the content of this page to a file in `spec/fixtures/html` rather than making a request for it that VCR re-records occasionally? I ask mainly to avoid that occasional request but also because if this site eventually fixes their character encoding problem, then this spec wouldn't be testing the new logic you're adding.

MothOnMars: We are not re-recording any cassettes any more, so that shouldn't be an issue. However, it's still the case that it would be re-recorded if, say, someone reworded the context or the spec, so that's a reasonable change to make.

PR #112, opened 2018-07-18
Fixes rspec tests that fail with a clean test db

HalahAb:


HalahAb:
Resolves #59

spec/engines/search_engine_spec.rb
```diff
@@ -17,6 +17,9 @@ def parse_search_engine_response(_)
 end
 
 describe SearchEngine do
+  before(:all) do
+    CachedSearchApiConnectionResponse = Struct.new(:response, :cache_namespace)
```
MothOnMars: @HalahAb , I see the issue with this spec, but I don't think that redefining the struct in the spec is the best approach. I think a safer fix would be to move the struct definition to `lib/cached_search_api_connection_response.rb`, to ensure that that class is auto-loaded.

lib/cached_search_api_connection_response.rb
```diff
@@ -0,0 +1,3 @@
+# :nocov:
```
MothOnMars: @HalahAb , please add a spec for this instead of skipping code coverage.

HalahAb: @MothOnMars, I added a test then had to remove it because that one line was still being seen as missing coverage. I couldn't figure out why and thought to remove it because it was only testing ruby struct functionality. I added the test back now, but it turns out it was the other struct definition in cached_search_api_connection.rb that was causing the coverage fail. I removed it and that seems to have fixed things.

spec/lib/cached_search_api_connection_response_spec.rb
```diff
@@ -0,0 +1,10 @@
+require 'spec_helper'
+
+describe CachedSearchApiConnectionResponse do
+  let(:cached_response) {  CachedSearchApiConnectionResponse.new(:response, :cache_name) }
```
MothOnMars: please remove extra space: `{  CachedSearchApiConnectionResponse...`

spec/models/image_search_spec.rb
```diff
@@ -24,8 +24,9 @@
     let(:use_commercial_results) { nil }
     let(:affiliate) { affiliates(:basic_affiliate) }
     let(:search_engine) { nil }
-    before { allow(affiliate).to receive(:search_engine).and_return(search_engine) }
     before do
+      allow(affiliate).to receive(:search_engine).and_return(search_engine)
+      allow(affiliate).to receive(:has_no_social_image_feeds?).and_return  false
```
MothOnMars: Please remove extra space: `and_return  false`

PR #116, opened 2018-07-20
[#157717611] remove SearchgovUrl tempfiles after indexing

MothOnMars:


app/models/searchgov_url.rb
```diff
@@ -85,11 +87,13 @@ def get_response
   end
 
   def download
-    file = Tempfile.open("SearchgovUrl:#{Time.now.to_i}", Rails.root.join('tmp'))
-    file.binmode
-    body = response.body
-    file.write body.readpartial until (file.write body.readpartial) == 0
-    file
+    @tempfile = begin
```
noremmie: Would this be a good candidate for memoization?

MothOnMars: That was my initial inclination, but I don't want to retry the download if the tempfile is nil.

noremmie: I don't get how a download would be retried if `@tempfile` is nil. If `@tempfile` is nil, then doesn't referencing `tempfile` simply return nil given the `attr_reader :tempfile` at the top of the class? I get how calling `download&.close!` would cause the file to be downloaded again even if it's already downloaded. But I don't get how calling `tempfile&.close!` would result in any download occurring at all.

MothOnMars: Sorry, you are right. `squashme` coming up.

noremmie:
4e421f7098f31aec71451326204b6a7df5321c8a LGTM too 👍

PR #119, opened 2018-07-24
Load all fixtures before rspec test suite is run.

HalahAb:


HalahAb:
Resolves #59

spec/requests/api/c/rss_search_spec.rb
```diff
@@ -5,7 +5,7 @@
 
   let(:affiliate) { affiliates(:basic_affiliate) }
 
-  before(:all) do
+  before(:each) do
```
MothOnMars: you can simplify this as `before do` (which is equivalent to `before(:each)`

spec/spec_helper.rb
```diff
@@ -51,11 +51,11 @@
     c.syntax = [:expect]
   end
 
-  # This prevents affiliate-related tests from failing with an empty test db
+  # This prevents fixture related fails with an empty test db
```
MothOnMars: @HalahAb , let's simplify this entire comment to be:
`# Create all fixtures in spec/fixtures/*.yml for all tests in alphabetical order.`

HalahAb: should I get rid of the comment about the language fixtures as well? or should we keep it for future reference?

PR #120, opened 2018-07-31
fixes spec dependency between rss_feed_data_spec.rb and update_https_spec.rb

HalahAb:
Resets RssFeedUrl's readonly attributes which are modified after invoking  the rake task in update_https_spec.rb, and ensures 'url' is not read only.

### DIFF
--- a/spec/lib/importers/rss_feed_data_spec.rb
+++ b/spec/lib/importers/rss_feed_data_spec.rb
@@ -421,6 +421,7 @@ describe RssFeedData do
         before do
           rss_feed_url.news_items.destroy_all
           allow(DocumentFetcher).to receive(:fetch).with(rss_feed_url.url, an_instance_of(Hash)).and_return({ status: "301", body: rss_feed_content, last_effective_url: "https://www.whitehouse.gov/feed/blog/white-house" })
+          allow(RssFeedUrl).to receive(:readonly_attributes).and_return(%w{ rss_feed_owner_type })
         end
 
         it 'updates the url' do


HalahAb:
Resolves #59

MothOnMars:
@Halah, this is an interesting case. You’ve actually found a bug in the code, rather than a test issue. 

The history of the bug you’ve found is that at the time the `update_https` rake task was written (https://github.com/GSA/search-gov/commit/ffa833f0bac0fbdb1eca5c845a627605ced924ed), `url` was a read-only attribute for the `RssFeedUrl` class. So that task removed and then added the `url` attribute to the `readonly_attributes` list (https://github.com/GSA/search-gov/commit/ffa833f0bac0fbdb1eca5c845a627605ced924ed#diff-49d7b19871b49d356286d27431031aedR76).

A subsequent commit made `RssFeedUrl#url` *not* read-only (https://github.com/GSA/search-gov/commit/e0c5ea0317b1c16718b686af3d137d4ddff6807b#diff-3725379569576e16e22f4f689378dab1L7). So that bug you found is that the `update_https` task is setting a read-only attribute that should *not* be read-only, which is breaking `rss_feed_data_spec.rb`.  That’s actually a dangerous bug, because it means that the rake task is erroneously setting the `readonly_attributes` set for `RssFeedUrl`.

The fix is two-fold:
1) update the `update_https` rake task to stop toggling the read-only attributes for `RssFeedUrl` (because `url` is no longer read-only for that class
2) prevent any similar problems for `FlickrProfile` or any other classes. Instead of setting a static array of read-only attributes for the `FlickrProfile` and `RssFeedUrl` classes, we should make the code more flexible so that it does not depend on any knowledge of the classes it’s updating. I’m going to close this PR and push a fix for that, because I want to revisit that code and make sure there’s nothing I’m missing in that approach.

PR #121, opened 2018-07-31
[#158922908] fix spec order dependency

MothOnMars:
- prevent update_https rake task from erroneously setting read-only
attributes

### DIFF
--- a/lib/tasks/update_https.rake
+++ b/lib/tasks/update_https.rake
@@ -11,12 +11,13 @@ namespace :usasearch do
       @secure_hosts = @file.readlines.map(&:strip)
       @insecure_hosts = []
       @updated_records = 0
-      toggle_url_readonly(:off) if [FlickrProfile,RssFeedUrl].include?(@klass)
+      @readonly_attributes = @klass.readonly_attributes
+      set_readonly_attributes(@readonly_attributes - ['url'])
       @klass.where("#{@column} != '' AND #{@column} not like 'https%'").find_each do |record|
         httpsify(record)
       end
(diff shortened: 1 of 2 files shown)


MothOnMars:
@noremmie , see https://github.com/GSA/search-gov/pull/120#issuecomment-409349998 for an explanation of this fix.

lib/tasks/update_https.rake
```diff
@@ -11,12 +11,13 @@ namespace :usasearch do
...
       @updated_records = 0
-      toggle_url_readonly(:off) if [FlickrProfile,RssFeedUrl].include?(@klass)
+      @readonly_attributes = @klass.readonly_attributes
+      set_readonly_attributes(@readonly_attributes - ['url'])
       @klass.where("#{@column} != '' AND #{@column} not like 'https%'").find_each do |record|
         httpsify(record)
       end
     ensure
-      toggle_url_readonly(:on) if [FlickrProfile,RssFeedUrl].include?(@klass)
+      set_readonly_attributes(@readonly_attributes)
```
noremmie: It's technically possible for `@readonly_attributes` to be nil in the case where an exception is thrown when reading the file. It would be safer to set `@readonly_attributes` higher up in the block (e.g., right after `@klass` is created).

However, since this is a rake task, I don't think it matters that the read-only attributes might be set to the wrong thing at the very end. The change is occurring only in the process running the rake task, and that process ends immediately after the rake task finishes.

So I don't think this needs to be changed, but I'll 👍 this PR and leave it up to you.

PR #122, opened 2018-07-31
Update video link

dawnpm:
I updated the getting started video that we include in the Welcome emails, to include both branding updates and a notice that before they launch we want to talk to them about indexing. This PR is just a tiny update to point at the new video instead of the old one.

### DIFF
--- a/db/email_templates/welcome_to_new_user_added_by_affiliate.html.erb
+++ b/db/email_templates/welcome_to_new_user_added_by_affiliate.html.erb
@@ -738,7 +738,7 @@ The first site you have access to is:<br>
         <td class="mcnImageCardTopImageContent" align="center" valign="top" style="padding-top:9px; padding-right:18px; padding-bottom:18px; padding-left:18px;">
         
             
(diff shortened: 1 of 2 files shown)


db/email_templates/welcome_to_new_user_added_by_affiliate.html.erb
```diff
@@ -738,7 +738,7 @@ The first site you have access to is:<br>
         <td class="mcnImageCardTopImageContent" align="center" valign="top" style="padding-top:9px; padding-right:18px; padding-bottom:18px; padding-left:18px;">
         
             
-            <a href="https://www.youtube.com/watch?v=TnlpuudK_WY" title="Getting Started With DigitalGov Search" class="" target="_blank">
```
noremmie: Sorry I missed this DigitalGov reference earlier!

dawnpm: NP! It was the title of the video over on YouTube anyway.

PR #126, opened 2018-08-06
Fixes spec order dependencies.

HalahAb:


spec/helpers/twitter_profiles_helper_spec.rb
```diff
@@ -21,6 +21,7 @@
     before do
       @affiliate = affiliates(:usagov_affiliate)
       @search_vertical = :web
+      I18n.locale = 'en'
```
MothOnMars: @HalahAb , I don't think the two `I18n.locale = 'en'` additions are necessary in this spec. As long as any specs that set the locale to something other than `en` set it back to `en` using an `after` block (as you have done in the other specs), you shouldn't need to explicitly set it to `en` in any specs, because `en` is the `default_locale`.  Unless I'm missing something?

HalahAb: Oh, no your not. I wasn't sure at that point if I had to change both since the test was explicitly specifying locale, your right it's a redundant change. I'll remove :)

PR #127, opened 2018-08-09
Ea/phone number

ErikSArnold:
I updated the phone number in the code. If there is anywhere else it could live outside of these 16 files, let me know. Thanks!


noremmie:
This LGTM. I'll merge this and roll it out after coordinating with @dawnpm.

PR #131, opened 2018-08-13
Fixes spec order dependencies.

HalahAb:
Ensures that RoutedQueryKeywordObserver is enabled, and that `ActiveRecord::Base.observers.enable :routed_query_keyword_observer` does not return nil.


HalahAb:
Resolves #59

PR #132, opened 2018-08-15
configure rspec to run with random order

HalahAb:
Changes spec_helper.rb config to run rspec tests in random order.


HalahAb:
resolves #59

PR #133, opened 2018-08-15
[#159809630] include language when searching i14y drawers from admin …

MothOnMars:
…center

### DIFF
--- a/app/views/sites/i14y_drawers/show.html.haml
+++ b/app/views/sites/i14y_drawers/show.html.haml
@@ -20,7 +20,7 @@
       %tr
         %td
           %ul.unstyled
-            %li= doc.title
+            %li= translate_bing_highlights(doc.title).html_safe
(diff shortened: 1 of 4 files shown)


app/views/sites/i14y_drawers/show.html.haml
```diff
@@ -20,7 +20,7 @@
       %tr
         %td
           %ul.unstyled
-            %li= doc.title
+            %li= translate_bing_highlights(doc.title).html_safe
```
MothOnMars: @noremmie , the changes on the I14y side allowed highlighting to work in the I14y drawer view in the admin center. This change takes care of formatting the newly highlighted title. We're using the `translate_bing_highlights` for generic highlighted results, so I thought it was okay to use it here. Alternatively, I could rename the method throughout the code to make it more generic - thoughts?

PR #135, opened 2018-08-17
[#156951238] Remove howto.gov and search.digitalgov.gov domain references

noremmie:


noremmie:
@MothOnMars  the process I used for this was to search the code for "search.digitalgov.gov" and "howto" and replace those strings as necessary. There is one remaining "search.digigalgov.gov" reference, but it refers to the GitHub repo and not the domain name. There are a few "howto" references remaining, but they aren't in domain names either. A lot of the changes I made weren't strictly necessary, but I did them anyway for the sake of completeness.

I will be following up with an equivalent commit in the cookbooks repo to update `secrets.yml.dev` over there.

MothOnMars:
My only nit is that this might be a good time to convert the `http` links in the email templates to `https`, i.e. http://search.gov/manual/no-results.html. That obviously ain't a requirement for this story, but I figured I'd call it out.

Otherwise, LGTM 👍

noremmie:
@MothOnMars good call. I just pushed another commit that changes HTTP links to HTTPS for the two email templates that are already being updated in this commit. I chose not to pull in any other email templates to change because that would dramatically increase the scope of acceptance testing for this story.

MothOnMars:
@noremmie , sounds good. I was also looking at those other templates, and they could probably use a separate review in a separate story, such as for old `search.usa.gov` links.

MothOnMars:
In other words: 👍

PR #139, opened 2018-08-19
test fix for ci cache restoring failure with forked PR's

HalahAb:


noremmie:
Thanks, @HalahAb. I've confirmed that this does indeed fix the problem that impacted @MothOnMars's test PR #138. I created a new branch off of your `ci_cache_fix` branch, cherry picked Martha's test commit onto it, and verified that, while all the specs passed, the code coverage test indicated that the new method had no tests.

I'm not familiar with the differences between `save_cache` and `persist_to_workspace` or the differences between `restore_cache` and `attach_to_workspace`. Would you mind explaining how this fix works please?

HalahAb:
@normmie Persist to workspace and save to cache are both ways to persist data in a workflow.  The only difference between saving to cache and persisting to a workspace is how that data is available in a specific workflow, and how it’s available across multiple workflows. 

Saving data to a cache allows any workflow to access that data, if it references that specific cache key. So in a sense, data saved to a cache is parallel in it’s availability across multiple workflow builds. 

Whereas saving data to a workspace, only allows for that data to exist in that specific workflow build and after the workflow is complete the data stored is then discarded. 

I think ci might do a better job at explaining this here, https://circleci.com/blog/persisting-data-in-workflows-when-to-use-caching-artifacts-and-workspaces/

In our case,  
`  - restore_cache:
	key: repo-{{ .Environment.CIRCLE_SHA1 }}
	paths:
		- ~/search-gov `

And 
`- persist_to_workspace:
	root: ~/search-gov
 paths: 
	-
`
Are pretty much doing the same thing, in that they are storing all the data in the `~/search-gov ` working directory and then restoring it. 

Despite it being stored to cache in the first case, the data is not available across multiple builds because the cache key used is the ‘.Environment.CIRCLE_SHA1 ‘ which is unique to each build. So in this case, we can just use a workspace instead of a cache even though they are both effectively doing the same thing. The only reason we needed to change is because of a security change by ci that made storing cache disabled for forked PR’s. 

 The only issue now is the cache restoring for the `gemfile.lock`, still needs to be fixed because despite the tests passing, I noticed that the cache being restored for that is from a much older build, when cache storing was still working. So, I need to fix this, before we can merge.

MothOnMars:
Closing this up. We've been testing & merging PRs from forked repos lately w/ no unexpected failures.

PR #143, opened 2018-08-31
[#160182011] remove Airbrake

MothOnMars:


MothOnMars:
Closing this up for now. We may keep the Airbrake code if we switch to Errbit.

PR #145, opened 2018-09-05
[#160291708] retry SearchgovDomain#check_status

MothOnMars:


Gemfile
```diff
@@ -96,6 +96,7 @@ gem 'public_suffix', '~> 3.0.2'
 gem 'counter_culture', '~> 2.0.0'
 gem 'aasm', '~> 4.12'
 gem 'active_scheduler', '~> 0.5.0'
+gem 'retriable', '~> 3.1'
```
peggles2: Okay super picky but, is there a reason why there is a space here?

MothOnMars: Picky is good :)

The gems that follow the extra line break are the assets-related gems. They had been in a separate `assets` group until the Rails 4 upgrade: https://github.com/GSA/search-gov/commit/bb6016c98ffb9f5b8befcbab1ca947a7a65c3149#diff-8b7db4d5cc4b8f6dc8feb7030baa2478

We removed the group but kept the line break. I think the grouping is helpful, so I can add a comment to clarify.

peggles2:
Looks good!

PR #146, opened 2018-09-12
[#160332453] modify DateRangeTopNFieldQuery to allow queries without an

MothOnMars:
affiliate
- refactor some logstash query code


app/models/logstash_queries/date_range_top_n_field_query.rb
```diff
@@ -1,15 +1,20 @@
 class DateRangeTopNFieldQuery < DateRangeTopNQuery
+  attr_reader :filters
+
   def initialize(affiliate_name, start_date, end_date, filter_field, filter_value, agg_options = {})
     super(affiliate_name, start_date, end_date, agg_options)
-    @filter_field, @filter_value = filter_field, filter_value
+    @filters = { 'affiliate' => @affiliate_name, filter_field => filter_value }.compact
   end
```
peggles2: How about using symbols has hash keys?  { 'name': value, another_name: value } format?

app/models/logstash_queries/top_query_match_query.rb
```diff
@@ -36,7 +35,8 @@ def query(json)
   end
 
   def booleans(json)
-    must_affiliate_date_range(json, @affiliate_name, @start_date, @end_date)
+    must_affiliate(json, @affiliate_name)
+    must_date_range(json, @start_date, @end_date)
     must_not_spider(json)
   end
```
peggles2: def booleans is the same for top_query_match_query.rb and Date_range_top_n_query.rb .  To apply a more DRY approach you may want to pull this method out and place it in as a concern module that both models can share.

MothOnMars: @peggles2 , that's similar to how the `DateRangeFilter` module behaved (removed in this commit), combining a couple of the `AnalyticsDsl` methods. I was torn between the DRY-ness of the `DateRangeFilter` module and the simplicity of the `AnalyticsDsl` methods, which each do a simple, single thing to the JSON. After crawling through the tangle of analytics code, I felt it was simpler to have the analytics methods consolidated in one place , `lib/analytics_dsl.rb`. That said, I can be talked out of that approach if you feel this is a devolution. (That code change was not necessary for my story; I was refactoring for clarity.)

app/models/logstash_queries/date_range_top_n_field_query.rb
MothOnMars: I'm using string keys because they need to be strings when passed to the JSON methods:
```ruby
    filters.each do |field, value|
      json.must do
        json.child! { json.term { json.set! field, value } }
      end
    end
```
I could use symbols, but I'd need to call `.to_s` on them to pass them to the JSON.

peggles2: Usually I prefer using the 1.9 and greater ruby syntax and its also faster. But if you feel strongly about it, you can leave it.

app/models/logstash_queries/top_query_match_query.rb
peggles2: If its not too much refactoring it be better to have one method handle all of them. Is the DateRangeFilter module very complex? If going to take too much time then we can revisit later.

app/models/logstash_queries/date_range_top_n_field_query.rb
peggles2: Actually so does rubocop :) . 
https://github.com/rubocop-hq/ruby-style-guide#symbols-as-keys

app/models/logstash_queries/top_query_match_query.rb
MothOnMars: No, it's not too complex. Mostly I was trying to move away from having the analytics methods scattered through a maze of modules. That said, I think I'll revert my refactorings that were not necessary for this story and cross my fingers that someday we'll be able to do a full refactoring of the analytics code.

app/models/logstash_queries/date_range_top_n_field_query.rb
MothOnMars: Yeah, normally I use symbols as keys, but in this case I'd prefer to keep it consistent with the string keys used throughout the analytics JSON, and avoid the extra `.to_s` call.

MothOnMars:
@peggles2 , I pushed a new `squashme` commit that moves the `must_affiliate_date_range` method from the `DateRangeFilter` module into the standard `AnalyticsDSL` module, which a) consolidates all the analytics methods into one module and b) keeps the code DRY.

PR #159, opened 2018-10-25
Feature/transition to usajobs api

peggles2:
SRCH-98  - Update jobs module to use usajobs api call.
SRCH-120 - Map each search-gov query field for jobs api to usajobs api fields.
SRCH-145 - post-process USAJOBs results


MothOnMars:
@peggles2 , be sure that your build is passing before assigning it for code review. Reassigning this back to you until it's passing.
https://cm.usa.gov/confluence/display/SRCH/How+to+Write+and+Work+on+Stories#HowtoWriteandWorkonStories-State:InProgress

peggles2:
@MothOnMars normally true but since we have a deadline, it be helpful if you could review it in parallel. These may be just me passing the access key in. I'll look into that tonight. Then Friday I can also respond to any code review changes.

spec/models/govbox_set_spec.rb
```diff
@@ -167,71 +167,38 @@
...
 
       context 'when highlighting is enabled by default' do
         it "translates '<em>' and '</em>'" do
           expect(Jobs).to receive(:search).
-            with(query: 'nursing jobs', hl: 1, size: 10, tags: 'federal').
+            with(query: 'nursing jobs', ResultsPerPage: 10).
             and_return job_openings
           govbox_set = GovboxSet.new('nursing jobs', affiliate, nil)
-          expect(govbox_set.jobs.first.position_title).to eq('<strong>Nurse</strong>')
+          expect(govbox_set.jobs.first.position_title).to eq('<em>Nurse</em>')
```
MothOnMars: Same q as above - I would expect this to fail, as we are not expecting anything to be highlighted.

app/models/govbox_set.rb
```diff
@@ -15,8 +15,8 @@ class GovboxSet
               :tweets,
               :video_news_items
 
-  def initialize(query, affiliate, geoip_info, options = {})
```
MothOnMars: `geoip_info` is not a location string. It's an instance of `GeoIP::City`, which is initialized in  the searches controller:
https://github.com/GSA/search-gov/blob/master/app/controllers/searches_controller.rb#L22

`@search = search_klass.new(@search_options.merge(geoip_info: GeoipLookup.lookup(request.remote_ip)))`

example:
```ruby
> g = GeoipLookup.lookup('67.170.115.191')
#<#<Class:0x00007fe27de17708>:GeoIP::City:0x7fe284879650
    area_code = 425,
    city_name = "Everett",
    continent_code = "NA",
    country_code2 = "US",
    country_code3 = "USA",
    country_name = "United States",
    dma_code = 819,
    ip = "67.170.115.191",
    latitude = 47.97900000000001,
    longitude = -122.2021,
    postal_code = "",
    real_region_name = "Washington",
    region_name = "WA",
    request = "67.170.115.191",
    timezone = "America/Los_Angeles"
>
```
So in order to pass a location string to USAJOBs, we'll need to extract it from that object. Something like:

```ruby
def location_name
  geoip_info = GeoipLookup.lookup(request.remote_ip))
  "#{geoip_info.city_name}, #{geoip_info.region_name}"
end
```
Or something like that. We should also verify the format that we need to pass to USAJOBs, such as for searchers in other countries, etc.

lib/jobs.rb
```diff
@@ -26,28 +26,27 @@ module Jobs
     :WC => 'Without Compensation'}.freeze
 
   def self.establish_connection!
-    jobs_api_config = Rails.application.secrets.jobs
-    @endpoint = jobs_api_config['endpoint']
-    @jobs_api_connection = Faraday.new jobs_api_config['host'] do |conn|
+    usajobs_api_config = Rails.application.secrets.jobs
```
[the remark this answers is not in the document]
peggles2: 👍

lib/jobs.rb
```diff
@@ -26,28 +26,27 @@ module Jobs
...
+    @usajobs_api_connection = Faraday.new usajobs_api_config['host'] do |conn|
       conn.request :json
       conn.response :mashify
       conn.response :json
       conn.use :instrumentation
-      conn.adapter jobs_api_config['adapter'] || Faraday.default_adapter
+      conn.adapter usajobs_api_config['adapter'] || Faraday.default_adapter
     end
-    @jobs_api_connection.headers[:accept] = 'application/vnd.usagov.position_openings.v3'
+    @usajobs_api_connection.headers[:accept] = 'application/vnd.usagov.position_openings.v3'
```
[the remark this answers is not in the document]
peggles2: authentication headers are in jobs_api_config = Rails.application.secrets.jobs
which is defined in secrets. I'll double check the necessity of the header

spec/models/govbox_set_spec.rb
```diff
@@ -4,10 +4,10 @@
 describe GovboxSet do
   fixtures :affiliates, :agencies, :federal_register_agencies, :rss_feed_urls, :rss_feeds, :agency_organization_codes
 
-  describe ".new(query, affiliate, geoip_info)" do
+  describe ".new(query, affiliate, location_name)" do
     let(:affiliate) { affiliates(:basic_affiliate) }
     let(:agency) { agencies(:irs) }
-    let(:geoip_info) { double('GeoipInfo', latitude: '12.34', longitude: '-34.56') }
+    let(:location_name) {"Washington, DC"}
```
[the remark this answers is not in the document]
peggles2: 👍

spec/models/govbox_set_spec.rb
```diff
@@ -167,71 +167,38 @@
...
-            with(:query => 'foo', :hl => 1, :size => 10, :organization_ids => 'ABCD,BCDE', :lat_lon => '12.34,-34.56').
+            with(query: 'foo',
+                 ResultsPerPage: 10,
+                 Organization: 'ABCD,BCDE',
+                 LocationName: 'Washington, DC').
             and_return(job_openings)
-          govbox_set = GovboxSet.new('foo', affiliate, geoip_info)
-          expect(govbox_set.jobs.first.position_title).to eq('<strong>Nurse</strong>')
+          govbox_set = GovboxSet.new('foo', affiliate, location_name)
+          expect(govbox_set.jobs.first.position_title).to eq('<em>Nurse</em>')
```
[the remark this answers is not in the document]
peggles2: nope passes.  returns it with default <em>Nurse</em> they must have the default highlighting on.

lib/jobs.rb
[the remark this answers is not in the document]
MothOnMars: AFAIK, we need to pass the `Host`, `User-Agent` and `Authorization-Key` headers:
https://developer.usajobs.gov/Guides/Authentication

spec/models/govbox_set_spec.rb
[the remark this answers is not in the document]
MothOnMars: oh - it looks like it's only passing because we're stubbing the return value of the `Jobs.search`, and the stubbed values include highlighting:
https://github.com/GSA/search-gov/pull/159/files#diff-306a6bc4d0ae572bc309be3facc23fcdL150

We should make sure that the stubbed responses in this spec reflect the JSON format returned by USAJOBs.

MothOnMars:
P.S. Be sure to include the JIRA issue id at the start of the commit message.

config/secrets.yml.dev
```diff
@@ -65,9 +65,11 @@ dev_settings: &DEV_SETTINGS
       writers:
         - localhost
   jobs:
-    host: 'https://jobs.search.gov'
```
MothOnMars: these are indented a couple of extra spaces

lib/jobs.rb
```diff
@@ -26,28 +26,32 @@ module Jobs
     :WC => 'Without Compensation'}.freeze
 
   def self.establish_connection!
-    jobs_api_config = Rails.application.secrets.jobs
-    @endpoint = jobs_api_config['endpoint']
-    @jobs_api_connection = Faraday.new jobs_api_config['host'] do |conn|
+    usajobs_api_config = Rails.application.secrets.jobs
+    @endpoint = usajobs_api_config['endpoint']
+    @usajobs_api_connection = Faraday.new('https://data.usajobs.gov') do |conn|
+      #conn.response :logger, @logger, :bodies => true
```
MothOnMars: remove debug lines

lib/jobs.rb
```diff
@@ -26,28 +26,32 @@ module Jobs
...
+      conn.headers['Authorization-Key'] = 'Qbk5RB/WRc1ctYqwojqlSKeoLVrwokT8OnSLq+G1qu0='
```
MothOnMars: update to use the key & user agent from `usajobs_api_config`

lib/jobs.rb
```diff
@@ -26,28 +26,32 @@ module Jobs
...
+      conn.headers['User-Agent'] = 'parissa.eggleston@gmail.com'
       conn.request :json
-      conn.response :mashify
       conn.response :json
       conn.use :instrumentation
-      conn.adapter jobs_api_config['adapter'] || Faraday.default_adapter
```
MothOnMars: Why this change? There is an `adapter` specified in the usajobs_api_config

lib/jobs.rb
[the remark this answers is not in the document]
MothOnMars: We still need to confirm whether this is needed:
`@usajobs_api_connection.headers[:accept] = 'application/vnd.usagov.position_openings.v3'`

app/models/govbox_set.rb
```diff
@@ -107,29 +107,19 @@ def init_jobs
...
-  def translate_jobs_highlights
-    pre_tag = (@highlighting_options[:pre_tags] || DEFAULT_JOB_HIGHLIGHTING_OPTIONS[:pre_tags]).first
-    post_tag = (@highlighting_options[:post_tags] || DEFAULT_JOB_HIGHLIGHTING_OPTIONS[:post_tags]).first
-    @jobs.each do |job_opening|
-      job_opening.position_title = job_opening.position_title.
-        gsub(/<em>/, pre_tag).
-        gsub(/<\/em>/, post_tag)
-    end
+    #extract location_name from @geoip_info
+    jobs_options.merge!(LocationName: location_name)
```
MothOnMars: #TODO: extract location_name from @geoip_info

app/views/searches/_jobs.mobile.haml
```diff
@@ -13,3 +13,5 @@
       = render partial: 'show_more_show_less'
   .content-block-item.content-block-item-footer
     .see-more= link_to_more_jobs search
+
+= search.jobs["SearchResult"]["SearchResultItems"]
```
MothOnMars: Is this line needed? The view code shouldn't need to change.

peggles2: yikes good catch that is not needed.

lib/jobs.rb
```diff
@@ -13,41 +13,29 @@ module Jobs
...
-    :SY => 'School Year',
-    :WC => 'Without Compensation'}.freeze
-
   def self.establish_connection!
-    jobs_api_config = Rails.application.secrets.jobs
-    @endpoint = jobs_api_config['endpoint']
-    @jobs_api_connection = Faraday.new jobs_api_config['host'] do |conn|
+    usajobs_api_config = Rails.application.secrets.jobs
+    @endpoint = usajobs_api_config['endpoint']
+    @usajobs_api_connection = Faraday.new('https://data.usajobs.gov') do |conn|
```
MothOnMars: use config value instead of hard-coded host

MothOnMars:
LGTM 👍

PR #162, opened 2018-10-25
Remove nutshell columns

MothOnMars:
SRCH-21 migration to remove nutshell_id columns
- database migration to remove nutshell_id columns from the `users` and `affiliates` tables

### DIFF
--- a/app/controllers/admin/affiliates_controller.rb
+++ b/app/controllers/admin/affiliates_controller.rb
@@ -66,7 +66,38 @@ class Admin::AffiliatesController < Admin::AdminController
       config.columns[c].form_ui = :textarea
     end
 
-    update_columns = attribute_columns.reject { |column| column =~ /\A(api_access_key|created_at|external_css_url|favicon_url|has_staged_content|id|nutshell_id|theme|updated_at)\z/i }
+    update_columns = %i[
+                         active
+                         agency
+                         bing_v5_key
+                         dap_enabled
+                         display_name
+                         domain_control_validation_code
+                         fetch_concurrency
+                         force_mobile_format
+                         ga_web_property_id
+                         gets_blended_results
+                         gets_commercial_results_on_blended_search
+                         gets_i14y_results
+                         google_cx
+                         google_key
+                         i14y_date_stamp_enabled
+                         is_bing_image_search_enabled
+                         is_federal_register_document_govbox_enabled
+                         is_medline_govbox_enabled
+                         is_photo_govbox_enabled
+                         is_related_searches_enabled
+                         is_rss_govbox_enabled
+                         is_sayt_enabled
+                         is_video_govbox_enabled
+                         jobs_enabled
+                         locale
+                         name
+                         raw_log_access_enabled
+                         search_consumer_search_enabled
+                         search_engine
+                         website
+                       ]
     config.update.columns = []
     enable_disable_column_regex = /^(is\_|dap_enabled|force_mobile_format|gets_blended_results|gets_commercial_results_on_blended_search|jobs_enabled|raw_log_access_enabled|search_consumer_search_enabled|gets_i14y_results)/.freeze
 
--- a/db/migrate/20181025225740_remove_nutshell_columns.rb
+++ b/db/migrate/20181025225740_remove_nutshell_columns.rb
@@ -1,6 +1,6 @@
 class RemoveNutshellColumns < ActiveRecord::Migration
   def change
-    remove_index :users, :nutshell_id
+    remove_index :users, column: :nutshell_id
     remove_column :users, :nutshell_id, :integer
     remove_column :affiliates, :nutshell_id, :integer
   end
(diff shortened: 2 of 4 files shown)


app/models/affiliate.rb
```diff
@@ -273,8 +273,7 @@ def self.do_not_dup_attributes
         column_name =~ /\A(header_tagline_logo|page_background_image|mobile_logo)/
       end
       %w(api_access_key
-         name
-         nutshell_id).push(*logo_attrs).freeze
+         name).push(*logo_attrs).freeze
     end
```
peggles2: Do you think this is short enough now to be in one line?

db/migrate/20181025225740_remove_nutshell_columns.rb
```diff
@@ -0,0 +1,13 @@
...
+    remove_column :users, :nutshell_id
+    remove_column :affiliates, :nutshell_id
+  end
+
+  def down
+    add_column :affiliates, :nutshell_id, :integer
+    add_column :users, :nutshell_id, :integer
+    add_index :users, :nutshell_id
+  end
+end
```
peggles2: The change method works for removing_index and remove_column.  
https://edgeguides.rubyonrails.org/active_record_migrations.html#using-the-change-method

peggles2: Migrations do not get run automatically correct?  This should not be run till after the code is pushed.

lib/tasks/user.rake
```diff
@@ -6,7 +6,8 @@ namespace :usasearch do
         user.set_approval_status_to_not_approved
         user.save!
         Emailer.user_approval_removed(user).deliver_now
-        NutshellAdapter.new.new_note(user, 'This user is no longer associated with any sites, so their approval status has been set to "not_approved".')
+        note = "User #{user.id}, #{user.email}, is no longer associated with any sites, so their approval status has been set to \"not_approved\"."
+        Rails.logger.info(note)
```
peggles2: tiny nit pick but you can do 'not_approved' instead of \"not_approved\" . looks cleaner.  :)

db/migrate/20181025225740_remove_nutshell_columns.rb
MothOnMars: Ah - my notes on when `remove_column` could be used were from an older version of Rails. Thanks, will update.

MothOnMars:
@peggles2 , this is (finally) ready for review.

app/controllers/admin/affiliates_controller.rb
```diff
@@ -66,7 +66,38 @@ class Admin::AffiliatesController < Admin::AdminController
       config.columns[c].form_ui = :textarea
     end
 
-    update_columns = attribute_columns.reject { |column| column =~ /\A(api_access_key|created_at|external_css_url|favicon_url|has_staged_content|id|nutshell_id|theme|updated_at)\z/i }
+    update_columns = %i[
```
MothOnMars: I slightly refactored this in the process of removing the reference to `nutshell_id`. I realized that this method followed a pattern that had caused problems in the past, where the array of columns was constructed by rejecting specified columns. Similar code caused problems when new columns were added to various tables, and those columns showed up automatically in Super Admin. This change ensures that any columns required for affiliate updates are added intentionally rather than automatically.

peggles2:
Can you add an description at the top about this pull request?

app/controllers/admin/affiliates_controller.rb
```diff
@@ -66,7 +66,38 @@ class Admin::AffiliatesController < Admin::AdminController
...
+                         is_video_govbox_enabled
+                         jobs_enabled
+                         locale
+                         name
+                         raw_log_access_enabled
+                         search_consumer_search_enabled
+                         search_engine
+                         website
+                       ]
     config.update.columns = []
```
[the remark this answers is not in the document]
MothOnMars: Possibly. However, I find it helpful to have them visible in the controller itself, because the various column lists in that controller are added and subtracted from each other. Moving all those lists to a separate initializer could make the code flow more difficult to read. That file could certainly use some cleanup in general, but that should probably be done as a separate story.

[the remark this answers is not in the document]
peggles2: Its best practice to have skinny controllers. Actually now days even skinny models.  https://medium.com/marmolabs/skinny-models-skinny-controllers-fat-services-e04cfe2d6ae. You can look into it as a separate cleanup story if you want. But certainly makes it harder to read with all that clutter.  :).  Why are these columns being added and subtracted?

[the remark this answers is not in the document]
MothOnMars: The addition and subtraction is just how that controller was originally written. The various column sets (used for displaying the various affiliate settings in super admin) are put together with logic such as "foo_colums = bar_columns + baz_columns", or "foo columns = bar_columns - baz_columns". Before doing any heavy refactoring on that controller, we'd probably want to audit the affiliates table in general to determine which columns can be completely dropped (as many of those support legacy features).

MothOnMars:
@peggles2 , can you give https://github.com/GSA/search-gov/pull/162/commits/633d7c092f4e4af4aadc0695df0557d754fb5050 a quick review? I realized when I went to re-test this on demo2 that the syntax for the index removal was incorrect (it worked for migrating, but not rolling back).

PR #164, opened 2018-11-01
SRCH-156 stub www.agency.gov/robots.txt response

MothOnMars:
- warn if entire site is disallowed

### DIFF
--- a/app/models/searchgov_crawler.rb
+++ b/app/models/searchgov_crawler.rb
@@ -21,7 +21,7 @@ class SearchgovCrawler
   end
 
   def crawl
-    Rails.logger.warn(disallowed_warning) unless @robotex.allowed? base_url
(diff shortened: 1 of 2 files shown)


app/models/searchgov_crawler.rb
```diff
@@ -21,6 +21,8 @@ def initialize(domain:, skip_query_strings: true, delay: nil, srsly: false)
   end
 
   def crawl
+    Rails.logger.warn(disallowed_warning) unless @robotex.allowed? base_url
+
```
peggles2: What if for some reason @robotex is null?

MothOnMars: `@robotex` is initialized as part of the crawler's `initialize` method. It is never nil.

PR #166, opened 2018-11-03
SRCH-165 increase re-record interval for Bing specs

MothOnMars:


MothOnMars:
@peggles2 , master and all PRs are currently failing CircleCI. This is because we have set the Bing specs to re-record their VCR cassettes every 4 months, and the new recordings are causing the BingV6 specs to fail. Because we are about to replace Bing V6, I'm resolving this by bumping up the re-record interval from 4 to 6 months, so that we can continue using the existing recordings for another 2 months.

PR #170, opened 2018-11-08
SRCH-140 - Cleanup Neogov jobs- related code

peggles2:
SRCH-140 - Remove all references to Neogov
- the code is free of references to "Neogov" jobs
- the site admin display page doesn't mention Neogov jobs as a source

### DIFF
--- a/app/views/sites/displays/_govboxes.html.haml
+++ b/app/views/sites/displays/_govboxes.html.haml
@@ -23,7 +23,7 @@
     - if @site.agency.present?
       %tr
         %td Job openings
-        %td.cell-1x USAJobs/NeoGov
+        %td.cell-1x USAJOBS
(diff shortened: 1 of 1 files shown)


app/views/sites/displays/_govboxes.html.haml
```diff
@@ -23,7 +23,7 @@
     - if @site.agency.present?
       %tr
         %td Job openings
-        %td.cell-1x USAJobs/NeoGov
```
MothOnMars: @peggles2 , per our conversation, this should be the only change in this commit. https://github.com/GSA/search-gov/pull/159 contains all the rest of the NeoGov code cleanup.

Also, let's change `USAJobs` to `USAJOBS`, as that is the correct capitalization.

PR #171, opened 2018-11-08
SRCH-197 de-dupe sitemap urls

MothOnMars:


MothOnMars:
@peggles2 , this is a tweak to fix a production bug that was blocking the www.osha.gov URLs from being indexed.

PR #175, opened 2018-11-14
Feature/bulk reindex urls

peggles2:
SRCH-183 - Update the SearchgovUrl#fetch method to set that boolean to false
SRCH-181 - Update the SearchgovURL.fetch_required scope to look for records where that boolean is true

### DIFF
--- a/app/models/searchgov_url.rb
+++ b/app/models/searchgov_url.rb
@@ -15,7 +15,8 @@ class SearchgovUrl < ActiveRecord::Base
 
   attr_accessible :last_crawl_status,
                   :last_crawled_at,
-                  :url, :lastmod,
+                  :url,
+                  :lastmod,
                   :enqueued_for_reindex
   attr_reader :response, :document, :tempfile
   attr_readonly :url
@@ -38,9 +39,9 @@ class SearchgovUrl < ActiveRecord::Base
     column_names: { ['searchgov_urls.last_crawled_at IS NULL'] => 'unfetched_urls_count' }
 
   scope :fetch_required, ->() do
-     where('last_crawled_at IS NULL OR
-            lastmod > last_crawled_at OR
-            enqueued_for_reindex')
+    where('last_crawled_at IS NULL
+           OR lastmod > last_crawled_at
+           OR enqueued_for_reindex')
   end
 
   class SearchgovUrlError < StandardError; end
--- a/app/models/searchgov_url.rb
+++ b/app/models/searchgov_url.rb
@@ -39,7 +39,7 @@ class SearchgovUrl < ActiveRecord::Base
     column_name: proc {|url| !url.fetched? ? 'unfetched_urls_count' : nil },
     column_names: { ['searchgov_urls.last_crawled_at IS NULL'] => 'unfetched_urls_count' }
 
-  scope :fetch_required, ->() do
+  scope :fetch_required, -> do
     where('last_crawled_at IS NULL
            OR lastmod > last_crawled_at
            OR enqueued_for_reindex')
@@ -50,7 +50,7 @@ class SearchgovUrl < ActiveRecord::Base
 
   def fetch
     raise DomainError.new("#{searchgov_domain.domain}: #{searchgov_domain.status}") if !searchgov_domain.available?
-    self.update(last_crawled_at: Time.now, enqueued_for_reindex: false)
+    update(last_crawled_at: Time.now, enqueued_for_reindex: false)
     self.load_time = Benchmark.realtime do
       DocumentFetchLogger.new(url, 'searchgov_url').log
       begin
--- a/app/models/searchgov_url.rb
+++ b/app/models/searchgov_url.rb
@@ -13,7 +13,7 @@ class SearchgovUrl < ActiveRecord::Base
                                 application/vnd.openxmlformats-officedocument.spreadsheetml.sheet
                               )
 
-  attr_accessible :last_crawl_status, :last_crawled_at, :url, :lastmod
+  attr_accessible :last_crawl_status, :last_crawled_at, :url, :lastmod, :enqueued_for_reindex
   attr_reader :response, :document, :tempfile
   attr_readonly :url
 
@@ -34,14 +34,14 @@ class SearchgovUrl < ActiveRecord::Base
     column_name: proc {|url| !url.fetched? ? 'unfetched_urls_count' : nil },
     column_names: { ['searchgov_urls.last_crawled_at IS NULL'] => 'unfetched_urls_count' }
 
-  scope :fetch_required, -> { where('last_crawled_at IS NULL OR lastmod > last_crawled_at') }
+  scope :fetch_required, -> { where('last_crawled_at IS NULL OR lastmod > last_crawled_at OR enqueued_for_reindex') }
 
   class SearchgovUrlError < StandardError; end
   class DomainError < StandardError; end
 
   def fetch
     raise DomainError.new("#{searchgov_domain.domain}: #{searchgov_domain.status}") if !searchgov_domain.available?
-    self.update_attributes(last_crawled_at: Time.now)
+    self.update(last_crawled_at: Time.now, enqueued_for_reindex: false)
     self.load_time = Benchmark.realtime do
       DocumentFetchLogger.new(url, 'searchgov_url').log
       begin
--- a/app/models/searchgov_url.rb
+++ b/app/models/searchgov_url.rb
@@ -37,7 +37,11 @@ class SearchgovUrl < ActiveRecord::Base
     column_name: proc {|url| !url.fetched? ? 'unfetched_urls_count' : nil },
     column_names: { ['searchgov_urls.last_crawled_at IS NULL'] => 'unfetched_urls_count' }
 
-  scope :fetch_required, -> { where('last_crawled_at IS NULL OR lastmod > last_crawled_at OR enqueued_for_reindex') }
+  scope :fetch_required, ->() do
+     where('last_crawled_at IS NULL OR
+            lastmod > last_crawled_at OR
+            enqueued_for_reindex')
+  end
 
   class SearchgovUrlError < StandardError; end
   class DomainError < StandardError; end
--- a/app/models/searchgov_url.rb
+++ b/app/models/searchgov_url.rb
@@ -4,14 +4,15 @@ class SearchgovUrl < ActiveRecord::Base
   include ActionView::Helpers::NumberHelper
 
   MAX_DOC_SIZE = 15.megabytes
-  SUPPORTED_CONTENT_TYPES = %w(
+  SUPPORTED_CONTENT_TYPES = %w[
                                 text/html
+                                text/plain
                                 application/msword
                                 application/pdf
                                 application/vnd.ms-excel
                                 application/vnd.openxmlformats-officedocument.wordprocessingml.document
                                 application/vnd.openxmlformats-officedocument.spreadsheetml.sheet
-                              )
+                              ]
 
   attr_accessible :last_crawl_status,
                   :last_crawled_at,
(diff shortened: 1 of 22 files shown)


MothOnMars:
@peggles2 , I thought you were going to make all the re-index changes on one branch? This PR duplicates the db/structure.sql changes made in https://github.com/GSA/search-gov/pull/172, but lacks the specs for the column addition. I think it would be best to wait for https://github.com/GSA/search-gov/pull/172 to be merged to master, and then open a PR with only the additive changes so that we don't have two PRs open with the same db change.

MothOnMars:
Adding a Do Not Merge label to this, as it's dependent on the migration in https://github.com/GSA/search-gov/pull/172

app/models/searchgov_url.rb
```diff
@@ -34,14 +34,14 @@ class SearchgovUrl < ActiveRecord::Base
     column_name: proc {|url| !url.fetched? ? 'unfetched_urls_count' : nil },
     column_names: { ['searchgov_urls.last_crawled_at IS NULL'] => 'unfetched_urls_count' }
 
-  scope :fetch_required, -> { where('last_crawled_at IS NULL OR lastmod > last_crawled_at') }
+  scope :fetch_required, -> { where('last_crawled_at IS NULL OR lastmod > last_crawled_at OR enqueued_for_reindex') }
```
MothOnMars: Let's wrap this long line. See [the style guide](https://github.com/rubocop-hq/ruby-style-guide#lambda-multi-line) for multi-line lambda styling.

db/structure.sql
```diff
@@ -1,8 +1,8 @@
--- MySQL dump 10.13  Distrib 5.6.35, for osx10.11 (x86_64)
```
MothOnMars: This change should not be part of this commit. Once https://github.com/GSA/search-gov/pull/172 is merged, this PR can be rebased against master so that it doesn't include the DB change.

spec/models/searchgov_url_spec.rb
```diff
@@ -34,6 +33,12 @@
         expect(SearchgovUrl.fetch_required.pluck(:url)).
           to eq %w[http://www.agency.gov/new http://www.agency.gov/outdated]
       end
+
+      it 'includes urls that have been enqueued for reindexing' do
```
MothOnMars: @peggles2 , for consistency with this scope spec, I'd create an "enqueued for reindex" URL in the `before` block:
```ruby
      before do
        SearchgovUrl.create!(url: 'http://www.agency.gov/new')
        SearchgovUrl.create!(
          url: 'http://www.agency.gov/outdated', last_crawled_at: 1.week.ago, lastmod: 1.day.ago
        )
        SearchgovUrl.create!(
          url: 'http://www.agency.gov/current', last_crawled_at: 1.day.ago, lastmod: 1.week.ago
        )
        SearchgovUrl.create!(
          url: 'http://www.agency.gov/enqueued', last_crawled_at: 1.day.ago, enqueued_for_reindex: true
        )
      end
```
And then update the expectation:
```ruby
      it 'includes new & outdated URLs, & URLs that have been enqueued for reindex' do
        expect(SearchgovUrl.fetch_required.pluck(:url)).
          to eq %w[http://www.agency.gov/new
                          http://www.agency.gov/outdated
                          http://www.agency.gov/enqueued]
      end
```

spec/models/searchgov_url_spec.rb
```diff
@@ -125,6 +130,17 @@
         fetch
       end
 
+      context 'when the record is enqueued for reindex' do
+        let(:searchgov_url) do
+          SearchgovUrl.create!(valid_attributes.merge(enqueued_for_reindex: true))
+        end
+
+        it('sets enqueued_for_reindex to false') do
```
MothOnMars: @peggles2 , we don't include parentheses in `it` blocks for stylistic consistency:
```ruby
it 'sets enqueued_for_reindex to false'  do
```

db/structure.sql
peggles2: reindex changes can not be done on one branch. We need to first deploy the migrations. Do the migration on production have it ready then deploy the second part.  When the migration pr goes to production, I will merge the master branch and you will not see the those changes on this branch anymore.

spec/models/searchgov_url_spec.rb
peggles2: I thought about that but thats not behavior we are looking to see. We are not creating a new URL. We are updating an existing url to be re-indexed.  New URLS are going to be indexed anyway.  Which will give you a false positive when you fetch it.  Plus it will be confusing to the developer that reads this test. because your checking links that were created. I like the way its testing this way because its clear its a url that was updated with enqueued for reindexing and now its showing up to be reindexed when the fetch method is called.

spec/models/searchgov_url_spec.rb
peggles2: doup! good catch! :) 👍

spec/models/searchgov_url_spec.rb
MothOnMars: I see your point. However, I think we should avoid making extra db calls in specs. This spec has to create and then update the record, which is not what the spec needs to test. 

Also, one problem with this spec is that it doesn't necessarily test that the URL is included in the `fetch_required` scope. The `searchgov_url` record in the spec does not have a `last_crawled_at` or `last_crawl_status` value, so it would be included in the scope anyway; this spec would not have failed before the code was written to make it pass. 

Let's chat about this spec today; I have an idea about refactoring.

peggles2: Yeah we can chat. I think update or create your hitting the db.  Tests are suppose to follow the same behavior a user does. The user will never create a url with enqueued for reindex set to true. That will never happen. A user will update an existing url. So the test needs to reproduce that behavior.  The one your suggesting does not do that.

peggles2: Oh and also I am using an existing url that has already been created.  I'm just updating it.

spec/models/searchgov_url_spec.rb
```diff
@@ -34,6 +33,12 @@
         expect(SearchgovUrl.fetch_required.pluck(:url)).
           to eq %w[http://www.agency.gov/new http://www.agency.gov/outdated]
       end
+
+      it 'includes urls that have been enqueued for reindexing' do
+        searchgov_url.update(enqueued_for_reindex: true)
+        expect(SearchgovUrl.fetch_required.pluck(:url)).
+            to include "http://www.agency.gov/boring.html"
```
MothOnMars: whoops, I missed this - indent 2 instead of 4 spaces

spec/models/searchgov_url_spec.rb
```diff
@@ -125,6 +130,17 @@
         fetch
       end
 
+      context 'when the record is enqueued for reindex' do
+        let(:searchgov_url) do
+          SearchgovUrl.create!(valid_attributes.merge(enqueued_for_reindex: true))
+        end
+
+        it 'sets enqueued_for_reindex to false' do
+          expect{ fetch }.to change{ searchgov_url.enqueued_for_reindex }.
```
MothOnMars: indent 2 instead of 4 spaces

spec/models/searchgov_url_spec.rb
MothOnMars: @peggles2 , following up on our chat on Friday: 
If using fixtures will be more trouble than it's worth at this point, I would proceed as I suggested originally. I wouldn't worry that a record won't normally be created in that state, as this spec is not testing any user behavior. Throughout the specs, there are records created with various attributes for testing, which is fine for unit tests.

MothOnMars:
@peggles2 , there are some conflicts in the VCR cassettes. I recommend checking out those cassettes from master, and then re-recording them to update them:
```
$ git checkout master spec/vcr_cassettes/search/gov
$ rspec spec/lib/tasks/searchgov_spec.rb
```

app/models/searchgov_url.rb
```diff
@@ -13,7 +13,10 @@ class SearchgovUrl < ActiveRecord::Base
                                 application/vnd.openxmlformats-officedocument.spreadsheetml.sheet
                               )
 
-  attr_accessible :last_crawl_status, :last_crawled_at, :url, :lastmod
+  attr_accessible :last_crawl_status,
+                  :last_crawled_at,
+                  :url, :lastmod,
```
MothOnMars: @peggles2, as long as you're breaking these into multiple lines, it would be good to separate `:url` and `:lastmod` into separate lines for readability.

app/models/searchgov_url.rb
MothOnMars: Please see the [style guide](https://github.com/rubocop-hq/ruby-style-guide#lambda-multi-line) for multi-line lambda styling.

app/models/searchgov_url.rb
```diff
@@ -34,14 +37,18 @@ class SearchgovUrl < ActiveRecord::Base
     column_name: proc {|url| !url.fetched? ? 'unfetched_urls_count' : nil },
     column_names: { ['searchgov_urls.last_crawled_at IS NULL'] => 'unfetched_urls_count' }
 
-  scope :fetch_required, -> { where('last_crawled_at IS NULL OR lastmod > last_crawled_at') }
+  scope :fetch_required, ->() do
+    where('last_crawled_at IS NULL OR
```
MothOnMars: minor nit, but this would read a bit better if the conjunctions preceded the conditions:
``` ruby
where('last_crawled_at IS NULL
       OR lastmod > last_crawled_at
       OR enqueued_for_reindex')
```

app/models/searchgov_url.rb
peggles2: ack! missed that thanks.

spec/models/sitemap_indexer_spec.rb
```diff
@@ -143,7 +143,11 @@
 
       it 'ignores them' do
         index
-        expect(SearchgovUrl.pluck(:url)).to eq ['http://agency.gov/doc1']
+        expect(SearchgovUrl.pluck(:url)).to include 'http://agency.gov/doc1'
+      end
+
+      it 'does not include urls that are from different domain' do
```
MothOnMars: What is the reason for this addition? This spec doesn't actually test anything.

app/models/searchgov_url.rb
MothOnMars: @peggles2 , the style guide recommends:
```ruby
l = lambda do |a, b|
  tmp = a * 7
  tmp * b / 50
end
```

spec/models/searchgov_url_spec.rb
MothOnMars: bump

spec/models/searchgov_url_spec.rb
MothOnMars: bump

spec/support/fetchable_behavior.rb
```diff
@@ -30,34 +30,52 @@
...
       describe '.fetched' do
         it 'includes successfully and unsuccessfully fetched records' do
           expect(described_class.fetched.pluck(:url)).
-            to match_array %w[http://agency.gov/ok http://agency.gov/failed]
+            to include('http://agency.gov/ok', 'http://agency.gov/failed')
+        end
+
+        it 'does not include unfetched records' do
+          expect(described_class.fetched.pluck(:url)).
+              not_to include 'http://agency.gov/unfetched'
```
MothOnMars: indent 2 instead of 4 spaces

spec/support/fetchable_behavior.rb
```diff
@@ -30,34 +30,52 @@
...
         end
       end
 
       describe '.unfetched' do
         it 'includes unfetched records' do
-          expect(described_class.unfetched.pluck(:url)).to eq ['http://agency.gov/unfetched']
+          expect(described_class.unfetched.pluck(:url)).
+              to include 'http://agency.gov/unfetched'
```
MothOnMars: indent 2 instead of 4 spaces

MothOnMars:
Almost there - just a few stylistic & indendation changes to make

spec/models/searchgov_url_spec.rb
peggles2: Not sure where you see 4 spaces.

MothOnMars: Weird - I must have commented on the wrong line. I do still still 4 instead of 2 spaces in a number of places throughout the `fetchable_behavior` spec.

app/models/searchgov_url.rb
MothOnMars: @peggles2 ^^

PR #187, opened 2018-11-16
SRCH-202 - scrub job keywords from query

peggles2:
SRCH - 202 - scrub job keywords from query
- No migrations needed 
- To pass search terms with out the job keywords that triggers the API call to usajobs.

### DIFF
--- a/app/models/govbox_set.rb
+++ b/app/models/govbox_set.rb
@@ -104,10 +104,13 @@ class GovboxSet
 
   def init_jobs
     if @affiliate.jobs_enabled?
-      job_results = Jobs.search(Keyword:        @query,
-                                Organization:   @affiliate.agency&.joined_organization_codes,
-                                LocationName:   @geoip_info&.location_name,
-                                ResultsPerPage: 10)&.search_result&.search_result_items
+      job_results = Jobs.search({
+        query: @query,
+        organization_codes: @affiliate.agency&.joined_organization_codes,
+        location_name: @geoip_info&.location_name,
+        results_per_page: 10
+      })&.search_result&.search_result_items
+
       if job_results.present?
         @jobs = JobResultsPostProcessor.new(results: job_results)&.post_processed_results
         @modules << 'JOBS'
--- a/app/models/govbox_set.rb
+++ b/app/models/govbox_set.rb
@@ -104,20 +104,15 @@ class GovboxSet
 
   def init_jobs
     if @affiliate.jobs_enabled?
-      job_results = Jobs.search(build_jobs_search_options)&.search_result&.search_result_items
+      job_results = Jobs.search(Keyword:        @query,
+                                Organization:   @affiliate.agency&.joined_organization_codes,
+                                LocationName:   @geoip_info&.location_name,
+                                ResultsPerPage: 10)&.search_result&.search_result_items
       if job_results.present?
         @jobs = JobResultsPostProcessor.new(results: job_results)&.post_processed_results
         @modules << 'JOBS'
       end
     end
-  end
-
-  def build_jobs_search_options
-    jobs_options = { Keyword: @query, ResultsPerPage: 10 }
-    org_hash = { Organization: @affiliate.agency&.joined_organization_codes }
-    jobs_options.merge!(org_hash)
-    jobs_options.merge!(LocationName: @geoip_info&.location_name)
-    jobs_options.compact
   end
 
   def init_federal_register_documents
--- a/lib/jobs.rb
+++ b/lib/jobs.rb
@@ -27,11 +27,11 @@ module Jobs
     end
   end
 
-  def self.scrub_keyword(keyword)
-    if keyword =~ /^#{JOB_RELATED_KEYWORDS}$/
-      keyword =~ /^(position|opening|posting|job|employment|)s?$/ ? "" : keyword
+  def self.scrub_query(query)
+    if query =~ /^#{JOB_RELATED_KEYWORDS}$/
+      query =~ /^(position|opening|posting|job|employment|)s?$/ ? "" : query
     else
-      keyword.remove(/\b#{JOB_RELATED_KEYWORDS}\b/, '').squish
+      query.remove(/\b#{JOB_RELATED_KEYWORDS}\b/, '').squish
     end
   end
 
@@ -49,11 +49,10 @@ module Jobs
   end
 
   def self.params(options)
-    { Keyword:        scrub_keyword(options[:query]),
+    { Keyword:        scrub_query(options[:query]),
       Organization:   options[:organization_codes],
       LocationName:   options[:location_name],
-      ResultsPerPage: options[:results_per_page]
-    }
+      ResultsPerPage: options[:results_per_page] }
   end
 
 end
--- a/lib/jobs.rb
+++ b/lib/jobs.rb
@@ -29,7 +29,7 @@ module Jobs
 
   def self.scrub_keyword(keyword)
     if keyword =~ /^#{JOB_RELATED_KEYWORDS}$/
-      keyword !~ /^(position|opening|posting|job|employment|)s?$/ ? keyword : ""
+      keyword =~ /^(position|opening|posting|job|employment|)s?$/ ? "" : keyword
     else
       keyword.remove(/\b#{JOB_RELATED_KEYWORDS}\b/, '').squish
     end
@@ -50,7 +50,7 @@ module Jobs
 
   def self.params(options)
     { Keyword:        scrub_keyword(options[:query]),
-      Organization:   options[:organization],
+      Organization:   options[:organization_codes],
       LocationName:   options[:location_name],
       ResultsPerPage: options[:results_per_page]
     }
--- a/lib/jobs.rb
+++ b/lib/jobs.rb
@@ -1,6 +1,6 @@
 module Jobs
   SIMPLE_SEARCHES = '(job|employment|internship)s?'
-  JOB_RELATED_KEYWORDS = '((position|opening|posting|job|employment|intern(ship)?|seasonal|trabajo|puesto|empleo|vacante)s?|(opportunit|vacanc)(y|ies))|(posicion|ocupacion|oportunidad|federal)(es\b|\b)|gobierno'
+  JOB_RELATED_KEYWORDS = '((position|opening|posting|job|employment|intern(ship)?|seasonal|trabajo|puesto|empleo|vacante)s?|(opportunit|vacanc)(y|ies))|(posicion|ocupacion|oportunidad|federal)(es)?|gobierno'
   SCRUB_KEYWORDS = JOB_RELATED_KEYWORDS.remove(/\|intern\(ship\)|\|seasonal|\|federal\||gobierno/)
   SIMPLE_SINGULARS = %w{
     statistic number level rate description trend growth projection survey forecast figure report verification record
@@ -29,7 +29,7 @@ module Jobs
   end
 
   def self.scrub_query(query)
-    query.remove(/\b#{SCRUB_KEYWORDS}\b/i, '').squish
+    query.remove(/\b#{SCRUB_KEYWORDS}\b/i).squish
   end
 
   def self.search(job_options)
--- a/lib/jobs.rb
+++ b/lib/jobs.rb
@@ -28,13 +28,14 @@ module Jobs
   end
 
   def self.job_scrub(keyword)
-    keyword =~ /^#{JOB_RELATED_KEYWORDS}$/ ? keyword  : keyword.gsub(/#{JOB_RELATED_KEYWORDS}/,"").strip
+    scrubbed_keyword = keyword.gsub(/#{JOB_RELATED_KEYWORDS}/, '').strip
+    keyword =~ /^#{JOB_RELATED_KEYWORDS}$/ ? keyword : scrubbed_keyword
   end
 
-  def self.search(options)
-    keyword = options[:Keyword]
-    options[:Keyword] = job_scrub keyword
-    @usajobs_api_connection.get(@endpoint, options).body if query_eligible?(keyword)
+  def self.search(job_options)
+    if query_eligible?(job_options[:query])
+      @usajobs_api_connection.get(@endpoint, params(job_options)).body
+    end
   rescue => error
     Rails.logger.error("Trouble fetching jobs information: #{error}")
     nil
@@ -42,6 +43,14 @@ module Jobs
 
   def self.query_eligible?(query)
     query =~ /\b#{JOB_RELATED_KEYWORDS}\b/i && !(query =~ /\b#{BLOCKED_KEYWORDS}\b/i) && !(query =~ /["():]|^-| -\S+/)
+  end
+
+  def self.params(options)
+    { Keyword:        job_scrub(options[:query]),
+      Organization:   options[:organization],
+      LocationName:   options[:location_name],
+      ResultsPerPage: options[:results_per_page]
+    }
   end
 
 end
--- a/lib/jobs.rb
+++ b/lib/jobs.rb
@@ -51,8 +51,7 @@ module Jobs
     { Keyword:        scrub_keyword(options[:query]),
       Organization:   options[:organization],
       LocationName:   options[:location_name],
-      ResultsPerPage: options[:results_per_page]
-    }
+      ResultsPerPage: options[:results_per_page] }
   end
 
 end
--- a/lib/jobs.rb
+++ b/lib/jobs.rb
@@ -28,7 +28,7 @@ module Jobs
   end
 
   def self.scrub_query(query)
-    query.gsub(/(position|opening|posting|job|employment)s?/, '').squish
+    query.remove(/\b(position|opening|posting|job|employment)s?\b/, '').squish
   end
 
   def self.search(job_options)
--- a/lib/jobs.rb
+++ b/lib/jobs.rb
@@ -27,8 +27,10 @@ module Jobs
     end
   end
 
-  def self.job_scrub(keyword)
-    scrubbed_keyword = keyword.gsub(/#{JOB_RELATED_KEYWORDS}/, '').strip
+  def self.scrub_keyword(keyword)
+    # Keep the job related keyword if its the only word being searched,
+    # so "" is not passed to api.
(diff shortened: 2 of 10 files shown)


app/models/govbox_set.rb
```diff
@@ -104,22 +104,17 @@ def init_news_items
...
   end
 
-  def build_jobs_search_options
-    jobs_options = { Keyword: @query, ResultsPerPage: 10 }
-    org_hash = { Organization: @affiliate.agency&.joined_organization_codes }
-    jobs_options.merge!(org_hash)
-    jobs_options.merge!(LocationName: @geoip_info&.location_name)
-    jobs_options.compact
-  end
-
```
peggles2: build_job_search_options was initially determine the correct value for fields for example the geoip_info based on longitude and latitude.  Now that we are using usajobs this method is not really needed. So I removed it and passed the options directly.

lib/jobs.rb
```diff
@@ -27,8 +27,15 @@ def self.establish_connection!
     end
   end
 
-  def self.search(options)
-    @usajobs_api_connection.get(@endpoint, options).body if query_eligible?(options[:Keyword])
+  def self.job_scrub(keyword)
```
MothOnMars: The naming of this method is a little confusing to me. Its purpose is to scrub keywords from the original query, so maybe we could call it something like `scrub_keywords(query)`

lib/jobs.rb
```diff
@@ -27,8 +27,15 @@ def self.establish_connection!
...
+    scrubbed_keyword = keyword.gsub(/#{JOB_RELATED_KEYWORDS}/, '').strip
```
MothOnMars: I think you want to use `squish` instead of `strip` here, to remove extra inner spaces too.

MothOnMars: Also, you could use `remove` here instead of `gsub` (either way is fine w/ me, although I find `remove` reads a little better):
```ruby
keyword.remove(/#{JOB_RELATED_KEYWORDS}/).squish
```

MothOnMars: Your regex needs word boundaries so that we don't partially strip words. Examples:
```ruby
> "grand reopening".gsub(/#{JOB_RELATED_KEYWORDS}/, '')
"grand re"

> "seasonality".gsub(/#{JOB_RELATED_KEYWORDS}/, '')
"ity"
```

```

File: lib/jobs.rb [hunk 4]
```diff
@@ -27,8 +27,15 @@ def self.establish_connection!
...
+    keyword =~ /^#{JOB_RELATED_KEYWORDS}$/ ? keyword : scrubbed_keyword
```
MothOnMars: @peggles2 , can you explain a scenario where this 2nd line of code is necessary? It seems to me that this method could be reduced to simply: 
```ruby
keyword.gsub(/#{JOB_RELATED_KEYWORDS}/, '').squish
```

spec/lib/jobs_spec.rb
```diff
@@ -2,7 +2,12 @@
 
 describe Jobs do
   describe '.search(options)' do
-    subject(:search) { Jobs.search(Keyword: 'jobs') }
+    subject(:search) do
+      Jobs.search({ query:'jobs',
+                    organization_code: '',
+                    location_name: 'Washington, DC, United States',
+                    results_per_page: 10})
+    end
```
MothOnMars: @peggles2 , this would be a good time to add a spec to check that we're searching with the right params (which you can build on for the Radius story we discussed). Something like:
```ruby
it 'searches USAJOBS with the correct params' do
  search
  expect(a_request(:get, usajobs_url).with( Keyword: 'jobs'...etc. )).to have_been_made
end
```
Check out the documentation on using the Webmock gem for this: 
https://github.com/bblimke/webmock#setting-expectations-in-rspec-with-a_request

You might have to play with the expectation a bit; let me know if you want to pair.

spec/lib/jobs_spec.rb
```diff
@@ -21,8 +26,16 @@
     end
   end
 
+  describe 'job_scrub(query)' do
```
MothOnMars: The Rspec standard is to preface class methods in a `describe` block with a dot, or a hash for instance methods:
```ruby
describe '.job_scrub(query)' do
```
https://github.com/rubocop-hq/rspec-style-guide#example-group-naming

lib/jobs.rb
peggles2: If they search for just the word 'job' . we don't want to remove it. Then they would search for everything. So in that case we just search job.  But if we search for Nursing job, then we want to pass Nursing through.

spec/lib/jobs_spec.rb
peggles2: We can add that to jobscrub story.  It be out of scope for this project since its not adding any additional params. Would you like to place that in the Jiraticket?

lib/jobs.rb
peggles2: I added a comment there in case the next developer runs into the same question.

MothOnMars: I think it's preferable to strip those keywords regardless, because the postings don't always include those keywords. For example, if you search without any keywords, there are currently 7437 results. If you search with `Keyword=job`, that number drops to 3427.

peggles2: I think that the word "jobs" still returns results and you are now changing the behavior off the application.

peggles2: and its not just jobs internship and seasonal they search and get everything vs getting their results narrowed to catching the word internship

MothOnMars: if `Keyword=jobs` returns only 10% of the available jobs results, that's a problem. That could even reduce the results to zero depending on the location. If we want to scope by "internship" or "seasonal", the best way to do that would be with a followup story where queries with those terms passed the associated `PositionOfferingTypeCode` param (which we decided not to do until we see the need).

spec/lib/jobs_spec.rb
MothOnMars: It's in scope for this story, because you're changing how and where we're building the params that we're passing to USAJOBs. For example, if you're calling `Jobs.search(query: 'foo')`, we need to verify that we are passing `Keyword: 'foo'` to USAJOBs.

lib/jobs.rb
peggles2: Right but for now if they don't get enough results with jobs they have the see more jobs link. We can change this when we do the followup for internship.

MothOnMars: The purpose of this story is to ensure that we are returning jobs results as frequently as possible. Approximately a third of production job queries include "job" or "jobs", so we would be working against the purpose of this story if we continue to severely limit the cases where a jobs search returns results. With the code as it stands, we are supporting a very infrequent use case ("seasonal" or internship" are in < %5 of job queries) at the expense of the most frequent use case.

spec/lib/jobs_spec.rb
peggles2: This will take some time to do. Rather not hold jobscrub over this. Lets code review the rest and we can look at this one together. Also this get request requires to pass authentication etc.

lib/jobs.rb
```diff
@@ -38,6 +48,14 @@ def self.query_eligible?(query)
     query =~ /\b#{JOB_RELATED_KEYWORDS}\b/i && !(query =~ /\b#{BLOCKED_KEYWORDS}\b/i) && !(query =~ /["():]|^-| -\S+/)
   end
 
+  def self.params(options)
+    { Keyword:        scrub_keyword(options[:query]),
+      Organization:   options[:organization_codes],
+      LocationName:   options[:location_name],
+      ResultsPerPage: options[:results_per_page]
+    }
```
MothOnMars: please re-fix the bracket alignment per https://github.com/GSA/search-gov/pull/187#discussion_r235543473

spec/lib/jobs_spec.rb
```diff
@@ -2,12 +2,30 @@
...
   describe '.search(options)' do
-    subject(:search) { Jobs.search(Keyword: 'jobs') }
+    subject(:search) do
+      Jobs.search({ query:'jobs',
+                    organization_codes: 'HE38',
+                    location_name: 'Washington, DC, USA',
+                    results_per_page: 10})
+    end
+
+    let(:usajobs_url) {'https://data.usajobs.gov/api/search'}
```
MothOnMars: content within curly brackets should be set off by spaces: `{ 'foo' }`

lib/jobs.rb
MothOnMars: @peggles2 , the regex still needs work:
```ruby
> Jobs.scrub_keyword('grand re')
"grand re"
```

We should also have test cases for these examples.

MothOnMars: These are also not working properly:
```ruby
>Jobs.scrub_keyword('seasonal jobs')
"seasonal jobs"

> Jobs.scrub_keyword('job posting')
"job posting"
```

spec/lib/jobs_spec.rb
```diff
@@ -2,12 +2,30 @@
 
 describe Jobs do
   describe '.search(options)' do
-    subject(:search) { Jobs.search(Keyword: 'jobs') }
+    subject(:search) do
+      Jobs.search({ query:'jobs',
+                    organization_codes: 'HE38',
+                    location_name: 'Washington, DC, USA',
+                    results_per_page: 10})
```
MothOnMars: Leave a space within curly brackets:
```ruby
 results_per_page: 10 }
```

spec/lib/jobs_spec.rb
```diff
@@ -2,12 +2,30 @@
...
     end
 
+    it 'searches USAJOBS with the correct params' do
+      search
+      expect(a_request(:get, usajobs_url).with(
+        query: {
+          Keyword:        '',
+          Organization:   'HE38',
+          LocationName:   'Washington, DC, USA',
+          ResultsPerPage: 10}
```
MothOnMars: Fix curly bracket alignment, either:
```ruby
        query: {
          Keyword:        '',
          Organization:   'HE38',
          LocationName:   'Washington, DC, USA',
          ResultsPerPage: 10
        }
```

or:
```ruby
        query: { Keyword:        '',
                 Organization:   'HE38',
                 LocationName:   'Washington, DC, USA',
                 ResultsPerPage: 10 }
```

spec/lib/jobs_spec.rb
```diff
@@ -21,8 +39,24 @@
     end
   end
 
+  describe '.scrub_keyword(query)' do
+    context 'when the search phrase contains a job related keyword' do
+      it 'returns the query with out generic job keywords' do
```
MothOnMars: typo: "without"

lib/jobs.rb
```diff
@@ -27,8 +27,18 @@ def self.establish_connection!
     end
   end
 
-  def self.search(options)
-    @usajobs_api_connection.get(@endpoint, options).body if query_eligible?(options[:Keyword])
+  def self.scrub_keyword(keyword)
```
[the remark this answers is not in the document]
peggles2: after having a conversation we agreed on scrub_query(query)

lib/jobs.rb
peggles2: weird thought I fixed that.

spec/lib/jobs_spec.rb
```diff
@@ -21,8 +39,24 @@
...
+        expect(Jobs.scrub_keyword('Nursing jobs')).to eq('Nursing')
+      end
+
+      it 'return blank if its equal to one of these job term keywords job,employment,posting,position' do
```
MothOnMars: typos: should be "returns blank", and `it's` instead of `its`

This spec lists the keywords that should be scrubbed, but doesn't test them. I'd reword the spec description to make it more generic, and test the keywords:

```
it 'returns blank when the query is a generic job keyword' do
  %w[job employment posting position].each do |query|
      expect(Jobs.scrub_keyword(query)).to eq('')
  end
end


```

spec/lib/jobs_spec.rb
peggles2: After our discussion added the test. I made a change to it expecting 10 results. There are currently 3 nursing jobs at NIH. Also can't put a abcd for organization because usajobs will return nothing. I changed it to HE38 which is NIH.

lib/jobs.rb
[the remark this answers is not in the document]
MothOnMars: 👍

spec/lib/jobs_spec.rb
peggles2: good catch

spec/lib/jobs_spec.rb
MothOnMars: bump

peggles2: this is out dated. Already fixed see how its flagged 'Outdated'

MothOnMars: I'm seeing this issue in the current version of this file:
https://github.com/GSA/search-gov/blob/b8c3c1f5e6d3539d19a12070b751b15c950762af/spec/lib/jobs_spec.rb#L19

peggles2: Look at it in the files changed tab.

spec/lib/jobs_spec.rb
```diff
@@ -2,10 +2,26 @@
...
+    it 'searches USAJOBS with the correct params' do
+      search
+      expect(a_request(:get, usajobs_url).with(
+        query: {
+          Keyword:        'Nursing',
+          Organization:   'HE38',
+          LocationName:   'Washington, DC, USA',
+          ResultsPerPage: 10 }
+      )).to have_been_made
     end
```
peggles2: @MothOnMars please review files changed tab for most recent changes.

spec/lib/jobs_spec.rb
MothOnMars: That's where I'm seeing it:
https://github.com/GSA/search-gov/pull/187/files#diff-42eb9f61e8fd848a9ea45dc8ae155816R23

You fixed the spacing issue, but not the bracket alignment.

peggles2: oh wait I think I see what you mean. I thought the issue was with the bracket.

lib/jobs.rb
MothOnMars: @peggles2 , I should have thought of this earlier, but I ran your scrubber against the list of queries from prod. Github didn't like the file, so I've attached it to the JIRA issue: https://cm-jira.usa.gov/secure/attachment/68014/scrubbed_job_queries.csv

The list shows the original query and the scrubbed version. It's looking pretty good, but there are some cases that jump out that we should cover (and add test cases for):
- the scrubber should be case-insensitive, so that it will scrub out "Jobs", etc.
- we should scrub out the generic Spanish terms, such as "empleo", "trabajo", etc.
- we should scrub out `(opportunit|vacanc)(y|ies)`

With those tweaks, this should be good to go.

peggles2: If we do that we should just scrub all the equivalent spanish words that matches our generic job terms. which then leaves us basically scrubbing everything but internship, seasonal, federal, and gobierno.  I ran some tests and found a bug in the current code.

  JOB_RELATED_KEYWORDS = '((position|opening|posting|job|employment|intern(ship)?|seasonal|trabajo|puesto|empleo|vacante)s?|(opportunit|vacanc)(y|ies))|(posicion|ocupacion|oportunidad|federal)es|gobierno'

Place that in Rubular and you will see that posicion, ocupacion, oportunidad and federal will not match unless you add "es".  I believe thats to make it plural?

The correct way is
  JOB_RELATED_KEYWORDS = '((position|opening|posting|job|employment|intern(ship)?|seasonal|trabajo|puesto|empleo|vacante)s?|(opportunit|vacanc)(y|ies))|(posicion|ocupacion|oportunidad|federal)(es\b|\b)|gobierno'

spec/lib/jobs_spec.rb
```diff
@@ -63,6 +64,11 @@
       it 'does include the job related keyword if it is part of another word' do
         expect(Jobs.scrub_query('grand reopening')).to eq('grand reopening')
       end
+
+      it 'is case sensitive when scrubing queries' do
+        expect(Jobs.scrub_query('JoB')).to eq('')
+      end
+
```
MothOnMars: remove extra line break

spec/lib/jobs_spec.rb
```diff
@@ -63,6 +64,11 @@
       it 'does include the job related keyword if it is part of another word' do
         expect(Jobs.scrub_query('grand reopening')).to eq('grand reopening')
       end
+
+      it 'is case sensitive when scrubing queries' do
```
MothOnMars: typo: should be "scrubbing"

lib/jobs.rb
```diff
@@ -28,7 +29,7 @@ def self.establish_connection!
   end
 
   def self.scrub_query(query)
-    query.remove(/\b(position|opening|posting|job|employment)s?\b/, '').squish
+    query.remove(/\b#{SCRUB_KEYWORDS}\b/i, '').squish
```
MothOnMars: I like this better - much cleaner.

You only need to pass the regex to `remove` (sorry I didn't catch that last time):
```ruby
query.remove(/\b#{SCRUB_KEYWORDS}\b/i).squish
```

lib/jobs.rb
```diff
@@ -1,6 +1,7 @@
 module Jobs
   SIMPLE_SEARCHES = '(job|employment|internship)s?'
-  JOB_RELATED_KEYWORDS = '((position|opening|posting|job|employment|intern(ship)?|seasonal|trabajo|puesto|empleo|vacante)s?|(opportunit|vacanc)(y|ies))|(posicion|ocupacion|oportunidad|federal)es|gobierno'
+  JOB_RELATED_KEYWORDS = '((position|opening|posting|job|employment|intern(ship)?|seasonal|trabajo|puesto|empleo|vacante)s?|(opportunit|vacanc)(y|ies))|(posicion|ocupacion|oportunidad|federal)(es\b|\b)|gobierno'
```
MothOnMars: Instead of:
`(posicion|ocupacion|oportunidad|federal)(es\b|\b)`

you should just need:
`(posicion|ocupacion|oportunidad|federal)(es)?`

```ruby
[17] Usasearch(main)> /(posicion|ocupacion|oportunidad|federal)(es)?/ === 'posicion'
true
[18] Usasearch(main)> /(posicion|ocupacion|oportunidad|federal)(es)?/ === 'posiciones'
true
```

MothOnMars:
👍 LGTM

PR #195, opened 2018-12-04
Feature/jobs radius

peggles2:
- SRCH-215
Expand the radius in requests to USAJOBS to 75 miles.


lib/jobs.rb
```diff
@@ -13,6 +13,7 @@ module Jobs
     "|#{Date.current.year.to_s}"+
     "|#{BLOCKED_PHRASES}"+
     "|(#{SIMPLE_SINGULARS.join('|')})s?"
+  SEARCH_RADIUS = 75
```
MothOnMars: @peggles2 , you've set the radius as a constant, but you're not passing it to USAJOBS.

spec/lib/jobs_spec.rb
```diff
@@ -34,6 +34,19 @@
         expect(search).to be_nil
       end
     end
+
+    context 'when the search is within the default radius, but in a different city' do
```
MothOnMars: @peggles2 , this doesn't necessarily test the radius, because this spec would pass if there were any jobs available in Baltimore. I would expect to see the `it 'searches USAJOBS with the correct params` spec updated with the expected radius.

MothOnMars:
@peggles2 , there are a few RSpec failures to be addressed

lib/jobs.rb
peggles2: Yes this is due to master getting merged. Not ready for code review yet. Thanks! :)

peggles2:
Ready for review

spec/lib/jobs_spec.rb
```diff
@@ -5,7 +5,7 @@
     subject(:search) do
       Jobs.search({ query:'Nursing jobs',
                     organization_codes: 'HE38',
-                    location_name: 'Washington, DC, USA',
+                    location_name: 'Baltimore, MD, USA',
```
MothOnMars: Why the change from DC to Baltimore? This doesn't affect the test.

peggles2: This tests NIH search results to come back with results even if we are in Baltimore.

PR #197, opened 2018-12-05
update secrets.yml.dev to contain where to find wiki to obtain and ad…

peggles2:
SRCH-111 - Update search.gov secrets.yml.dev to reference where to obtain jobs api key


MothOnMars:
LGTM 👍

PR #211, opened 2018-12-20
SRCH-286 retain unknown HTML tags

MothOnMars:
This change allows us to parse the content from custom HTML tags, which we had previously been stripping out.

### DIFF
--- a/app/models/html_document.rb
+++ b/app/models/html_document.rb
@@ -41,8 +41,19 @@ class HtmlDocument < WebDocument
     # it to html, then running it through Loofah again, it's because Loofah's 'to_text' method is
     # much smarter about whitespace than Nokogiri, but you can't run `to_text` on the Nokogiri
     # elements extracted by "html.at('main')". Hence the jumping through hoops.
-    plain_text = Loofah.fragment(main_html).scrub!(:whitewash).to_text(encode_special_chars: false)
+    plain_text = Loofah.fragment(main_html).
+      scrub!(tag_scrubber).
+      scrub!(:whitewash).
+      to_text(encode_special_chars: false)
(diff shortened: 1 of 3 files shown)


app/models/html_document.rb
```diff
@@ -41,10 +41,21 @@ def parse_content
     # it to html, then running it through Loofah again, it's because Loofah's 'to_text' method is
     # much smarter about whitespace than Nokogiri, but you can't run `to_text` on the Nokogiri
     # elements extracted by "html.at('main')". Hence the jumping through hoops.
-    plain_text = Loofah.fragment(main_html).scrub!(:whitewash).to_text(encode_special_chars: false)
+    plain_text = Loofah.fragment(main_html).
+      scrub!(tag_scrubber).
+      scrub!(:whitewash).
+      to_text(encode_special_chars: false)
     plain_text.gsub(/[ \t]+/,' ' ).gsub(/[\n\r]+/, "\n").chomp.lstrip
   end
```
peggles2: what if plain_text is nil?

app/models/html_document.rb
```diff
@@ -41,10 +41,21 @@ def parse_content
...
 
+  def tag_scrubber
+    # convert custom tags to plain 'ol divs to ensure they are not
+    # stripped out during HTML sanitization
+    Loofah::Scrubber.new do |node|
+      node.name = 'div' if /-/ === node.name
+    end
```
peggles2: Is the === necessary here? https://github.com/rubocop-hq/ruby-style-guide#no-case-equality

app/models/html_document.rb
MothOnMars: It won't ever be nil. Even if `main_html` is nil, `to_text` will convert that to an empty string:
```ruby
Loofah.fragment(nil).to_text
#=> ""
```

PR #212, opened 2018-12-29
SRCH-207 omit nav & footer elements when parsing HTML

MothOnMars:
This change prevents us from indexing content that is likely to be the same on the majority of pages on a website.

### DIFF
--- a/app/models/html_document.rb
+++ b/app/models/html_document.rb
@@ -52,7 +52,7 @@ class HtmlDocument < WebDocument
     Loofah::Scrubber.new do |node|
       # convert custom tags to plain 'ol divs to ensure they are not
       # stripped out during HTML sanitization
-      node.name = 'div' if /-/.match(node.name)
+      node.name = 'div' if node.name =~ /-/
 
       # omit common elements
       node.remove if %w[footer nav].include?(node.name)
--- a/app/models/html_document.rb
+++ b/app/models/html_document.rb
@@ -49,10 +49,13 @@ class HtmlDocument < WebDocument
   end
 
   def tag_scrubber
-    # convert custom tags to plain 'ol divs to ensure they are not
-    # stripped out during HTML sanitization
     Loofah::Scrubber.new do |node|
+      # convert custom tags to plain 'ol divs to ensure they are not
+      # stripped out during HTML sanitization
       node.name = 'div' if /-/.match(node.name)
+
+      # omit common elements
+      node.remove if %w[footer nav].include?(node.name)
     end
   end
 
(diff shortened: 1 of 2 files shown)


app/models/html_document.rb
```diff
@@ -49,10 +49,13 @@ def parse_content
...
-    # convert custom tags to plain 'ol divs to ensure they are not
-    # stripped out during HTML sanitization
     Loofah::Scrubber.new do |node|
+      # convert custom tags to plain 'ol divs to ensure they are not
+      # stripped out during HTML sanitization
       node.name = 'div' if /-/.match(node.name)
+
+      # omit common elements
+      node.remove if %w[footer nav].include?(node.name)
     end
```
peggles2: Wouldn't it be easier to read to do  node.name =~ /footer|nav/ ? And you are not looping through an array.

app/models/html_document.rb
```diff
@@ -49,10 +49,13 @@ def parse_content
   end
 
   def tag_scrubber
-    # convert custom tags to plain 'ol divs to ensure they are not
-    # stripped out during HTML sanitization
     Loofah::Scrubber.new do |node|
+      # convert custom tags to plain 'ol divs to ensure they are not
+      # stripped out during HTML sanitization
       node.name = 'div' if /-/.match(node.name)
+
```
peggles2: Its more performant if you write this node.name =~ /-/.  https://rubocop.readthedocs.io/en/latest/cops_performance/#performanceredundantmatch

app/models/html_document.rb
MothOnMars: I find it easier to read as-is, and I try to avoid using regexes unless it's necessary. In this case, it's not necessary, because we're looking for an exact string match. The performance difference between the two approaches is negligible:
```ruby
[17] Usasearch(main)> Benchmark.realtime{ 1000.times{  %w[nav footer].include? '<nav>' } }
0.005774999997811392
[18] Usasearch(main)> Benchmark.realtime{ 1000.times{  '<nav>' =~ /nav|footer/ } }
0.004207999998470768
```

app/models/html_document.rb
MothOnMars: Will update.

MothOnMars: Note: Once we are on ruby 2.4+, the preferred method will be `.match?`:
https://rubocop.readthedocs.io/en/latest/cops_performance/#performanceregexpmatch

peggles2: Yes in 2.4  its because the methods avoid creating a MatchData object or saving backref.  It also recommends using .match? instead of .match. But right now we don't have match? so the more performant way to do it is node.name =~ /-/.

PR #216, opened 2019-01-03
job trigger words that are in a search word should not trigger. Also add career as a job trigger keyword.

peggles2:
"career" should be a keyword that triggers a job search
"international" should NOT trigger a job search
"international affairs jobs" should trigger a job search 
"intern" should trigger a job search


peggles2:
@MothOnMars  can you review one more time? I forgot to add careers as a trigger word.

lib/jobs.rb
```diff
@@ -1,18 +1,28 @@
...
-  SIMPLE_SINGULARS = %w{
-    statistic number level rate description trend growth projection survey forecast figure report verification record
-    authorization card classification form hazard poster fair board outlook grant funding factor other cut
+  SIMPLE_SEARCHES = '(job|employment|internship)s?'.freeze
+  JOB_RELATED_KEYWORDS = '((position|opening|posting|job|employment|career|' \
+                         'intern(ship)?|seasonal|trabajo|puesto|empleo|' \
+                         'carrera|vacante+?)(s\b|\b)|(opportunit|vacanc)' \
+                         '(y|ies))|(posicion|ocupacion|oportunidad|federal)'\
+                         '(es)?|gobierno'.
+                         freeze
```
MothOnMars: @peggles2 , it would be cleaner to use `# frozen_string_literal: true` at the top of the file than to use `freeze` on all the strings.

lib/jobs.rb
```diff
@@ -1,18 +1,28 @@
...
+  SCRUB_KEYWORDS = JOB_RELATED_KEYWORDS.
+                   remove(/\|intern\(ship\)|\|seasonal|\|federal\||gobierno/).
+                   freeze
```
MothOnMars: If you're wrapping a chained method to another line, be sure to indent the following line:
https://github.com/rubocop-hq/ruby-style-guide#consistent-multi-line-chains

peggles2: ah looks like we have to update rubocop.yml.  I was following that. RuboCop: Layout/MultilineMethodCallIndentation: Align `freeze` with `'((position|opening|posting|job|employment|career|' \` on line 3.

MothOnMars: 👍

MothOnMars:
LGTM 👍

peggles2:
@MothOnMars review one last time :)

MothOnMars:
LGTM, pending specs passing

PR #217, opened 2019-01-03
SRCH-311 stop re-recording VCR cassettes

MothOnMars:
Prevent VCR cassettes from being re-recorded at set intervals.


MothOnMars:
Note: you'll see that the cassettes in `spec/vcr_cassettes/ApiBingDocsSearch` & ``spec/vcr_cassettes/ApiBingSearch` have been deleted, and new ones have been added in `spec/vcr_cassettes/api_bing_docs_search` & `spec/vcr_cassettes/api_bing_search`. For some reason, removing the re-record interval from those specs changed the cassette directory naming from camel case to snake case. I am not sure why that would be, but they are now named consistently with our other cassette directories, and there should be no change in functionality.

spec/spec_helper.rb
```diff
@@ -111,12 +111,6 @@
       full_context = description.chomp(test).strip
       name = full_context.split(/\s+/, 2).join('/').underscore.gsub(/\./,'/').gsub(/[^\w\/]+/, '_').gsub(/\/$/, '')
 
-      #disable re-recording Azure cassettes until tests are removed:
```
MothOnMars: This is no longer necessary, as the `re_record_interval` for all specs is now `nil`.

PR #220, opened 2019-01-09
SRCH-294 cache I14Y results

MothOnMars:
Cache I14y responses to speed up I14y searches and reduce the number of queries we send to I14y.


lib/i14y_collections.rb
```diff
@@ -22,7 +26,7 @@ def self.get(handle)
   end
 
   def self.search(params)
-    response = i14y_connection.get "#{API_ENDPOINT}/search", params
-    response.body
+    response = cached_connection.get "#{API_ENDPOINT}/search", params
```
peggles2: I think we decided to do parenthesis around method calls with arguments.

peggles2: While your in this file can you update line 2 to be API_ENDPOINT = "/api/v1/collections".freeze?  :)

MothOnMars:
@peggles2 , this is ready for re-review. I made some additional changes per rubocop's recommendations.

lib/i14y_collections.rb
MothOnMars: Good suggestions - updated.

Note: I froze strings in those files using the `# frozen_string_literal: true` magic comment, as that will be the default in Ruby 3.

MothOnMars:
Closing this in favor of https://github.com/GSA/search-gov/pull/222.

PR #221, opened 2019-01-10
SRCH-305 - Set admin session timeout to 1 hour

peggles2:
- set admin interface to time out after 1 hour of inactivity
- cleanup code

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -1,4 +1,5 @@
 # frozen_string_literal: true
+
 class User < ActiveRecord::Base
   APPROVAL_STATUSES = %w[pending_email_verification
                          pending_approval approved
@@ -16,7 +17,7 @@ class User < ActiveRecord::Base
   validates :password,
             format: { with: PASSWORD_FORMAT,
                       if: :require_password?,
-                      message: 'must include a combination of letters, '\
+                      message: 'must include a combination of letters, ' \
                                'numbers, and special characters.' }
   validate :confirm_current_password,
            on: :update,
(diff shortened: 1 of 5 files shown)


MothOnMars:
@peggles2 , please make sure your branch is passing before assigning it for code review.

peggles2:
Okay ready for review. Some rubocop warnings for code that was previously there we may need to address later.

features/user_sessions.feature
```diff
@@ -66,9 +66,22 @@ Feature: User sessions
       | contact_name | email            | password  | password_updated_at | approval_status |
       | Jane         | jane@example.com | test1234! | 2017-01-01          | approved        |
     And a clear email queue
-    And the date is 2017-01-02
+    And the time is 2017-03-30 23:55
     And I log in with email "jane@example.com" and password "test1234!"
-    And the date becomes 2017-04-15
```
MothOnMars: Let's use `And the time becomes` rather than `And the time is` in this line. Functionally they are equivalent, but `And the time becomes` is a better indicator of the passage of time that we want to capture in this feature.

features/user_sessions.feature
```diff
@@ -66,9 +66,22 @@ Feature: User sessions
...
+    And the time is 2017-03-31 00:20
     And I follow "Add Site"
+    And the time is 2017-03-31 01:10
```
MothOnMars: This line is not necessary for this test.

app/models/user.rb
```diff
@@ -1,30 +1,42 @@
+# frozen_string_literal: true
+
 class User < ActiveRecord::Base
-  APPROVAL_STATUSES = %w( pending_email_verification pending_approval approved not_approved )
+  APPROVAL_STATUSES = %w[pending_email_verification
+                         pending_approval approved
```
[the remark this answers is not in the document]
peggles2: I think its readable and doesn't push the code down with a long list.

features/user_sessions.feature
peggles2: Its to show that the password now expired. So the login was done at 03/30 password expires 03/31 because its 90 days since the password was set.

app/models/user.rb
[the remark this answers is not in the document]
MothOnMars: ok

features/user_sessions.feature
peggles2: I missed it was already there. Gotcha thanks

MothOnMars:
LGTM 👍

PR #227, opened 2019-01-18
SRCH-300 - JOBS whitelist

peggles2:
Add to the list whitelist of words that triggers the jobs trigger words when it should not. Such as 
Job statement should not trigger USAJOBS while nursing jobs should.

Looking through the production log files the following words have been added:
application creating agreement certificate changes denial
earnings income participation change loss offer
qualifications requirement statement


peggles2:
After conversation with Martha we agreed that the findings were not significant enough to whitelist.
Closing this PR

PR #229, opened 2019-01-24
SRCH-352 - Upgrade rake gem for search-gov

peggles2:
In search-gov the rake gem was downgraded to version 10 because: Rake 11.0.1 removes the last_comment method which Rails 2.3 rspec-core (< 3.4.4) uses. This issue has been resolved and now the latest version of rake supports last_comment message. 
see: https://stackoverflow.com/questions/35893584/nomethoderror-undefined-method-last-comment-after-upgrading-to-rake-11

"UPDATE: This has now been fixed in rspec, so the only thing necessary should be updating rspec."

For Rails 5 we need the latest version of rake so its necessary to upgrade rake.


Gemfile
```diff
@@ -2,7 +2,7 @@ source 'https://rubygems.org'
 
 # Temporarily limiting rake version:
 # #http://stackoverflow.com/questions/35893584/nomethoderror-undefined-method-last-comment-after-upgrading-to-rake-11
-gem 'rake', '~> 10.0'
```
MothOnMars: Let's continue to use pessimistic version constraints unless there's a reason to lock a specific version (and if we're locking the version, we should be sure to leave a comment listing the reason).

Gemfile
```diff
@@ -2,7 +2,7 @@ source 'https://rubygems.org'
 
 # Temporarily limiting rake version:
```
MothOnMars: This comment can also be removed.

peggles2: oh! good catch! forgot to remove that.

MothOnMars:
LGTM 👍

PR #233, opened 2019-01-30
SRCH-376 - passwords and password confirmations

peggles2:
Update passwords log

### DIFF
--- a/config/initializers/filter_parameter_logging.rb
+++ b/config/initializers/filter_parameter_logging.rb
@@ -1,4 +1,4 @@
 # Be sure to restart your server when you modify this file.
 
 # Configure sensitive parameters which will be filtered from the log file.
-Rails.application.config.filter_parameters += [:password, :password_confirmation]
+Rails.application.config.filter_parameters += [:password]
--- a/spec/controllers/user_sessions_controller_spec.rb
+++ b/spec/controllers/user_sessions_controller_spec.rb
@@ -12,7 +12,7 @@ describe UserSessionsController do
 
   describe "#create" do
     let(:post_create) do
-      post :create, user_session: { email: user.email , password: user.password }
+      post :create, user_session: { email: user.email, password: user.password }
     end
 
     it 'filters passwords in the logfile' do
--- a/spec/controllers/user_sessions_controller_spec.rb
+++ b/spec/controllers/user_sessions_controller_spec.rb
@@ -15,12 +15,15 @@ describe UserSessionsController do
       post :create, user_session: { email: user.email, password: user.password }
     end
 
-    it 'filters passwords in the logfile' do
-      allow(Rails.logger).to receive(:info)
-      expect(Rails.logger).to receive(:info).
-        with(/ \"password\"=>\"\[FILTERED\]\"/)
-      post :create, user_session: { email: users('developer').email,
-                                    password: 'test1234!'}
+    context 'when the user is approved' do
+      let(:user) { users(:affiliate_manager) }
+
+      it 'filters passwords in the logfile' do
+        allow(Rails.logger).to receive(:info)
+        expect(Rails.logger).to receive(:info).
+          with(/ \"password\"=>\"\[FILTERED\]\"/)
+        post_create
+      end
     end
 
     context 'when the user is not approved' do
--- a/spec/controllers/user_sessions_controller_spec.rb
+++ b/spec/controllers/user_sessions_controller_spec.rb
@@ -15,6 +15,14 @@ describe UserSessionsController do
       post :create, user_session: { email: user.email , password: user.password }
     end
 
+    it 'filters passwords in the logfile' do
+      allow(Rails.logger).to receive(:info)
+      expect(Rails.logger).to receive(:info).
+        with(/ \"password\"=>\"\[FILTERED\]\"/)
+      post :create, user_session: { email: users('developer').email,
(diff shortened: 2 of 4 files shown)


MothOnMars:
P.S. There's a typo in the commit message

config/initializers/filter_parameter_logging.rb
```diff
@@ -1,4 +1,4 @@
 # Be sure to restart your server when you modify this file.
 
 # Configure sensitive parameters which will be filtered from the log file.
-Rails.application.config.filter_parameters += [:password, :password_confirmation]
+Rails.application.config.filter_parameters += [:password]
```
peggles2: took password_confirmation off its not needed as rails handles the regex.

spec/controllers/users_controller_spec.rb
```diff
@@ -89,22 +89,41 @@
...
+        'email': 'changed@foo.com' }
     end
+
     context 'when logged in as affiliate' do
       before { activate_authlogic }
       include_context 'approved user logged in'
 
-      it { is_expected.to permit(*permitted_params).for(:update, params: { user: update_params }) }
+      context 'when changing the password' do
+        let(:update_password_params) do
```
MothOnMars: It would be better to set this variable as `let(:update_params)`. That way you can use `let(:update_user)` to run the update:
```ruby
let(:update_user) do
  post: update, id: current_user.id, user: update_params
end 
```
Then you can DRY up the various `post :update`s with `update_user`, as we discussed.

spec/controllers/user_sessions_controller_spec.rb
```diff
@@ -15,6 +15,14 @@
       post :create, user_session: { email: user.email , password: user.password }
     end
 
+    it 'filters passwords in the logfile' do
+      allow(Rails.logger).to receive(:info)
+      expect(Rails.logger).to receive(:info).
+        with(/ \"password\"=>\"\[FILTERED\]\"/)
+      post :create, user_session: { email: users('developer').email,
```
MothOnMars: You can replace this with `post_create`.

spec/controllers/users_controller_spec.rb
```diff
@@ -89,22 +89,41 @@
...
+               user: update_password_params
+        end
+      end
 
       context 'when the User#update_attributes was successfully' do
         before do
-          expect(current_user).to receive(:update_attributes).with(update_params).
-            and_return(true)
+          expect(current_user).
+            to receive(:update_attributes).
```
MothOnMars: When chaining method calls on multiple lines, you can use one level of indentation for all the lines:
```ruby
foo.bar.
  baz.
  biz
```
Also, there are a few indentations in this file with 4 spaces that should be 2.

peggles2: I was following what rubocop advised. It wanted to indent it that way. RuboCop: Layout/MultilineMethodCallIndentation: Indent `with` 2 spaces more than `receive(:update_attributes)` on line 120.

peggles2: We discussed updating rubucop rules for this.

spec/controllers/user_sessions_controller_spec.rb
```diff
@@ -12,7 +12,18 @@
 
   describe "#create" do
     let(:post_create) do
-      post :create, user_session: { email: user.email , password: user.password }
+      post :create, user_session: { email: user.email, password: user.password }
+    end
+
+    context 'when the user is approved' do
+      let(:user) { users(:affiliate_manager) }
```
MothOnMars: You don't need a separate context for this, because this should apply to all users. You can just move the  `let(:user)` line to below `describe "#create` so that the `user` variable will be available in all the `create` specs.:
```ruby
  describe "#create" do
    let(:user) { users(:affiliate_manager) }
    let(:post_create) do
      post :create, user_session: { email: user.email, password: user.password }
    end

    it 'filters passwords in the logfile' do
      allow(Rails.logger).to receive(:info)
      expect(Rails.logger).to receive(:info).
        with(/ \"password\"=>\"\[FILTERED\]\"/)
      post_create
    end
```

spec/controllers/users_controller_spec.rb
```diff
@@ -88,23 +88,41 @@
...
-      { 'contact_name' => 'BAR', 'email' => 'changed@foo.com' }
+      { 'contact_name': 'BAR',
+        'email': 'changed@foo.com' }
     end
+
     context 'when logged in as affiliate' do
       before { activate_authlogic }
       include_context 'approved user logged in'
 
-      it { is_expected.to permit(*permitted_params).for(:update, params: { user: update_params }) }
```
MothOnMars: Please un-delete this spec :)

spec/controllers/users_controller_spec.rb
```diff
@@ -88,23 +88,42 @@
...
-      { 'contact_name' => 'BAR', 'email' => 'changed@foo.com' }
+      { 'contact_name': 'BAR',
+        'email': 'changed@foo.com' }
     end
+
     context 'when logged in as affiliate' do
       before { activate_authlogic }
       include_context 'approved user logged in'
 
       it { is_expected.to permit(*permitted_params).for(:update, params: { user: update_params }) }
```
MothOnMars: Please add a line break between this spec & the context below.

MothOnMars:
LGTM 👍

PR #242, opened 2019-02-19
Feature/strong params

peggles2:
remove the protected_attributes gem and update code to use strong params.


spec/controllers/complete_registration_controller_spec.rb
```diff
@@ -43,7 +43,8 @@
       end
 
       it "@user should be assigned" do
-        post :update, :id => "known"
+        post :update, { id: "known",
```
MothOnMars: Per our Slack conversation, let's bump the line limit to 85 characters. The we can unwrap lines like this.

spec/lib/tasks/bulk_import_spec.rb
```diff
@@ -18,11 +18,12 @@
         expect(@rake[task_name].prerequisites).to include("environment")
       end
 
-      context "when a file and default user email is specified" do
+      context 'when a file and default user email is specified' do
         before do
           @user = users(:affiliate_manager)
           Affiliate.where("name LIKE ?", "test%").each{|aff| aff.destroy }
-          @existing_affiliate = Affiliate.create({:name => 'test1', :display_name => 'Test 1'}, :as => :test)
+          @existing_affiliate = Affiliate.
```
MothOnMars: unwrap line

spec/models/affiliate_spec.rb
```diff
@@ -261,15 +263,17 @@
       expect(Affiliate.find(affiliate.id).staged_footer).to eq('staged footer')
     end
 
-    it "should populate search labels for English site" do
-      english_affiliate = Affiliate.create!(valid_attributes.merge(:locale => 'en'), :as => :test)
+    it 'should populate search labels for English site' do
+      english_affiliate = Affiliate.
```
MothOnMars: unwrap line

spec/models/affiliate_spec.rb
```diff
@@ -261,15 +263,17 @@
...
+        create!(valid_attributes.merge(locale: 'en'))
       english_affiliate.default_search_label = ''
       english_affiliate.save!
       expect(english_affiliate.default_search_label).to eq('Everything')
     end
 
     it "should populate search labels for Spanish site" do
-      spanish_affiliate = Affiliate.create!(valid_attributes.merge(:locale => 'es'), :as => :test)
+      spanish_affiliate = Affiliate.
```
MothOnMars: unwrap line

spec/controllers/complete_registration_controller_spec.rb
peggles2: Had to bump to 90.  85 was not enough.

MothOnMars:
@peggles2 , CodeClimate is now reporting "962 issues to fix". I'm sure that includes lines that are longer than 90 char, so I'm going to force-approve it.

config/environments/development.rb
```diff
@@ -37,9 +37,6 @@
     # Expands the lines which load the assets
     config.assets.debug = true
   end
-
-  # Raise exception on mass assignment protection for Active Record models
-  config.active_record.mass_assignment_sanitizer = :strict
```
MothOnMars: @peggles2 , let's retain the behavior of raising an error in test/dev environments (which I find more useful than the default of logging the params):
`config.action_controller.action_on_unpermitted_parameters = :raise`

config/environments/test.rb
```diff
@@ -38,9 +38,6 @@
   config.static_cache_control = "public, max-age=3600"
 
   config.active_job.queue_adapter = :test
-
-  # Raise exception on mass assignment protection for Active Record models
-  config.active_record.mass_assignment_sanitizer = :strict
```
MothOnMars: same comment re `config.action_controller.action_on_unpermitted_parameters = :raise` in `config/environments/development.rb`

config/environments/development.rb
MothOnMars: `:log` is the default in dev/test environments. I'm only suggesting we change it in dev/test environments, because it makes bug hunting faster. Raising an error was our setting until we upgraded to Rails 4 (I didn't realize the behavior had changed until I tested this story).

MothOnMars: Also, if we set it to `:raise` and our tests break, then we should address those tests and/or code.

config/environments/test.rb
peggles2: There are 143 errors in rspec. This can be due to legacy code or old rspec tests that pass everything. It should probably be a separate ticket to do that.

MothOnMars:
Per my discussion with @peggles2 , we are going to stick with logging unpermitted params in dev/test environments. There over 100 test failures if we raise an error. SRCH-454 has been filed to address those failures at a later date.

PR #245, opened 2019-02-21
Release 1.13.0

lsamuels-fearless:
[SRCH-387] - make BingV7 default search engine
[SRCH-430] - add healthchecker to ensure cloudfront can access assets


MickinSahni:
Adding approval here as well

PR #247, opened 2019-02-25
SRCH-438, SRCH-431 - fix the css so magnifying glass glyphicon displays properly and sayt works on main pages

peggles2:
SRCH-438 - Search magnifying glass was not displaying properly
SRCH-431 - sayt compiled assets prevents users from outside using js in their landing page.

### DIFF
--- a/app/assets/stylesheets/searches/custom_bootstrap/_sprites.less
+++ b/app/assets/stylesheets/searches/custom_bootstrap/_sprites.less
@@ -16,8 +16,8 @@
 
 // Sprite icons path
 // -------------------------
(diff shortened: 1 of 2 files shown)


MothOnMars:
Per my discussion w/ @peggles2 , we've decided that adding a test for this would be tricky, as well as fairly hacky. Several of our bootstrap/CSS-related gems are out of date and/or unsupported. We will see if we can test this more robustly if/when we update our asset infrastructure.

peggles2:
I added SRCH-431 to this pull request.  I'll update the branch name when we squash and merge.

MothOnMars:
LGTM 👍

PR #248, opened 2019-02-27
SRCH-418 - include_in_all is not allowed for indices and is deprecated

peggles2:
- remove include_in_all from queries. It is deprecated.


peggles2:
Merged to https://github.com/GSA/search-gov/pull/253

PR #252, opened 2019-02-28
SRCH-416-update queries to use text instead of string

peggles2:
The string field is unsupported for indexes created in 5.x in favor of the text and keyword fields. 

The `string` type in elasticsearch 5.6 has been replaced by the `text` (analyzed) and `keyword` (not analyzed) types.


peggles2:
I added a do not merge label to this. The text field is not available till ES 5. Therefore can not deploy this code till we move our indices to the ES 5.6 cluster.

peggles2:
moving this pr against https://github.com/GSA/search-gov/pull/253

PR #254, opened 2019-03-01
SRCH-455 ClickCounter logs errors for missing I14y documents

MothOnMars:
This fixes a bug that I discovered when I started updating click counts in production. For URLs that have clicks but are not indexed (such as clicks from URLs that had been indexed, but are no longer available), the `ClickCounter` should log an error rather than dying completely.

### DIFF
--- a/app/models/click_counter.rb
+++ b/app/models/click_counter.rb
@@ -38,5 +38,7 @@ class ClickCounter
                         handle: 'searchgov')
   rescue ActiveRecord::RecordNotFound
     Rails.logger.error("SearchgovUrl not found for clicked URL: #{url}")
+  rescue I14yDocument::I14yDocumentError
+    Rails.logger.error("I14yDocument not found for clicked URL: #{url}")
   end
 end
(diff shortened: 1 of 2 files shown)


app/models/click_counter.rb
```diff
@@ -38,5 +38,7 @@ def update_click_count(url:, count:)
                         handle: 'searchgov')
   rescue ActiveRecord::RecordNotFound
     Rails.logger.error("SearchgovUrl not found for clicked URL: #{url}")
+  rescue I14yDocument::I14yDocumentError
+    Rails.logger.error("I14yDocument not found for clicked URL: #{url}")
   end
 end
```
peggles2: This may be better to be more global so other methods can use it too.  See https://medium.com/rails-ember-beyond/error-handling-in-rails-the-modular-way-9afcddd2fe1b.

MothOnMars: `I14yDocument::I14yDocumentError` is already a custom error class that can be used by other methods. However, I realized that ``I14yDocument::I14yDocumentError` may be raised for reasons other than a missing document, so I'm going to make this message include the specific message returned by I14y.

peggles2:
LGTM 👍

PR #258, opened 2019-03-04
SRCH-422 - not_analyzed does not have any effect

peggles2:
SRCH-422 - remove not_analyzed as it does not have any effect

### DIFF
--- a/app/models/elastic_news_item.rb
+++ b/app/models/elastic_news_item.rb
@@ -9,7 +9,7 @@ class ElasticNewsItem
       dynamic: :strict,
       _analyzer: { path: "language" },
       properties: {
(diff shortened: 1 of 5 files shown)


app/models/elastic_news_item.rb
```diff
@@ -9,7 +9,7 @@ class ElasticNewsItem
       dynamic: :strict,
       _analyzer: { path: "language" },
       properties: {
-        language: { type: "string", index: :not_analyzed },
+        language: { type: 'keyword', index: true },
         rss_feed_url_id: { type: 'integer' },
```
peggles2: this line of code is already done in https://github.com/GSA/search-gov/pull/256 as well.

PR #260, opened 2019-03-06
SRCH-420 -  deprecated analyzer

peggles2:
SRCH-420 - _analyzer deprecated. update anywhere that is referencing _analyzer


peggles2:
Will create another ticket to address the correct way to do the language handling. https://cm-jira.usa.gov/browse/SRCH-472

PR #263, opened 2019-03-28
Feature/update filtered

peggles2:
filtered is deprecated in ES56 needs to be replaced with bool, must.

### DIFF
--- a/lib/elastic_mappings.rb
+++ b/lib/elastic_mappings.rb
@@ -13,7 +13,7 @@ module ElasticMappings
       publish_start_on: { type: 'date', format: 'YYYY-MM-dd' },
       publish_end_on: { type: 'date', format: 'YYYY-MM-dd', null_value: '9999-12-31' },
       title: { type: 'text', term_vector: 'with_positions_offsets' },
-      match_keyword_values_only: { type: 'keyword',
+      match_keyword_values_only: { type: 'boolean',
                                    null_value: 'false' },
       keyword_values: ElasticSettings::KEYWORD }
   ).freeze
--- a/lib/elastic_mappings.rb
+++ b/lib/elastic_mappings.rb
@@ -1,7 +1,6 @@
 module ElasticMappings
   COMMON = {
     dynamic: :strict,
-    _analyzer: { path: 'language' },
     properties: {
       language: { type: 'keyword', index: true },
       affiliate_id: { type: 'integer' },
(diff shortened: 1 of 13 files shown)


lib/elastic_mappings.rb
```diff
@@ -13,8 +13,7 @@ module ElasticMappings
       publish_start_on: { type: 'date', format: 'YYYY-MM-dd' },
       publish_end_on: { type: 'date', format: 'YYYY-MM-dd', null_value: '9999-12-31' },
       title: { type: 'text', term_vector: 'with_positions_offsets' },
-      match_keyword_values_only: { type: 'boolean',
-                                   index: :not_analyzed,
+      match_keyword_values_only: { type: 'keyword',
```
MothOnMars: Why the change from `boolean` to `keyword`? `match_keyword_values_only` should only include boolean values.

spec/models/elastic_tweet_spec.rb
```diff
@@ -15,7 +15,7 @@
     ElasticTweet.commit
   end
 
-  describe ".search_for" do
+  pending ".search_for" do
```
MothOnMars: Can you see if these specs pass now? Also, if these still need to be disabled, please add a comment re. the Twitter deprecation.

spec/support/shared_vcr_setup.rb
```diff
@@ -36,7 +36,7 @@
     /codeclimate.com/ ===  URI(request.uri).host
   end
 
-  config.ignore_request { |request| URI(request.uri).port == 9200 } #Elasticsearch
+  config.ignore_request { |request| URI(request.uri).port == 9256 } #Elasticsearch
```
MothOnMars: This should probably match the range 9200-9299, so that we are ignoring requests for multiple Elasticsearch ports. Using a range rather than an array of a couple of ports gives us the flexibility to tweak the ports for future ES versions without changing the VCR setup each time.

spec/models/elastic_tweet_spec.rb
peggles2: Ok so, back then we decided to comment out because elastic_tweet is dead code and rspec was failing but at this point discovered analyze is also causing tests cases to fail which may be the result of this one as well. So placing it back to describe for now till we get analyze set to standard. Then after we will see if rspec still fails and decide from there.

spec/support/shared_vcr_setup.rb
peggles2: good point 👍

PR #267, opened 2019-04-17
SRCH-75 upgrade to ruby 2.5.5

MothOnMars:
SRCH-75 upgrade to ruby 2.5.5
- bump scrypt
- add missing spec for RejectInvalidRequestUri
- remove redundant Dockerize installation for CircleCi


MothOnMars:
@peggles2, this PR ready for review. The PR is only failing due to Code Climate. We'll need to leave those as-is for now, as the changes suggested by CC would not be compatible with Ruby 2.3 (which will still be used in production until we update the cookbooks).

I created and then deleted an issue to clean up those complaints en masse. I think we're safer proceeding as we have been, fixing issues in specific files that we are working on. Making all the code changes for the 2.5 issues would involve re-testing two dozen different classes, which would be riskier than focusing on classes we're already updating.

.circleci/images/primary/Dockerfile
```diff
@@ -8,11 +8,4 @@ RUN apt-get update && apt-get install -y \
   libprotobuf-dev \
   protobuf-compiler
 
-# Install Dockerize to ensure services are started before test runs
```
MothOnMars: I realized this was unnecessary; the base CircleCi images contain `dockerize`.

spec/lib/middlewares/reject_invalid_request_uri_spec.rb
```diff
@@ -6,10 +6,19 @@
...
+    let(:response) { middleware.call(env) }
+    let(:status) { response.first }
 
-    it 'should return with status 400' do
-      status, header, body = middleware.call env
+    it 'returns a 400 status' do
       expect(status).to eq(400)
     end
+
+    context 'when invalid arguments are passed' do
```
MothOnMars: The bump to 2.5 resulted in `simplecov` identifying coverage missing for a condition in `RejectInvalidRequestUri`. This spec simply tests that conditional path.

app/controllers/sites/query_downloads_controller.rb
```diff
@@ -48,7 +48,7 @@ def top_queries
   def ctr(click_count, query_count)
     return '--' if click_count == 0 || query_count == 0
 
-    sprintf("%.1f%", click_count.to_f * 100 / query_count)
+    sprintf('%.1f%%', click_count.to_f * 100 / query_count)
```
MothOnMars: This was the only code change necessary for the 2.5 upgrade. See https://bugs.ruby-lang.org/issues/13761. This change is supported by both 2.3 and 2.5:
```ruby
> sprintf("%.1f%", click_count.to_f * 100 / query_count)
"150.0%"
> sprintf("%.1f%%", click_count.to_f * 100 / query_count)
"150.0%"
```

README.markdown
```diff
@@ -12,11 +12,10 @@ working development environment for Rails up and running, including the database
 
 ## Ruby
 
-You will need Ruby 2.3.8 Verify that your path points to the correct version of Ruby:
+You will need the version of Ruby specified in `.ruby-version`. Verify that your path points to the correct version of Ruby:
 
-    devbox:usasearch
     $ ruby -v
-    ruby 2.3.8p459 (2018-10-18 revision 65136) [x86_64-darwin16]
+    ruby 2.5.5p157 (2019-03-15 revision 67260) [x86_64-darwin18]
```
peggles2: I just realized if we are not going to specify the version, which is good idea you put "the version of Ruby specified in '.ruby-version', " in line 15, you may want to do the same here.      ruby 2.5.5p157 (2019-03-15 revision 67260) [x86_64-darwin18] is too specific and users won't have that exact output when they run ruby -v.  They may have ruby 2.5.5p200 or something.  I would say " For example, you should see something similar to the following:"  or "The output should start with something like ruby 2.5.1"

PR #268, opened 2019-04-22
Feature/site limit parameter

peggles2:
- Implement sitelimit parameter on Garfield

### DIFF
--- a/app/engines/i14y_formatted_query.rb
+++ b/app/engines/i14y_formatted_query.rb
@@ -11,12 +11,18 @@ class I14yFormattedQuery < FormattedQuery
   end
 
   def query_with_sites(user_query)
-    remaining_chars = QUERY_STRING_ALLOCATION - user_query.length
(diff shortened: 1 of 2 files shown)


app/engines/i14y_formatted_query.rb
```diff
@@ -11,12 +11,18 @@ def delimiters
...
+
+  def get_sites(user_query, excluded, domains)
     sites = [user_query]
-    sites << excluded unless excluded.blank?
-    sites << domains unless domains.blank?
+    sites << excluded if excluded.present?
+    sites << domains if domains.present?
     sites.join(' ')
   end
 end
```
peggles2: Rubocop reported that this method complexity was high. The method was doing too many things and too many lines. I broke it into two methods to make Rubocop happy.

peggles2: Rubocop reported that this method complexity was high. The method was doing too many things and too many lines. I broke it into two methods to make Rubocop happy.

PR #271, opened 2019-05-08
SRCH-529 - implement sitelimit parameter

peggles2:
Implement sitelimit parameter on Garfield


app/engines/domain_scope_options_builder.rb
```diff
@@ -3,12 +3,19 @@ def self.build(site:, site_limits: nil, collection: nil)
...
   def self.included_domains(collection)
     collection.url_prefixes.map do |url_prefix|
       UrlParser.strip_http_protocols(url_prefix.prefix)
     end.presence
   end
+
+  def self.strip_sitelimit_protocol(site_limits)
+    return if site_limits.blank?
+
+    site_limits.split(/\+/).
```
MothOnMars: Let's split on a simple string (`'+'`), in the spirit of only using Regexes when we need to (it also mirrors the re-joining with `'+'` at the end). I would also break this method up a bit differently for readability:
```ruby
    site_limits.split('+').map do |site_limit|
      UrlParser.strip_http_protocols(site_limit)
    end.join('+')
```

spec/engines/domain_scope_options_builder_spec.rb
```diff
@@ -30,5 +30,17 @@
         )
       end
     end
+
+    context 'when a sitelimit is passed' do
```
MothOnMars: it's a good practice to define the context within the context itself, rather than the expectation. That allows you to re-use the context conditions in subsequent specs. So in this case, you could do:
```ruby
    context 'when a sitelimit is passed' do
      let(:args) do
        { site: affiliate, site_limits: 'https://nps.gov/foo' }
      end
 
      it 'strips out the protocols' do
        expect(DomainScopeOptionsBuilder.
          build(args)).to eq(
...
```
However, we can clean that spec up even more by setting a `subject`:
```ruby
  describe '.build' do
    subject(:build) { DomainScopeOptionsBuilder.build(args) }
.
.
.
    context 'when a sitelimit is passed' do
      let(:args) do
        { site: affiliate, site_limits: 'https://nps.gov/foo' }
      end
 
      it 'strips out the protocols' do
        expect(build).to eq(
            included_domains: ['nps.gov'],
            excluded_domains: [],
            scope_ids: [],
            site_limits: 'nps.gov/foo'
          )
      end
    end
```

It would also be a good idea to add a spec for `when multiple sitelimits are passed`. That's covered by the integration-y spec in `i14y_search_spec`, but it should be tested in the class doing the work.

spec/models/i14y_search_spec.rb
```diff
@@ -95,6 +94,35 @@
     end
   end
 
+  context 'when a site limit is specified' do
+    let!(:site_domains) { affiliate.site_domains.create!(domain: 'nih.gov') }
+    let(:i14y_search) { I14ySearch.new(affiliate: affiliate,
```
MothOnMars: for multi-line `let` blocks, use `do`/`end`:
```ruby
let(:i14y_search) do
  I14ySearch.new(affiliate: affiliate,...etc
end
```

spec/models/i14y_search_spec.rb
```diff
@@ -95,6 +94,35 @@
...
+    it 'passes the sitelimits to i14y with out http/https' do
+      expect(I14yCollections).to receive(:search).
+        with(hash_including(query: 'marketplase site:nih.gov/foo'))
+      i14y_search.run
+    end
+  end
+
+  context 'when multiple site limits are specified' do
+    let!(:site_domains) { affiliate.site_domains.create!(domain: 'nih.gov') }
+    let(:i14y_search) {
```
MothOnMars: for multi-line `let` blocks, use `do`/`end`

MothOnMars:
LGTM 👍

MothOnMars:
LTGM 👍

PR #275, opened 2019-05-14
Dawnpm usagov page error update

dawnpm:
USA.gov created a search error-specific page, and I updated the redirect in the cookbooks, not realizing that the app is also redirecting folks to usa.gov a fair amount as well. This PR updates all mentions of `www.usa.gov/page-not-found` to `www.usa.gov/search-error`.


dawnpm:
The codeclimate error about the page being not found is as expected. https://www.usa.gov/search-error does return a 404 error.

app/controllers/application_controller.rb
```diff
@@ -6,7 +6,7 @@ class ApplicationController < ActionController::Base
   protect_from_forgery with: :exception
```
peggles2: At the very top of this file add this code and it will fix the code climate error.  
 # frozen_string_literal: true

dawnpm: Thanks, made the update.

dawnpm:
@peggles2 any ideas on how to get the cucumber tests to pass?

peggles2:
This pull request has been moved to - https://github.com/GSA/search-gov/pull/283

PR #276, opened 2019-05-14
SRCH-595 - upgrade active scaffold to 3.5.0

peggles2:
In order to upgrade to Rails 5 we have to upgrade active scaffold gem to version 3.5.0.
This version required some code changes. Including
Setting inverse_of in associations in the models that active scaffold uses.
"ActiveScaffold 3.4 was monkey patching association reflections to look for reverse associations, but it caused some issues and it was removed in 3.5, and then column.association was changed to return own ActiveScaffold class instead of ActiveRecord's association reflection."

Passing record parameter to options_for_association_conditions.  It took only 1 value now requires 2.


MothOnMars:
@peggles2 , this LGTM, pending the cucumber tests passing. I'm looking into those now. Per our discussion, this will be followed up by ensuring all associations have `inverse_of` set as necessary:
https://cm-jira.usa.gov/browse/SRCH-615

MothOnMars:
@peggles2 , https://github.com/GSA/search-gov/pull/279 should resolve the failure for the `Add/edit/remove RSS Feed` test, which also occurs on master. The failure for `failed Add/edit/remove i14y drawers` appears to be related to your code changes. That test passes if you revert the gem updates. (I discovered that this PR hasn't actually passed CI yet; the only CI test that was run before was CodeClimate. Seems like the manual CC approval may have prevented the rest of the checks from running.)

PR #278, opened 2019-05-15
SRCH-609 display matching site limit partial on I14y SERP

MothOnMars:
- display matching site limit partial on I14y SERP
- improve cucumber test for searching with sitelimits

### DIFF
--- a/app/models/i14y_search.rb
+++ b/app/models/i14y_search.rb
@@ -2,13 +2,14 @@ class I14ySearch < FilterableSearch
   include SearchInitializer
   include Govboxable
   I14Y_SUCCESS = 200
-  attr_reader :collection
+  attr_reader :collection, :matching_site_limits
 
   def initialize(options = {})
     super
     @enable_highlighting = !(false === options[:enable_highlighting])
     @collection = options[:document_collection]
     @site_limits = options[:site_limits]
+    @matching_site_limits = formatted_query_instance.matching_site_limits
   end
 
   def search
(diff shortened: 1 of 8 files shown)


app/models/i14y_search.rb
```diff
@@ -2,13 +2,14 @@ class I14ySearch < FilterableSearch
...
-  attr_reader :collection
+  attr_reader :collection, :matching_site_limits
 
   def initialize(options = {})
     super
     @enable_highlighting = !(false === options[:enable_highlighting])
     @collection = options[:document_collection]
     @site_limits = options[:site_limits]
+    @matching_site_limits = formatted_query_instance.matching_site_limits
   end
```
peggles2: There is a bit of redundancy with @site_limits and @matching_site_limits. I think this should be refactored.

MothOnMars: Those two instance variables are used in different ways and return different values. `@site_limits` is used by the `DomainScopeOptionsBuilder`, which provides the formatted version of the site limits. `@matching_site_limits` is the formatted site limits, which is the  instance variable required in the views.

MothOnMars: Note: `@matching_site_limits` mirrors the usage in `WebSearch`:
```ruby
app/models/web_search.rb:  

@matching_site_limits = formatted_query_instance.matching_site_limits
```
Ultimately, the various search classes should be refactored to reduce redundancy and resolve inconsistencies, but that would be an epic that we will have to tackle at another time.

PR #286, opened 2019-06-06
SRCH-674 parse alternate language metadata formats

MothOnMars:
This PR simplifies and refactors the language code handling to ensure consistent parsing of various language code formats. It also cleans up some Rubocop offenses in the `Language` class.

### DIFF
--- a/app/models/language.rb
+++ b/app/models/language.rb
@@ -1,8 +1,7 @@
 # frozen_string_literal: true
 
 class Language < ActiveRecord::Base
-  validates :code, presence: true
-  validates :name, presence: true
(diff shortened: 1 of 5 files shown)


app/models/language.rb
```diff
@@ -1,7 +1,12 @@
+# frozen_string_literal: true
+
 class Language < ActiveRecord::Base
-  validates_presence_of :code, :name
+  validates :code, presence: true
+  validates :name, presence: true
+
```
peggles2: you can combine these since they are validating same thing.

MothOnMars: 👍

PR #289, opened 2019-06-17
update query_downloads_controller_spec to return correct format

peggles2:
SRCH-696 - After rails 5 upgrade the rspec tests for query downloads failed due to slight change in the format.


MothOnMars:
LGTM 👍

peggles2:
When setting it up for testing discovered we have to set the autoloading to true for production env or we can't access the list of affiliates in admin interface.

PR #290, opened 2019-06-17
Remove federal and gobierno as trigger words, add rspec test

peggles2:
SRCH-655 - Remove specific USAJOBS trigger words - "federal" "gobierno"

### DIFF
--- a/spec/lib/jobs_spec.rb
+++ b/spec/lib/jobs_spec.rb
@@ -88,6 +88,13 @@ describe Jobs do
       end
     end
 
+    context 'when the search phrase includes "federal" or "gobierno" ' do
+      it 'should not trigger job search' do
+        %w['federal' 'gobierno'].
+          each { |phrase| expect(Jobs.query_eligible?(phrase)).to be_falsey }
+      end
+    end
+
     context 'when the search phrase is blocked' do
       it 'should return false' do
         ["employment data", "employment statistics", "employment numbers", "employment levels", "employment rate",
--- a/spec/lib/jobs_spec.rb
+++ b/spec/lib/jobs_spec.rb
@@ -90,8 +90,7 @@ describe Jobs do
 
     context 'when the search phrase does not include job-related keyword' do
       it 'does not trigger job search' do
-        %w[federal gobierno].
-          each { |phrase| expect(Jobs.query_eligible?(phrase)).to be_falsey }
+        expect(Jobs.query_eligible?('federal')).to be_falsey
       end
     end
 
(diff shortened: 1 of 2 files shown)


spec/lib/jobs_spec.rb
```diff
@@ -88,6 +88,13 @@
       end
     end
 
+    context 'when the search phrase includes "federal" or "gobierno" ' do
```
MothOnMars: @peggles2 , I think this spec context should be more generic:
`context 'when the search phrase does not include job-related keyword' do`

That makes it more explicit that the logic applies to any non-job-related keyword, not just those two.

spec/lib/jobs_spec.rb
```diff
@@ -88,6 +88,13 @@
...
+      it 'should not trigger job search' do
+        %w['federal' 'gobierno'].
```
MothOnMars: Quotes within `%w[]` are redundant:
```ruby
%w['federal' 'gobierno']
=> ["'federal'", "'gobierno'"]
```

spec/lib/jobs_spec.rb
```diff
@@ -88,6 +88,13 @@
       end
     end
 
+    context 'when the search phrase includes "federal" or "gobierno" ' do
+      it 'should not trigger job search' do
```
MothOnMars: Avoid `should` in spec descriptions:
https://github.com/rubocop-hq/rspec-style-guide#should-in-example-docstrings

`it 'does not trigger a job search' do`

spec/lib/jobs_spec.rb
```diff
@@ -88,6 +88,13 @@
...
+        %w['federal' 'gobierno'].
+          each { |phrase| expect(Jobs.query_eligible?(phrase)).to be_falsey }
```
MothOnMars: As we really only need to test that generic, non-job-related words are not query eligible, I would simplify this to:
```ruby
expect(Jobs.query_eligible?('federal')).to be_falsey
```

peggles2: I put both because in the future if a dev plans to take it out the test will fail and they will then see that we wanted those two phrases not to be used. Otherwise someone may think it was must an over site.

MothOnMars: I wouldn't worry about that, because someone would need to intentionally add those words back in for some reason. Otherwise, we could argue that we should test any number of words that someone might add as a keyword. Testing two words rather than one makes the spec less clear, because we only need to verify that unrelated words are not query eligible, rather than specific words.

PR #292, opened 2019-06-19
SRCH-659 display videos on I14y SERP

MothOnMars:
This PR ensures that the video module is displayed on the I14y SERP.

### DIFF
--- a/app/views/searches/i14y.mobile.haml
+++ b/app/views/searches/i14y.mobile.haml
@@ -10,6 +10,7 @@
   = render partial: 'med_topic', locals: { search: @search } if @search.med_topic.present?
   = render partial: 'jobs', locals: { search: @search } if @search.jobs.present?
   = render partial: 'tweets', locals: { search:  @search } if @search.has_tweets?
+  = render partial: 'govbox_video_news_items', locals: { search:  @search, search_params: @search_params } if @search.has_video_news_items?
   = render partial: 'govbox_news_items', locals: { search:  @search } if @search.has_fresh_news_items?
 
   - if @search.results.present?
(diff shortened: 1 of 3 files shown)


app/views/searches/i14y.mobile.haml
```diff
@@ -10,6 +10,7 @@
   = render partial: 'med_topic', locals: { search: @search } if @search.med_topic.present?
   = render partial: 'jobs', locals: { search: @search } if @search.jobs.present?
   = render partial: 'tweets', locals: { search:  @search } if @search.has_tweets?
+  = render partial: 'govbox_video_news_items', locals: { search:  @search, search_params: @search_params } if @search.has_video_news_items?
```
MothOnMars: @peggles2 , this ugly line is copy-pasta'd directly from `index.mobile.haml`:
https://github.com/GSA/search-gov/blob/master/app/views/searches/index.mobile.haml#L12

We've had a longstanding story to refactor and DRY up these SERP views (SRCHAR-2733), but I was explicitly asked not to do any deeper cleanup for this story. So I'm holding my nose and leaving this line as-is, to make it easy to see what is identical between the various SERPs when we eventually refactor (possibly when we retire the blended SERP).

peggles2: @search and @search_params are instance variables then they can be accessed through the partial with out needing to be passed through. You can just reference them from the partial as @search and @search_params.

MothOnMars: Yeah, that could be part of the deeper refactoring. I'm going to leave it as-is for now, to simplify the refactoring when we get to SRCHAR-2733.

PR #293, opened 2019-06-20
SRCH-688 - set users unapproved for inactive accounts

peggles2:
- Create rake task for accounts inactive for more than 90 days. Set the status to unapproved. 
- Log a message
- rspec tests


lib/tasks/user.rake
```diff
@@ -2,17 +2,26 @@ namespace :usasearch do
...
+  def set_user_not_approved(users, message, email = false)
+    users.each do |user|
+      user.set_approval_status_to_not_approved
+      user.save
+      Emailer.user_approval_removed(user).deliver_now if email
+      note = <<~NOTE.squish
+        User #{user.id},  #{user.email}, #{message}, so their approval status has been set to "not_approved".
+      NOTE
+      Rails.logger.info(note)
     end
```
peggles2: This will be refactored again when we have inactive user email setup. That is next sprint. For now it will log and not email yet.

spec/models/user_spec.rb
```diff
@@ -155,6 +155,43 @@
...
 
+  describe 'scopes' do
+    let(:approved_affiliate_user) { users(:affiliate_manager) }
+    let(:not_approved_user) { users(:affiliate_manager_with_not_approved_status) }
+    let(:pending_user) { users(:affiliate_manager_with_pending_approval_status) }
+    let(:non_affiliate) { users(:developer) }
+    let(:inactive_approved_user) { users(:inactive_developer) }
+
+    describe '.approved_affiliate' do
+      subject(:not_approved) { User.approved_affiliate }
```
MothOnMars: should be `subject(:approved_affiliate)`

spec/models/user_spec.rb
```diff
@@ -155,6 +155,43 @@
...
+      let(:approved_user) { users(:affiliate_manager) }
+      let(:not_approved_user) { users(:affiliate_manager_with_not_approved_status) }
+      let(:pending_user) { users(:affiliate_manager_with_pending_approval_status) }
+
+      it { is_expected.to include(not_approved_user) }
+      it { is_expected.not_to include(approved_user) }
+      it { is_expected.not_to include(pending_user) }
+    end
+
+    describe '.not_active' do
```
MothOnMars: @peggles2 , this is totally up to you, but I have a slight preference for `inactive` over `not_active`. I think it's more standard, and it just rolls of the...brain...a little better.  I see that the `not_active` fixture you set up is using `inactive`, as well as the spec variable: `let(:inactive_approved_user)`. We should be consistent with whichever we use.

spec/lib/tasks/user_spec.rb
```diff
@@ -47,4 +48,32 @@
       @rake[task_name].invoke
     end
   end
+
+  describe 'usasearch:user:update_inactive_approval_status' do
+    let(:task_name) { 'usasearch:user:update_inactive_approval_status' }
+
+    before do
+      @rake[task_name].reenable
+      User.destroy_all("email != 'inactivedeveloper@fixtures.org'")
```
MothOnMars: @peggles2 , you shouldn't need to destroy all the other fixtures. To ensure that specs won't fail due to `#<Emailer (class)> received :user_approval_removed with unexpected arguments`, you can add `allow` to the expectation:
```ruby
allow(Rails.logger).to receive(:info)
expect(Rails.logger).to receive(:info).with(expected_message)
```

That tells the spec, "don't fail if you receive x method with some random arguments", and only tests that the expected ones are received.

spec/lib/tasks/user_spec.rb
```diff
@@ -47,4 +48,32 @@
...
+    end
+
+    it "has 'environment' as a prereq" do
+      expect(@rake[task_name].prerequisites).to include('environment')
+    end
+
+    it 'sets inactive users to not_approved' do
+      @rake[task_name].invoke
```
MothOnMars: @peggles2 , this change is at your discretion, but this would be a good time to use the `change` method for the expectation:
```ruby
      expect{ @rake[task_name].invoke }.
        to change { inactive_user.is_not_approved? }.from(false).to(true)
```
I like that format because it ensures that the expected pre-condition is correct. For example, if someone changed the relevant fixture to have `approval_status: not_approved`, the original spec could pass as a false positive. The version using `change` would not.

spec/models/user_spec.rb
MothOnMars: Ah, scratch (some of) that. I see that Authlogic uses `not_active`, so we should be consistent with that (I'm guessing why that's why you used that in the first place...). Let's just make sure we use `not_active` consistently instead of `inactive`.

lib/tasks/user.rake
```diff
@@ -2,17 +2,26 @@ namespace :usasearch do
...
+
+    desc 'Set accounts that are not active for more than 90 days to not_approved'
+    task update_inactive_approval_status: :environment do
+      set_user_not_approved(User.not_active, 'has been inactive for 90 days')
+    end
+  end
 
-        Rails.logger.info(note)
-      end
+  def set_user_not_approved(users, message, email = false)
```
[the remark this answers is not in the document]
peggles2: Interesting! Noticed that you had done it in update_https.rake and also fake_searches.rake, and elastic_rails.rake have methods defined in them too. We can easily make these methods unique by adding the namespaces in the method name.  Generate a Jira ticket so those can get addressed as well. Thanks for bringing that up.

spec/fixtures/users.yml
```diff
@@ -78,6 +78,15 @@ developer:
   contact_name: A New Developer
   welcome_email_sent: true
 
+inactive_developer:
```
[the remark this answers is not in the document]
peggles2: That is a good point. Reason why I named it that way was to keep it consistent with the users in that fixture. We have developers and affiliate users in there. So to keep it consistent I named mine inactive_developer. I think that entire fixture would need to be refactored to not be confusing.

spec/fixtures/users.yml
```diff
@@ -134,3 +143,5 @@ affiliate_requiring_manual_approval:
   <<: *DEFAULTS
   email: affiliate_requiring_manual_approval@email.gov
   requires_manual_approval: true
+  approval_status: not_approved
```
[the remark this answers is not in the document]
peggles2: Ah yes this was the fixture I found that seemed to be incorrect because it had the approval status as approved even though it required manual approval. I was going to fix it while I was in the code. Did not know about pending_approval as one of the status's will correct.

spec/lib/tasks/user_spec.rb
peggles2: This one, I was following what we are currently doing for update_approval_status. Lets discuss

lib/tasks/user.rake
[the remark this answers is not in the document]
peggles2: Actually now looking further its best to move the guts over to a class. going to do that now. We still should consider creating a jira ticket to move those other ones into their own class under lib as well.

spec/lib/tasks/user_spec.rb
peggles2: Doesn't seem to be needed, I took it out and the test passes beautifully and fails when I break the code.

lib/user_approval.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+module UserApproval
```
MothOnMars: @peggles2 , I realize this module was refactored out of the rake task, but it would be preferable to have specs for this module directly. Right now this code is only tested indirectly via the rake task. I'll approve for now, but consider adding a module spec when refactoring.

peggles2: Yeah I thought about that. If the methods in the module fails the specs will fail when it accesses the module. Nevertheless I am not apposed to more tests so sure we can do that.

MothOnMars: 👍

spec/models/user_spec.rb
```diff
@@ -162,6 +162,8 @@
     let(:non_affiliate) { users(:developer) }
     let(:not_active_user) { users(:not_active_user) }
     let(:active_user) { users(:active_user) }
+    let(:never_active_user) { users(:never_active_user)}
```
MothOnMars: Needs a space before the closing curly bracket: `{ users(:never_active_user) }`

spec/models/user_spec.rb
```diff
@@ -162,6 +162,8 @@
...
+    let(:new_non_active_user) { users(:new_non_active_user)}
```
MothOnMars: Needs a space before the closing curly bracket: `{ users(:new_non_active_user) }`

MothOnMars:
LGTM, pending CI passing.

PR #295, opened 2019-06-20
SRCH-689 - update generalized unauthorized message

peggles2:
Login error message given have to be ambiguous so we updated the login failed and un-approved accounts to have the same message.

### DIFF
--- a/app/models/user_session.rb
+++ b/app/models/user_session.rb
@@ -2,7 +2,7 @@ class UserSession < Authlogic::Session::Base
   allow_http_basic_auth false
   consecutive_failed_logins_limit 10
   failed_login_ban_for 30.minutes
-  generalize_credentials_error_messages ''
+  generalize_credentials_error_messages 'These credentials are not recognized as valid for accessing Search.gov. Please contact search@support.digitalgov.gov if you believe this is in error.'
   logout_on_timeout true
 
   def persisted?
--- a/app/models/user_session.rb
+++ b/app/models/user_session.rb
@@ -1,11 +1,16 @@
+# frozen_string_literal: true
+
 class UserSession < Authlogic::Session::Base
+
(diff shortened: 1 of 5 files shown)


app/models/user_session.rb
```diff
@@ -2,8 +2,10 @@ class UserSession < Authlogic::Session::Base
   allow_http_basic_auth false
   consecutive_failed_logins_limit 10
   failed_login_ban_for 30.minutes
-  generalize_credentials_error_messages 'Login failed due to invalid ' \
-                                        'username and/or password.'
+  generalize_credentials_error_messages 'These credentials are not recognized as valid' \
```
MothOnMars: @peggles2 , this is getting hard to read. Let's clean this up with a constant (which we should freeze):
```ruby
# frozen_string_literal: true

class UserSession < Authlogic::Session::Base
  INVALID_LOGIN_MESSAGE = <<~MESSAGE.chomp
    These credentials are not recognized as valid for accessing Search.gov.
    Please contact search@support.digitalgov.gov if you believe this is in error.
  MESSAGE

  ...
  generalize_credentials_error_messages INVALID_LOGIN_MESSAGE
```

spec/controllers/password_resets_controller_spec.rb
```diff
@@ -1,6 +1,9 @@
 require 'spec_helper'
 
-NOT_AUTHORIZED_MESSAGE = 'You are not authorized to access Search.gov.'.freeze
+NOT_AUTHORIZED_MESSAGE = 'These credentials are not recognized as valid '\
```
MothOnMars: These could also be cleaned up with a heredoc per my comment above.

peggles2:
I added a comment for the files so other devs will know that the cookbook writes over these files.

PR #298, opened 2019-06-21
SRCH-700 update sitemaps gem to 0.2.3

MothOnMars:
This PR upgrades the `sitemaps` gem to the latest version, which is compatible with Rails 5.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -30,7 +30,7 @@ gem 'tweetstream', '~> 2.6.1' # no longer maintained?
 gem 'twitter', '~> 5.5'
 gem 'flickraw', '~> 0.9.9'
 gem 'active_scaffold', '~> 3.5.0'
-gem 'active_scaffold_export', github: 'naaano/active_scaffold_export'
+gem 'active_scaffold_export', git: 'https://github.com/naaano/active_scaffold_export'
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -30,7 +30,7 @@ gem 'tweetstream', '~> 2.6.1' # no longer maintained?
 gem 'twitter', '~> 5.5'
 gem 'flickraw', '~> 0.9.9'
 gem 'active_scaffold', '~> 3.5.0'
-gem 'active_scaffold_export', github: 'naaano/active_scaffold_export'
+gem 'active_scaffold_export', git: 'https://github.com/naaano/active_scaffold_export'
```
MothOnMars: @peggles2 , I updated this per the error message generated during `bundle install`:

The git source `git://github.com/naaano/active_scaffold_export.git` uses the `git` protocol, which transmits data without encryption. Disable this warning with `bundle config git.allow_insecure true`, or switch to the `https` protocol to keep your data secure.

peggles2: would it be possible to use https protocol?

MothOnMars: My change does use https:
`gem 'active_scaffold_export', git: 'https://github.com/...`

PR #300, opened 2019-06-26
SRCH-607 - Rails 5 fix SearchesController rspec errors

peggles2:
- Fix rspec error for searches controller spec

### DIFF
--- a/spec/controllers/searches_controller_spec.rb
+++ b/spec/controllers/searches_controller_spec.rb
@@ -387,7 +387,7 @@ describe SearchesController do
       end
     end
   end
-  
+
   context 'when Affiliate.force_mobile_format = true' do
     let(:affiliate) { affiliates(:basic_affiliate) }
 
@@ -400,7 +400,7 @@ describe SearchesController do
     it 'requests the mobile format' do
       expect(request.params['format']).to eq 'mobile'
     end
-    
+
     it { is_expected.to render_template 'layouts/searches' }
     it { is_expected.to render_template 'searches/index' }
   end
--- a/spec/controllers/searches_controller_spec.rb
+++ b/spec/controllers/searches_controller_spec.rb
@@ -387,17 +387,19 @@ describe SearchesController do
       end
     end
   end
-
+  
   context 'when Affiliate.force_mobile_format = true' do
     let(:affiliate) { affiliates(:basic_affiliate) }
 
     before do
       expect(Affiliate).to receive(:find_by_name).and_return(affiliate)
       expect(affiliate).to receive(:force_mobile_format?).and_return(true)
-      expect(request).to receive(:format=).with(:mobile)
-      get :index, query: 'gov', affiliate: affiliate.name, m: 'true'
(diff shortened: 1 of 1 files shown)


spec/controllers/searches_controller_spec.rb
```diff
@@ -387,17 +387,19 @@
       end
     end
   end
-
+
```
MothOnMars: Remove extra spaces

spec/controllers/searches_controller_spec.rb
```diff
@@ -387,17 +387,19 @@
...
+  
   context 'when Affiliate.force_mobile_format = true' do
     let(:affiliate) { affiliates(:basic_affiliate) }
 
     before do
       expect(Affiliate).to receive(:find_by_name).and_return(affiliate)
       expect(affiliate).to receive(:force_mobile_format?).and_return(true)
-      expect(request).to receive(:format=).with(:mobile)
-      get :index, query: 'gov', affiliate: affiliate.name, m: 'true'
+      get :index, query: 'gov', affiliate: affiliate.name #, m: 'true'
```
MothOnMars: remove commented parameter at the end of the line

spec/controllers/searches_controller_spec.rb
```diff
@@ -387,17 +387,19 @@
...
     end
 
+    it 'requests the mobile format' do
+      expect(request.params['format']).to eq 'mobile'
+    end
```
MothOnMars: Add a line break after this spec:
```ruby
    it 'requests the mobile format' do
      expect(request.params['format']).to eq 'mobile'
    end

    it { is_expected.to render_template 'layouts/searches' }
    it { is_expected.to render_template 'searches/index' }
```

spec/controllers/searches_controller_spec.rb
peggles2: Not sure what space you want me to remove. 
Screen Shot 2019-07-01 at 5 06 06 PM

peggles2: ahh okay so Rubymine did not highlight it because it was a tab.

spec/controllers/searches_controller_spec.rb
```diff
@@ -394,12 +394,13 @@
     before do
       expect(Affiliate).to receive(:find_by_name).and_return(affiliate)
       expect(affiliate).to receive(:force_mobile_format?).and_return(true)
-      get :index, query: 'gov', affiliate: affiliate.name #, m: 'true'
+      get :index, query: 'gov', affiliate: affiliate.name, m: 'true'
```
MothOnMars: @peggles2 , the `m: 'true'` part needs to be removed entirely. With that present, it makes this spec a false positive, because it the spec is testing that the mobile format is returned if `force_mobile_format?` is `true`, but `m: 'true'` also forces the mobile format. So this line should just be:
```ruby
get :index, query: 'gov', affiliate: affiliate.name
```

PR #302, opened 2019-06-27
SRCH-714 - Rails 5 upgrade - rspec errors for missing template errors

peggles2:
Before Rails 5, when we forget to add template for an action, we get ActionView::MissingTemplate exception. with a 406 response. Now it returns 204 response and does not return MissingTemplate exception.

### DIFF
--- a/spec/controllers/application_controller_spec.rb
+++ b/spec/controllers/application_controller_spec.rb
@@ -30,17 +30,15 @@ describe ExampleController do
     Rails.application.reload_routes!
   end
 
-  context "when a request raises an ActionView::MissingTemplate error" do
-    context "when the format for the request is not a valid format" do
-      it "should render a 204" do
-        get :missing_template, :format => "orig"
-        expect(response.code).to eq("204")
-      end
+  context 'when the format for the request is not a valid format' do
+    it 'responds with a 204' do
+      get :missing_template, :format => "orig"
+      expect(response.code).to eq('204')
     end
   end
 
-  describe "#handle_unverified_request" do
-    it "raises an error" do
+  describe '#handle_unverified_request' do
+    it 'raises an error' do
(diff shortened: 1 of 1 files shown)


spec/controllers/application_controller_spec.rb
```diff
@@ -31,16 +31,10 @@ def missing_template
   end
 
   context "when a request raises an ActionView::MissingTemplate error" do
-    context "when the format for the request is a valid format" do
-      it "should raise an error" do
-        expect { get :missing_template, :format => 'json' }.to raise_error(ActionView::MissingTemplate)
-      end
-    end
-
```
peggles2: In Rails5 it will not return a MissingTemplate error.

spec/controllers/application_controller_spec.rb
```diff
@@ -31,16 +31,10 @@ def missing_template
...
     context "when the format for the request is not a valid format" do
-      it "should render a 406" do
+      it "should render a 204" do
```
MothOnMars: We could clean this up by using single instead of double quotes throughout this spec, and by omitting the `should` from the spec description:
```ruby
it 'responds with a 204' do
```

spec/controllers/application_controller_spec.rb
```diff
@@ -30,17 +30,15 @@ def missing_template
...
 
-  context "when a request raises an ActionView::MissingTemplate error" do
-    context "when the format for the request is not a valid format" do
-      it "should render a 204" do
-        get :missing_template, :format => "orig"
-        expect(response.code).to eq("204")
-      end
+  context 'when the format for the request is not a valid format' do
+    it 'responds with a 204' do
+      get :missing_template, :format => "orig"
```
MothOnMars: Use single quotes, and may as well use current hash syntax:
```ruby
get :missing_template, format: 'orig'
```

PR #317, opened 2019-07-19
Filterable search behavior

peggles2:
Typo in filterable_search_behavior.rb


peggles2:
@MothOnMars this needs to be tested again I had a typo

PR #318, opened 2019-07-23
SRCH-726 - Bug/account approval

peggles2:
rake task for setting users with out site to un approved had a bug that is fixed in this request.

### DIFF
--- a/spec/lib/tasks/user_spec.rb
+++ b/spec/lib/tasks/user_spec.rb
@@ -1,6 +1,6 @@
 require 'spec_helper'
 
-describe 'User rake tasks', type: :task do
+describe 'User rake tasks' do
   fixtures :users, :affiliates, :memberships
 
   before(:all) do
--- a/spec/lib/tasks/user_spec.rb
+++ b/spec/lib/tasks/user_spec.rb
@@ -1,6 +1,6 @@
 require 'spec_helper'
 
-describe 'User rake tasks' do
+describe 'User rake tasks', type: :task do
   fixtures :users, :affiliates, :memberships
 
   before(:all) do
--- a/spec/lib/tasks/user_spec.rb
+++ b/spec/lib/tasks/user_spec.rb
@@ -11,14 +11,19 @@ describe 'User rake tasks' do
   end
 
   let(:user) { users(:affiliate_manager_with_no_affiliates) }
+  let(:not_active_user) { users(:not_active_user) }
+  let(:task_name) { 'usasearch:user:update_approval_status' }
+  let(:affiliate) { affiliates(:basic_affiliate) }
 
   describe 'usasearch:user:update_approval_status' do
-    let(:task_name) { 'usasearch:user:update_approval_status' }
 
     before do
       @rake[task_name].reenable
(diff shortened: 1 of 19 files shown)


spec/lib/tasks/user_spec.rb
```diff
@@ -12,15 +12,19 @@
 
   let(:user) { users(:affiliate_manager_with_no_affiliates) }
   let(:not_active_user) { users(:not_active_user) }
+  let(:task_name) { 'usasearch:user:update_approval_status' }
```
MothOnMars: I would move this back to its original location, within the `describe` block that it applies to:
```ruby
describe 'usasearch:user:update_approval_status' do
  let(:task_name) { 'usasearch:user:update_approval_status' }
```

spec/lib/tasks/user_spec.rb
```diff
@@ -30,7 +34,13 @@
       expect(user.is_not_approved?).to be true
     end
 
+    it 'leaves approved users with sites to approved' do
```
MothOnMars: reword: `'leaves approved users with sites as approved'` ("leave them as" instead of "leave them to")

spec/lib/tasks/user_spec.rb
```diff
@@ -30,7 +34,13 @@
...
+      @rake[task_name].invoke
+      expect(user_with_one_site.reload.is_approved?).to be true
+    end
+
     it 'sends admin the user_approval_removed email' do
+      User.destroy_all("email != 'affiliate_manager_with_no_affiliates@fixtures.org'")
```
MothOnMars: Is this still necessary?

spec/lib/tasks/user_spec.rb
```diff
@@ -44,6 +54,35 @@
         so their approval status has been set to "not_approved".
       MESSAGE
 
+      allow(Rails.logger).to receive(:info)
+      expect(Rails.logger).to receive(:info).with(expected_message)
+      @rake[task_name].invoke
+    end
+  end
+
+  describe 'usasearch:user:update_not_active_approval_status' do
```
MothOnMars: Looks like you've accidentally duplicated the entire `describe 'usasearch:user:update_not_active_approval_status'` block.

peggles2: GAHHH!!

spec/lib/tasks/user_spec.rb
peggles2: ah miss understood your slack message thought you wanted all the let defs together. Makes sense.

spec/lib/tasks/user_spec.rb
peggles2: users who don't have sites get an email when ther user approval is removed. It was what was on prod. I would think the test is necessary as we still send email yes.

peggles2: Oh if you mean the destroy_all yes still needed.

MothOnMars: 👍

MothOnMars:
LGTM

PR #321, opened 2019-07-24
SRCH-728 - rspec error - user_spec.rb

peggles2:
Rspec tests are now passing. This test was refactored during 
SRCH-688 - set users unapproved for inactive accounts
SRCH-726 - Bug/account approval

Just removing the pending code in the test.


peggles2:
Ah yeah that was because the changes in Master fixes that one. merged the changes and passed now.

PR #328, opened 2019-08-02
Feature/test http request methods

peggles2:
Update test http request methods. In Rails_5 request methods will accept only keyword arguments in future Rails versions.

### DIFF
--- a/spec/controllers/api/v1/agencies_controller_spec.rb
+++ b/spec/controllers/api/v1/agencies_controller_spec.rb
@@ -4,28 +4,36 @@ describe Api::V1::AgenciesController do
 
   describe "#search" do
 
-    context "when results are available" do
+    context 'when results are available' do
       before do
-        @agency = Agency.create!(name: "National Park Service", abbreviation: "NPS")
-        AgencyOrganizationCode.create!(organization_code: "NP00", agency: @agency)
-        AgencyOrganizationCode.create!(organization_code: "NP01", agency: @agency)
+        @agency = Agency.
+          create!(name: 'National Park Service', abbreviation: 'NPS')
+        AgencyOrganizationCode.create!(organization_code: 'NP00', agency: @agency)
+        AgencyOrganizationCode.create!(organization_code: 'NP01', agency: @agency)
       end
 
-      it "should return valid JSON with just the first organization code" do
-        get :search, :query => 'the nps', :format => 'json'
+      it 'should return valid JSON with just the first organization code' do
+        get :search,
+            params: {
+              query: 'the nps'
+            },
+            format: 'json'
         expect(response).to be_success
         expect(response.body).to eq({name: @agency.name, abbreviation: @agency.abbreviation,
                                  organization_code: @agency.agency_organization_codes.first.organization_code }.to_json)
       end
     end
 
-    context "when search returns nil or raises an exception" do
-      it "should return error string" do
-        get :search, :query => 'error', :format => 'json'
+    context 'when search returns nil or raises an exception' do
+      it 'should return error string' do
+        get :search,
+            params: {
+              query: 'error',
+              format: 'json'
+            }
         expect(response).not_to be_success
         expect(response.body).to match(/No matching agency could be found./)
       end
     end
   end
-
 end
--- a/spec/support/shared_sites_controller_shared_behavior.rb
+++ b/spec/support/shared_sites_controller_shared_behavior.rb
@@ -1,7 +1,15 @@
 shared_examples 'restricted to approved user' do |request_method, action, parameters = nil, sessions = nil, flash = nil|
+
+  let(:args) do
+    { params: parameters,
+    flash: flash }
+  end
+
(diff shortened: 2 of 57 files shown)


peggles2:
@MothOnMars Lots of files had to be updated 57 total. So to make it easier broken them into several commits. Hope that helps. Let me know if you have any questions.

spec/support/shared_sites_controller_shared_behavior.rb
```diff
@@ -1,7 +1,15 @@
 shared_examples 'restricted to approved user' do |request_method, action, parameters = nil, sessions = nil, flash = nil|
```
MothOnMars: These specs are still generating deprecation warnings, because you are still passing the params & flash as a positional rather than a keyword argument. These can be re-written as:
```ruby
#note the omission of `sessions=nil`, which is not needed
shared_examples 'restricted to approved user' do |request_method, action, parameters = nil, flash = nil|

  context 'when user is not logged in' do
    it 'should redirect to login page' do
      send request_method, action, params: parameters, flash: flash
...etc.
```

spec/controllers/api/v1/agencies_controller_spec.rb
```diff
@@ -4,28 +4,36 @@
...
+        @agency = Agency.
+          create!(name: 'National Park Service', abbreviation: 'NPS')
+        AgencyOrganizationCode.create!(organization_code: 'NP00', agency: @agency)
+        AgencyOrganizationCode.create!(organization_code: 'NP01', agency: @agency)
       end
 
-      it "should return valid JSON with just the first organization code" do
-        get :search, :query => 'the nps', :format => 'json'
+      it 'should return valid JSON with just the first organization code' do
+        get :search,
```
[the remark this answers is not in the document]
peggles2: Just to be consistent with same format when I was going through adding the parameters. It's readable.

spec/controllers/api/v1/agencies_controller_spec.rb
```diff
@@ -4,28 +4,36 @@
...
-        AgencyOrganizationCode.create!(organization_code: "NP01", agency: @agency)
+        @agency = Agency.
+          create!(name: 'National Park Service', abbreviation: 'NPS')
+        AgencyOrganizationCode.create!(organization_code: 'NP00', agency: @agency)
+        AgencyOrganizationCode.create!(organization_code: 'NP01', agency: @agency)
       end
 
-      it "should return valid JSON with just the first organization code" do
-        get :search, :query => 'the nps', :format => 'json'
+      it 'should return valid JSON with just the first organization code' do
```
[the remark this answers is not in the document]
peggles2: I agree, there is a lot that has to be cleaned up. I did not even get to all the double quotes, and space after parenthesis. Unfortunately with 57 files, wont be able to get to it in this pull request.

spec/support/shared_sites_controller_shared_behavior.rb
peggles2: args is already passing parameters 
``` let(:args) do
    { params: parameters, flash: flash } ```
Reason it had to be set this way is because it gives an error if you pass session and just nil. you don't need it if its nil. But I figured some tests it may not be nil. So I check see if sessions exists before adding it to args.  I do not see positional error. Can you slack message me that? I will take a look.  If you have questions slack message me and we can discuss.

Comment (peggles2) [reply spec/support/shared_sites_controller_shared_behavior.rb]: Reason it had to be set this way is because it gives an error if you pass session and just nil. you don't need it if its nil. But I figured some tests it may not be nil. So I check see if sessions exists before adding it to args.  I do not see positional error. Can you slack message me that? I will take a look.  If you have questions slack message me and we can discuss.

Comment (MothOnMars) [reply spec/support/shared_sites_controller_shared_behavior.rb]: Sorry, I realized I was seeing those deprecation warnings on the main Rails5 branch, rather than this branch. However, I still prefer the style I suggested for a couple of reasons:
- `sessions` is never used, so we can remove that
- `send request_method, action, params: parameters, flash: flash` is more explicit and readable than `send request_method, action, args`

I'm also thinking we can omit the `flash`, which we never use in the specs

File: spec/controllers/api/v1/agencies_controller_spec.rb (hunk 3 above)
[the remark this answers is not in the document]
Comment (MothOnMars) [reply spec/controllers/api/v1/agencies_controller_spec.rb]: Word.

File: spec/support/shared_sites_controller_shared_behavior.rb [hunk 4]
```diff
@@ -26,7 +20,7 @@
     before { UserSession.create(users(:affiliate_manager_with_pending_approval_status)) }
 
     it 'should redirect to affiliates page' do
-      self.send request_method, action, args
+      self.send request_method, action, params: parameters
```
MothOnMars: Let's remove the `self` in `self.send` for consistency with the other specs.

peggles2: 👍

PR #330, opened 2019-08-08
cleanup ActiveREcord::base

peggles2:
SRCH-782 - cleanup activrecord::base
All models are inherited from ApplicationRecord which is inherited by ActiveRecord::Base


app/models/concerns/content_tools.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
```
MothOnMars: delete per our discussion

app/models/concerns/search_domain.rb
```diff
@@ -25,4 +28,3 @@ def normalize_domain
 
 end
```
MothOnMars: This empty line should be removed

config/initializers/extensions/active_record.rb
```diff
@@ -1 +0,0 @@
-require 'active_record_extension'
```
MothOnMars: Can this entire file be deleted? Or maybe we should auto-load everything in `lib/extensions`?

features/step_definitions/i14y_drawer_steps.rb
```diff
@@ -3,11 +3,11 @@
  # See https://github.com/GSA/usasearch/blob/master/README_I14Y.markdown
 
 Given /^we don't want observers to run during these cucumber scenarios$/ do
-  ActiveRecord::Base.observers.disable :all
```
MothOnMars: This and subsequent calls to `ActiveRecord::Base.observers` should be left as `ActiveRecord::Base.observers`. The observer classes still inherit from `ActiveRecord::Observer`, and can be called using `ActiveRecord::Base.observers`. See https://github.com/rails/rails-observers.
```ruby
> ActiveRecord::Base.observers
[
    [0] :sayt_filter_observer,
    [1] :misspelling_observer,
    [2] :indexed_document_observer,
    [3] :affiliate_observer,
    [4] :navigable_observer,
    [5] :searchable_observer,
    [6] :rss_feed_url_observer,
    [7] :i14y_drawer_observer,
    [8] :routed_query_keyword_observer,
    [9] :watcher_observer
]
```

features/vcr_cassettes/Dashboard/Performing_site_autodiscovery.yml
```diff
@@ -152,4 +152,1335 @@ http_interactions:
         }
     http_version: 
   recorded_at: Thu, 22 Feb 2018 07:39:53 GMT
+- request:
```
MothOnMars: This recording should not be updated as part of this commit. When the Rails 5 changes are complete, we should re-run all the specs locally in case any cassettes need to be updated at that point.

config/initializers/extensions/active_record.rb
peggles2: Yes because its being loaded in ApplicationRecord. Now all models inherit.

features/step_definitions/i14y_drawer_steps.rb
peggles2: ApplicationRecord inherits from ActiveRecord::Base so this is fine.  And if you remember https://github.com/GSA/search-gov/pull/329 tests were failing till I changed it to ApplicationRecord.  
See
https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#upgrading-from-rails-4-2-to-rails-5-0

app/models/concerns/search_domain.rb
peggles2: I can remove the space but that was there from before and I was just updating the code.

peggles2: as that file needs a lot of cleanup!! :)

app/models/concerns/content_tools.rb
peggles2: 👍

features/vcr_cassettes/Dashboard/Performing_site_autodiscovery.yml
peggles2: ah that one slipped in. Good catch!

app/models/concerns/search_domain.rb
MothOnMars: I was referring to the empty line at the end of the file, which is the result of the line that was removed (`ActiveRecord::Base.extend SearchDomain`).

PR #331, opened 2019-08-08
cucumber admin center manage content test failing

peggles2:
SRCH-783 admin_center_manage_content cucumber test

### DIFF
--- a/lib/extensions/active_record_extension.rb
+++ b/lib/extensions/active_record_extension.rb
@@ -4,8 +4,7 @@ module ActiveRecordExtension
   end
 
   def destroy_on_blank(attributes, *keys)
-    attributes.each do |attribute|
-      item = attribute[1]
+    attributes.each do |k, item|
       item[:_destroy] = true if keys.all? { |key| item[key].blank? }
     end
   end
--- a/lib/extensions/active_record_extension.rb
+++ b/lib/extensions/active_record_extension.rb
@@ -4,13 +4,13 @@ module ActiveRecordExtension
   end
 
   def destroy_on_blank(attributes, *keys)
-    attributes.each do |k, item|
+    attributes.each do |_k, item|
       item[:_destroy] = true if keys.all? { |key| item[key].blank? }
     end
   end
 
   def truncate_value(field, length_limit)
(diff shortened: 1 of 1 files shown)


lib/extensions/active_record_extension.rb
```diff
@@ -4,8 +4,7 @@ def swap_error_key(from, to)
   end
 
   def destroy_on_blank(attributes, *keys)
-    attributes.each do |attribute|
-      item = attribute[1]
+    attributes.each do |k, item|
```
MothOnMars: Because the `k` parameter is unused, it should be prefixed with an underscore:
```ruby
attributes.each do |_k, item|
```
https://github.com/rubocop-hq/ruby-style-guide#underscore-unused-vars

peggles2: I'll take that one change and give you 2 changes! :)

lib/extensions/active_record_extension.rb
```diff
@@ -4,14 +4,13 @@ def swap_error_key(from, to)
...
   def destroy_on_blank(attributes, *keys)
-    attributes.each do |attribute|
-      item = attribute[1]
+    attributes.each do |_k, item|
       item[:_destroy] = true if keys.all? { |key| item[key].blank? }
     end
   end
 
   def truncate_value(field, length_limit)
-    self.send("#{field}=", self.send(field)&.truncate(length_limit))
```
MothOnMars: @peggles2 , the `self` should be used in the first case, because it's "calling a self write accessor": 
```ruby
self.send("#{field}=", send(field)&.truncate(length_limit))
```
https://github.com/rubocop-hq/ruby-style-guide#no-self-unless-required

If Rubocop was complaining, I'm guessing it's because it was tricked by `send`ing a private method rather than writing it directly, i.e. `self.whatever_field =`

peggles2: hmm yeah rubocop was complaining

peggles2: Screen Shot 2019-08-09 at 3 41 37 PM

PR #333, opened 2019-08-12
SRCH-783 - admin_center_manage_display cucumber test

peggles2:
Rails 5 params are an object now not an array. 
I also cleaned up update_attribute. Its best practice to avoid this because it skips validations. 
although aware there are several update_attributes in this file for now just cleaning up the one that the cucumber was failing on.


app/models/affiliate.rb
```diff
@@ -517,13 +517,14 @@ def load_template_schema
...
 
   def save_template_schema(saved_template_schema)
-    merged_hash = if self.template_schema.blank?
-      (Template.default.schema).deep_merge(saved_template_schema)
-    else
-      (JSON.parse(template_schema)).deep_merge(saved_template_schema)
-    end
+    merged_template_schema =
+      if template_schema.blank?
+        saved_template_schema.deep_merge(Template.default.schema)
```
MothOnMars: @peggles2 , this changes the logic of the method, because it changes the order of preference when the hashes are merged. You can see when you run the test that this change overwrites the value provided by the user (the "Verdana" font) with the default template font ('Tahoma'):
```ruby
#old code
(Template.default.schema).deep_merge(saved_template_schema)['css']['font']['default_font']
=> "Verdana"

#new code
saved_template_schema.deep_merge(Template.default.schema)['css']['font']['default_font']
=> "Tahoma"
```
It looks like the real issue here is that when the values are merged in Rails 5, the `font_family` is getting dropped on the floor, which is causing the error in the template code (`@site.load_template_schema.css.font.font_family.split(', ')`):
```ruby
#Rails 4
> (Template.default.schema).deep_merge(saved_template_schema)['css']['font']
{
    "default_font" => "Verdana",
     "font_family" => "Arial, Helvetica, Tahoma, Verdana, sans-serif"
}

#Rails 5
> (Template.default.schema).deep_merge(saved_template_schema)['css']['font']
{
    "default_font" => "Verdana"
}
```
This can be fixed without overwriting the user input by converting the params to a hash before merging:
```ruby
(Template.default.schema).deep_merge(saved_template_schema.to_hash)['css']['font']
{
    "default_font" => "Verdana",
     "font_family" => "Arial, Helvetica, Tahoma, Verdana, sans-serif"
}
```
(Normally I'd say we should update the tests to ensure that the specs break if the user input is overwritten, but this is Walking Dead code related to Search Consumer, so let's slay this :zombie: and call it a day...)

peggles2: Thanks for the suggestion but this solution breaks the rspec test:
```   1) Affiliate has_many :affiliate_templates #save_template_schema merges defaults and saves the schema
     Failure/Error: (Template.default.schema).deep_merge(saved_template_schema)['css']['font']

     NoMethodError:
       undefined method `[]' for nil:NilClass
     # ./app/models/affiliate.rb:522:in `save_template_schema'
     # ./spec/models/affiliate_spec.rb:1435:in `block (4 levels) in <top (required)>'
     # /Users/parissa/.rvm/gems/ruby-2.3.8@searchgov-rails42/gems/vcr-4.0.0/lib/vcr/util/variable_args_block_caller.rb:9:in `call_block'
     # /Users/parissa/.rvm/gems/ruby-2.3.8@searchgov-rails42/gems/vcr-4.0.0/lib/vcr.rb:188:in `use_cassette'
     # ./spec/spec_helper.rb:114:in `block (2 levels) in <top (required)>'

  2) Affiliate has_many :affiliate_templates #port_classic_theme merges existing colors into template_schema
     Failure/Error: expect(affiliate.load_template_schema.css.colors.header.header_text_color).to eq "#000000"

     NoMethodError:
       undefined method `colors' for nil:NilClass
     # ./spec/models/affiliate_spec.rb:1496:in `block (4 levels) in <top (required)>'
     # /Users/parissa/.rvm/gems/ruby-2.3.8@searchgov-rails42/gems/vcr-4.0.0/lib/vcr/util/variable_args_block_caller.rb:9:in `call_block'
     # /Users/parissa/.rvm/gems/ruby-2.3.8@searchgov-rails42/gems/vcr-4.0.0/lib/vcr.rb:188:in `use_cassette'
     # ./spec/spec_helper.rb:114:in `block (2 levels) in <top (required)>'
```

peggles2: Template.default.schema doesn't always have  ['css']['font'].  As you see in rspec ./spec/models/affiliate_spec.rb:1433 the schema is only {"schema"=>{"default"=>"default"}}

peggles2: Ok so some more digging here. You were on the right path with the to_h. But not necessary to add 
``` ['css']['font']
{
    "default_font" => "Verdana",
     "font_family" => "Arial, Helvetica, Tahoma, Verdana, sans-serif"
}  ```

Just the to_h does it. Also found stack overflow here that explains it very nicely. https://stackoverflow.com/questions/51954683/how-to-merge-default-values-into-nested-params-in-rails-5

MothOnMars: Yeah, I just included the hash keys as an example to show that it fixed the issue, not as part of the code change.

PR #336, opened 2019-08-13
SRCH-785 - sanitize will strip out an javascript in the link tag

peggles2:
Sanitize the url so no javascript can be injected into the url link.


peggles2:
To strip off the javascript need to put sanitize around the entire link.

peggles2:
actually doing it this way wont allow the user to put any javascript anyway. But I like to keep the sanitation there juuuuust in case something in the code changes.

peggles2:
@MothOnMars this is ready for testing

spec/models/affiliate_spec.rb
```diff
@@ -465,6 +465,26 @@
       expect(affiliate.errors[:base]).to include("Locale must be valid")
     end
 
+    describe 'header tagline validation' do
+      it 'validates a valid url and returns true' do
```
MothOnMars: @peggles2 , a few notes:
-  this can be DRY'd up with some `let`s - see below
- It's a good practice to avoid saving data to the database in test runs for speed purposes, so it's better to check the validation using `valid?` or the `be_valid` matcher than to check the return value of `save`. https://www.rubydoc.info/gems/rspec-rails/RSpec%2FRails%2FMatchers:be_valid
- to avoid false positives in case a different bad attribute is causing the record to be invalid, I like to check that a record has the expected error on the specific attribute, i.e.:
`expect(whatever).errors[:whatever_field].to include('is invalid')`

```ruby
    describe 'header tagline validation' do
      let(:affiliate) do
        Affiliate.new(valid_create_attributes.
          merge(header_tagline_url: header_tagline_url))
      end
      let(:header_tagline_url) { 'http://www.google.com' }

      context 'when the URL is valid' do
        it 'is valid' do
          expect(affiliate).to be_valid
        end
      end

      context 'when the URL is invalid' do
        let(:header_tagline_url) { 'foo' }

        it 'is invalid' do
          expect(affiliate).not_to be_valid
          expect(affiliate.errors[:header_tagline_url]).to include 'is not a valid URL'
        end
      end

      context 'when the URL is includes javascript' do
        let(:header_tagline_url) { 'javascript:alert(document.domain)' }

        it 'is invalid' do
          expect(affiliate).not_to be_valid
          expect(affiliate.errors[:header_tagline_url]).to include 'is not a valid URL'
        end
      end
    end
```

It's worth noting that we have a whole bunch of different URL attributes for our models. I've long wanted to consolidate the specs for those (see `spec/support/fetchable_behavior.rb` for a start), but the various URL fields have been added by different people over the years and normalized & validated in different ways. So eventually this type of thing should be simplified (possibly with the [gem matcher](https://github.com/perfectline/validates_url#with-activerecord) from `validate_url`), but this will be fine to get this out the door.

peggles2: slack messaged you some questions. thanks!

PR #337, opened 2019-08-14
SRCH-779 - update validations to use \A instead of ^

peggles2:
SRCH-779 - Use \A and \z to match the start and end of the string, ^ and $ match the start/end of a line.
https://guides.rubyonrails.org/security.html#regular-expressions

### DIFF
--- a/spec/controllers/help_docs_controller_spec.rb
+++ b/spec/controllers/help_docs_controller_spec.rb
@@ -31,7 +31,9 @@ describe HelpDocsController do
     end
 
     context 'when url does match search.gov with newline characters' do
-      let(:url) { "https://search.gov/manual/site-information.html\n<script>dangerous_stuff();</script>" }
+      let(:url) do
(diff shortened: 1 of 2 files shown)


peggles2:
I also added a rspec test to test this as well.

spec/controllers/help_docs_controller_spec.rb
```diff
@@ -30,6 +30,17 @@
       it { is_expected.to redirect_to('https://www.usa.gov/search-error') }
     end
 
+    context 'when url does match search.gov with newline characters' do
+      let(:url) { "https://search.gov/manual/site-information.html\n<script>dangerous_stuff();</script>" }
```
MothOnMars: For readability, this long line can be wrapped as:
```ruby
let(:url) do
  "https://search.gov/manual/site-information.html\n<script>dangerous_stuff();</script>"
end
```

PR #338, opened 2019-08-15
SRCH-775 - rspec test for link_to_federal_register_advanced_search

peggles2:
Add rspec test for link_to_federal_register_advanced_search

### DIFF
--- a/spec/helpers/federal_register_documents_helper_spec.rb
+++ b/spec/helpers/federal_register_documents_helper_spec.rb
@@ -29,8 +29,8 @@ describe FederalRegisterDocumentsHelper do
     it { is_expected.to eq result }
   end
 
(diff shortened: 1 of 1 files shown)


spec/helpers/federal_register_documents_helper_spec.rb
```diff
@@ -29,6 +29,25 @@
     it { is_expected.to eq result }
   end
 
+  describe 'federal register links' do
```
MothOnMars: Because this is testing the `link_to_federal_register_advanced_search` method, the test description & subject should include that method:
```ruby
describe `#link_to_federal_register_advanced_search` do
  subject(:link_to_federal_register_advanced_search) do
...etc
```

PR #339, opened 2019-08-15
SRCH-774 - add a rspec test for federal register page info 

peggles2:
SRCH-774 - add a rspec test for federal register page  info


spec/helpers/federal_register_documents_helper_spec.rb
```diff
@@ -29,6 +34,20 @@
     it { is_expected.to eq result }
   end
 
+  describe '#federal_register_document_page_info' do
+    subject(:federal_register_document_info) do
+      helper.federal_register_document_page_info(document)
+    end
+
+    let(:result) do
+      <<~HTML.delete!("\n")
```
MothOnMars: This does not need to be a heredoc, because it's a one-line string:
```ruby
let(:result) { 'Pages 0 - 47 (0 pages) [FR DOC #: 2019-55555]' }
```

spec/helpers/federal_register_documents_helper_spec.rb
```diff
@@ -29,6 +34,16 @@
...
   end
 
+  describe '#federal_register_document_page_info' do
+    subject(:federal_register_document_info) do
+      helper.federal_register_document_page_info(document)
+    end
+
+    let(:result) { 'Pages 0 - 47 (0 pages) [FR DOC #: 2019-55555]' }
+
+    it { is_expected.to eq result }
```
MothOnMars: Another thing I should have caught last time:

This spec is simple and short enough that there's no reason to use a `let` statement for the result. This can be simplified as: 
```ruby
it { is_expected.to eq 'Pages 0 - 47 (0 pages) [FR DOC #: 2019-55555]' }
```

PR #350, opened 2019-08-23
SRCH-539 set size to 100 for news item terms aggregations

MothOnMars:
@peggles2, these aggregations appear to have been used in the sidebar of the legacy SERP - I don't believe they are used at all by the current SERPs, but for safety's sake I'm going to leave them as part of this query. The story above indicates that we set `size: 0`  to return the max. number of bucketed aggregations in order to "make it easy for users to see all ...DC subjects in the left nav". 100 should be plenty for this likely deprecated feature.

This PR fixes the above, and sets the remaining failing news item specs to `pending`.


MothOnMars:
Closing this up - I'm going to combine this with a couple of other fixes.

PR #355, opened 2019-08-27
SRCH-828 re-enable ElasticBoostedContent specs where possible

MothOnMars:
This PR fixes and re-enables specs that were failing erroneously. Several of the specs in elastic_boosted_content_spec.rb were failing because we have temporarily switched from per-language analyzers to the `standard` analyzer. This broke specs where, for example, a query for `carrot` is not returning results for `carrots`. The specs in question were not intended to test language analysis, so it's preferred to simplify those specs; language analysis has its own specs, which have been commented out temporarily.

### DIFF
--- a/spec/models/elastic_boosted_content_spec.rb
+++ b/spec/models/elastic_boosted_content_spec.rb
@@ -10,10 +10,11 @@ describe ElasticBoostedContent do
       q: query,
       affiliate_id: affiliate.id,
       size: 1,
-      offset: 1,
+      offset: 0,
       language: affiliate.indexing_locale
     }
   end
+  let(:search) { ElasticBoostedContent.search_for(search_params) }
 
   before do
     ElasticBoostedContent.recreate_index
@@ -76,7 +77,7 @@ describe ElasticBoostedContent do
     end
   end
 
-  describe "highlighting results" do
+  describe 'highlighting results' do
     before do
       affiliate.boosted_contents.create!(title: 'Tropical Hurricane Names',
                                          status: 'active',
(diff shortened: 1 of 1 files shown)


spec/models/elastic_boosted_content_spec.rb
```diff
@@ -66,9 +76,7 @@
     end
   end
 
-  # Temporarily disabling these specs during ES56 upgrade
-  # https://cm-jira.usa.gov/browse/SRCH-828
-  pending "highlighting results" do
+  describe "highlighting results" do
```
peggles2: Lets do single quotes.

spec/models/elastic_boosted_content_spec.rb
```diff
@@ -98,7 +106,7 @@
       end
 
       it 'should escape the entity but show the highlight' do
-        search = ElasticBoostedContent.search_for(q: 'carrot', affiliate_id: affiliate.id, language: affiliate.indexing_locale)
+        search = ElasticBoostedContent.search_for(q: 'carrots', affiliate_id: affiliate.id, language: affiliate.indexing_locale)
```
peggles2: Line is too long.  Try something like this below. It will also be easier to read.
```
search = ElasticBoostedContent.search_for(
  q: 'carrots',
  affiliate_id: affiliate.id, 
  language: affiliate.indexing_locale
)

```

spec/models/elastic_boosted_content_spec.rb
```diff
@@ -313,27 +323,35 @@
     end
 
     describe "title and description" do
-      it 'should be case insentitive' do
-        expect(ElasticBoostedContent.search_for(q: 'OBAMA', affiliate_id: affiliate.id, language: affiliate.indexing_locale).total).to eq(1)
-        expect(ElasticBoostedContent.search_for(q: 'BIDEN', affiliate_id: affiliate.id, language: affiliate.indexing_locale).total).to eq(1)
+      it 'is case insentitive' do
+        expect(ElasticBoostedContent.search_for(search_params.merge(q: 'YOSEMITE')).total)
+          .to eq(1)
       end
```
peggles2: Place the . on the previous line, together with the method call receiver.

MothOnMars:
Mmm, tacos!!!

PR #356, opened 2019-08-27
SRCH-821 re-enable ElasticSaytSuggestion specs where possible

MothOnMars:
@peggles2 , these changes are similar to https://github.com/GSA/search-gov/pull/355. I have re-enabled specs that were failing only because their searches required English text analysis (i.e. searching for `suggest` to find a doc with `suggests`). Because the per-language analysis is tested separately, I have disabled the per-language specs, and updated the other specs so they do not depend on English analyzers.

### DIFF
--- a/spec/models/elastic_sayt_suggestion_spec.rb
+++ b/spec/models/elastic_sayt_suggestion_spec.rb
@@ -1,10 +1,18 @@
 require 'spec_helper'
 
-# Temporarily disabling these specs during ES56 upgrade
-# https://cm-jira.usa.gov/browse/SRCH-821
 describe ElasticSaytSuggestion do
   fixtures :affiliates
   let(:affiliate) { affiliates(:basic_affiliate) }
+  let(:search_params) do
+    {
+      q: 'suggest',
+      affiliate_id: affiliate.id,
+      size: 1,
+      offset: 0,
+      language: affiliate.indexing_locale
+    }
+  end
+  let(:search) { ElasticSaytSuggestion.search_for(search_params) }
 
   before do
     ElasticSaytSuggestion.recreate_index
@@ -12,9 +20,19 @@ describe ElasticSaytSuggestion do
     affiliate.locale = 'en'
   end
 
-  pending ".search_for" do
-    describe "results structure" do
+  describe '.search_for' do
+    describe 'results structure' do
       context 'when there are results' do
+        let(:search_params) do
+          {
+            q: 'suggest',
+            affiliate_id: affiliate.id,
+            size: 1,
+            offset: 1,
+            language: affiliate.indexing_locale
+          }
+        end
+
         before do
           affiliate.sayt_suggestions.create!(phrase: 'hi suggest me', popularity: 30)
           affiliate.sayt_suggestions.create!(phrase: 'suggest me too', popularity: 29)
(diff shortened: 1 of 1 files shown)


spec/models/elastic_sayt_suggestion_spec.rb
```diff
@@ -1,29 +1,47 @@
...
           affiliate.sayt_suggestions.create!(phrase: 'suggest me too', popularity: 29)
           affiliate.sayt_suggestions.create!(phrase: 'suggest me three suggests', popularity: 28)
           ElasticSaytSuggestion.commit
         end
 
-        it 'should return results in an easy to access structure ordered by most popular' do
-          search = ElasticSaytSuggestion.search_for(q: 'suggests', affiliate_id: affiliate.id, size: 1, offset: 1, language: affiliate.indexing_locale)
+        it 'returns results in an easy to access structure ordered by most popular' do
+          search = ElasticSaytSuggestion.search_for(search_params)
           expect(search.total).to eq(3)
```
peggles2: Just a note in rails 5 this will be 
          `search = ElasticSaytSuggestion.search_for(params: search_params)`

May want to create a Jira ticket when this goes to pro so I can update it in the rails 5 release branch.

spec/models/elastic_sayt_suggestion_spec.rb
```diff
@@ -38,33 +56,32 @@
...
+        search = ElasticSaytSuggestion.search_for(search_params)
         first = search.results.first
         expect(first.phrase).to eq("hi <strong>suggest</strong> me")
       end
     end
 
     context 'when highlight is turned off' do
       it 'should not highlight matches' do
-        search = ElasticSaytSuggestion.search_for(q: 'suggests', affiliate_id: affiliate.id, language: affiliate.indexing_locale, highlighting: false)
+        search = ElasticSaytSuggestion.search_for(q: 'suggest', affiliate_id: affiliate.id, language: affiliate.indexing_locale, highlighting: false)
```
peggles2: This line is very long. You can break it something like below. Or even better I would make it its own variable as rails 5 we have to also assign it to params hash. 

```
search = ElasticSaytSuggestion.search_for(
  q: 'suggest', 
  affiliate_id: affiliate.id, 
  language: affiliate.indexing_locale, 
  highlighting: false
)
```

spec/models/elastic_sayt_suggestion_spec.rb
```diff
@@ -86,14 +103,16 @@
...
       before do
         affiliate.sayt_suggestions.create!(phrase: 'the exact match', popularity: 30)
         ElasticSaytSuggestion.commit
       end
 
-      it "should ignore exact matches regardless of case" do
+      # Temporarily disabling these specs during ES56 upgrade
+      # https://cm-jira.usa.gov/browse/SRCH-838
+      xit "should ignore exact matches regardless of case" do
         ['the exact match', 'THE EXACT MATCH'].each do |query|
```
peggles2: Change the double quotes to single quotes.

spec/models/elastic_sayt_suggestion_spec.rb
```diff
@@ -1,29 +1,47 @@
...
+        let(:search_params) do
+          {
+            q: 'suggest',
+            affiliate_id: affiliate.id,
+            size: 1,
+            offset: 1,
+            language: affiliate.indexing_locale
+          }
+        end
+
```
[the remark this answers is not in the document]
MothOnMars: That was my initial plan, but the inclusion of `offset: 1` in the `.search_for` specs makes the params slightly different from what I would expect to be the default (`offset: 0`). I was trying to avoid refactoring the specs to use a different offset. 

That said, it clearly bugs us both that these params are not DRY, so I'm going to suck it up and do that refactoring. So there will be some additional spec changes coming beyond the `search_params` reuse.

spec/models/elastic_sayt_suggestion_spec.rb
MothOnMars: Per our discussion, this does not need to change for Rails 5.

PR #365, opened 2019-09-04
SRCH-820 - remove the template_not_found method not needed for rails 5

peggles2:
The proper way to return 406 error for wrong format is through defining the format in the controller then anything else will return a 406. Currently what is on production is not working and we get 500 errors for the wrong format. Rails 5 does not even go through this method anyway so taking it out.


MothOnMars:
P.S. 🌮

PR #366, opened 2019-09-04
Feature/rubocop errors

peggles2:
Because I moved the concerns into the correct directory, rubocop is spitting out the errors for these files. I cleaned up the low hanging fruit ones but we will have to plug our nose and silence them again.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -449,6 +449,7 @@ GEM
     naught (1.1.0)
     net-http-persistent (2.9.4)
     newrelic_rpm (5.0.0.342)
+    nio4r (2.3.1)
     nokogiri (1.10.4)
       mini_portile2 (~> 2.4.0)
     nokogumbo (1.5.0)
--- a/app/models/concerns/best_bet.rb
+++ b/app/models/concerns/best_bet.rb
@@ -32,8 +32,7 @@ module BestBet
   end
 
   def match_keyword_values_only_requires_keywords(keywords_collection)
-    if keywords_collection.reject(&:marked_for_destruction?).blank?
-      && match_keyword_values_only?
+    if keywords_collection.reject(&:marked_for_destruction?).blank? && match_keyword_values_only?
       errors.add(:match_keyword_values_only, 'requires at least one keyword')
     end
   end
--- a/app/models/concerns/best_bet.rb
+++ b/app/models/concerns/best_bet.rb
@@ -32,8 +32,8 @@ module BestBet
   end
 
   def match_keyword_values_only_requires_keywords(keywords_collection)
-    if keywords_collection.
-         reject(&:marked_for_destruction?).blank? && match_keyword_values_only?
+    if keywords_collection.reject(&:marked_for_destruction?).blank?
+      && match_keyword_values_only?
       errors.add(:match_keyword_values_only, 'requires at least one keyword')
     end
   end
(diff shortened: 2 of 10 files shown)


app/models/concerns/best_bet.rb
```diff
@@ -12,13 +14,14 @@ module BestBet
...
-    validates_inclusion_of :status, in: STATUSES, message: 'must be selected'
+    validates :status, inclusion: { in: STATUSES, message: 'must be selected' }
+
     validate :publish_start_and_end_dates
   end
 
   def active_and_searchable?
     if publish_end_on
-      is_active? && (publish_start_on..publish_end_on).include?(Date.current)
+      is_active? && (publish_start_on..publish_end_on).cover?(Date.current)
```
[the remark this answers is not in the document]
peggles2: Rubocop reported it. Agree may be better to refactor it in the future.

app/models/concerns/best_bet.rb
```diff
@@ -29,8 +32,9 @@ def display_status
   end
 
   def match_keyword_values_only_requires_keywords(keywords_collection)
-    if keywords_collection.reject(&:marked_for_destruction?).blank? && match_keyword_values_only?
-      self.errors.add(:match_keyword_values_only, 'requires at least one keyword')
+    if keywords_collection.
```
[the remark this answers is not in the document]
peggles2: hmm this would give an error unexpected &&

app/models/concerns/time_window_validator.rb
```diff
@@ -3,11 +3,13 @@ class TimeWindowValidator < ActiveModel::EachValidator
   MAX_DAYS_BACK = 29
 
   def validate_each(record, attribute, value)
-    record.errors.add attribute, (options[:message] || "exceeds the maximum of #{MAX_DAYS_BACK} days") unless recent_enough?(value)
```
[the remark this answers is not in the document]
peggles2: This would be incorrect. You would want to only return if recent_enough?(value) otherwise want to run record.errors.add().  So correct way is 
```
def validate_each(record, attribute, value)
  return if recent_enough?(value)

  record.errors.add(attribute, (options[:message] ||
                                "exceeds the maximum of #{MAX_DAYS_BACK} days"))
end
```

peggles2:
@MothOnMars this is ready for review

Gemfile.lock
```diff
@@ -449,6 +449,7 @@ GEM
     naught (1.1.0)
     net-http-persistent (2.9.4)
     newrelic_rpm (5.0.0.342)
+    nio4r (2.3.1)
```
MothOnMars: What prompted this gem addition?

app/models/concerns/best_bet.rb
```diff
@@ -30,7 +33,7 @@ def display_status
 
   def match_keyword_values_only_requires_keywords(keywords_collection)
     if keywords_collection.reject(&:marked_for_destruction?).blank? && match_keyword_values_only?
```
MothOnMars: Github is still possessed  - my comment disappeared again. Hopefully the 3rd time will be a charm...

A guard clause is preferred to a conditional here:
```ruby
  def match_keyword_values_only_requires_keywords(keywords_collection)
    return unless keywords_collection.reject(&:marked_for_destruction?).blank? &&
      match_keyword_values_only?

    errors.add(:match_keyword_values_only, 'requires at least one keyword')
  end
```

Gemfile.lock
peggles2: I keep taking it out it keeps sneaking back in. hang on.

peggles2: We discussed this is actually needed for actioncable.

PR #372, opened 2019-09-20
SRCH-857 - add uid  to users table

peggles2:
Add uid r field to our user table so that we can use login.gov to authenticate.

DB - Migration is needed for this change.


db/structure.sql
```diff
@@ -2306,6 +2308,8 @@ INSERT INTO `schema_migrations` (version) VALUES
 ('20181025225740'),
 ('20181109212904'),
 ('20181213153332'),
-('20190205200912');
+('20190205200912'),
+('20190920174118'),
```
MothOnMars: `20190920174118` references a non-existent migration (maybe a previous version of this migration?), so it should be removed.

peggles2: We are planning to use email address for lookups. That is the only field currently common with the app and login.gov to allow us to find pre-existing users. 

uid and provider returned from login.gov and will always be unique. Login.gov handles that. The user will never input this value so a dupe would never happen.

The only time we will be looking up uid and providers would be if we have authentication from multiple providers. Which I doubt will happen. We will always have users login with login.gov.

we don't want to limit the size of UID. We want login.gov to return the proper UID. Provider will always be logindotgov and again no limit necessary.

MothOnMars: If we are expecting the UID and provider to be unique, it would still be a good idea to add a unique index. Users won't be inputting that information directly, but adding a uniqueness check at the db level is a best practice that will protect against any internal data management issues, such as bugs or accidental bad updates by a script or a dev.

dawnpm: Chiming in here to add something I heard in our meeting with them - that the UID associated with a given email address might change. Does this current approach replace an existing Login.gov UID with a new one, if they're different? The UID would change in the event that a user had to delete their account and start fresh, similar to what might happen in our context if an affiliate gets deleted and a new one created using the previous site handle.

peggles2: When a user wishes to change their email address they have to change it on login.gov
https://www.login.gov/help/changing-settings/change-my-email-address/.
But then after that if they try to sign in, obviously the new email address will not find the user's record and it will create a new one. 
If they go on search.gov and change their email address. Then go to login.gov and sign it with old email address then again, search.gov will not be able to find their account and would create a new one.
A user would have to first change their email on search.gov then go to login.gov and create the new email to log in. 
An alternative is if we can't find the user by the email address to lookup the user by uid. If we find then we update the account with new email
If cant find by email or by uid then its a new user and new account is created.

PR #376, opened 2019-09-24
SRCH-886 resolve additional Rubocop failures

MothOnMars:
This PR is a follow-up to https://github.com/GSA/search-gov/pull/375. It contains an additional batch of stylistic cleanup per Rubocop.


app/models/custom_index_queries/elastic_link_popularity_query.rb
```diff
@@ -17,13 +17,13 @@ def body
...
-                    json.set! "params.url", links
+                    json.set! 'params.url', links
                   end
                 end
                 json.child! do
                   json.range do
-                    json.set! "@timestamp" do
-                      json.gt "now-#{@days_back}d/d"
+                    json.set! '@timestamp' do
+                      json.gt 'now-#{@days_back}d/d'
```
peggles2: #{@days_back} is a variable. So in this case you want string interpolation to resolve the variable.

MothOnMars: Yikes! Good catch. I'm also going to add a spec for this (which will be rewritten as part of the Logstash upgrade, but I'll add one to cover us for now), and move this file to `app/models/logstash_queries`.

peggles2: Yeah I was going to say rspec should have caught that. 👍

PR #384, opened 2019-10-14
login.gov - non gov users

peggles2:
In this pull request if the email is a non gov address, they are set to pending approval instead of pending email verification. Login.gov does the authorization so this step is no longer needed. Also cleaned up tests.


features/admin_center_dashboard.feature
```diff
@@ -123,8 +123,8 @@ Feature: Dashboard
     And I follow "Manage Users"
     Then I should see the following table rows:
       | Affiliate Manager affiliate_manager@fixtures.org            |
```
MothOnMars: When updating cucumber tests, keep the "table" pipes aligned:
```
| My table | This is my awesome table |
| Row 1    | row 1 content            |
```

features/admin_center_dashboard.feature
```diff
@@ -123,8 +123,8 @@ Feature: Dashboard
...
-      | Jane Admin admin2@fixtures.gov (Pending email verification) |
-      | John Admin admin1@fixtures.gov (Pending email verification) |
+      | Jane Admin admin2@fixtures.gov |
```
MothOnMars: These lines should still include the approval status, to confirm that the expected info appears on the page:
```
| Jane Admin admin2@fixtures.gov (Pending approval) |
| John Admin admin1@fixtures.gov (Pending approval) |
```

features/users.feature
```diff
@@ -2,6 +2,17 @@ Feature: Users
 
   @javascript
   Scenario: Logged-in non-developer user visits account page
```
MothOnMars: The `Logged-in non-developer user visits account page` scenario does not need to change for this story. The scenario that should be updated is `Registering as a new affiliate user without government affiliated email address`, which tests the behavior for a user with a non-.gov/.mil email address, who is not yet approved.

features/users.feature
```diff
@@ -2,6 +2,17 @@ Feature: Users
...
+    Given I am logged in with email "affiliate_manager_with_pending_approval_status@fixtures.org"
+    When I go to the user account page
+    Then I should see the browser page titled "My Account"
+    And I should see "Contact Information"
+    And I should see "Name"
+    And I should see "Agency"
+    And I should see "Email"
+    And I should see "Because you don't have a .gov or .mil email address, we need additional information. If you are a contractor on an active contract, please use your .gov or .mil email address on this account, or have your federal POC email to confirm your status."
```
MothOnMars: Cucumber tests don't need to check for the full message. To keep the line lengths under control, you could limit this to:
```
And I should see "Because you don't have a .gov or .mil email address, we need additional information."
```

spec/models/user_spec.rb
```diff
@@ -227,50 +197,21 @@
       end
     end
 
-    it "should set requires_manual_approval if the user is an affiliate and the email is not government_affiliated" do
-      %w( aff@agency.COM aff@anotheragency.com admin.gov@agency.org anotheradmin.MIL@agency.ORG escape_the_dot@foo.xmil ).each do |email|
+    it 'should set requires_manual_approval if the user is an affiliate and the email is not government_affiliated' do
```
MothOnMars: When cleaning up specs, you can also remove the `should`s:
`it 'sets requires_manual_approval...`

spec/models/user_spec.rb
```diff
@@ -227,50 +197,21 @@
...
+      %w[aff@agency.COM aff@anotheragency.com admin.gov@agency.org anotheradmin.MIL@agency.ORG escape_the_dot@foo.xmil].each do |email|
```
MothOnMars: As long as you're cleaning this up, it's a good time to break up this long line:
```ruby
%w[aff@agency.COM
   aff@anotheragency.com 
...
```

app/models/user.rb
```diff
@@ -189,7 +165,7 @@ def detect_deliver_welcome_email
 
   def set_initial_approval_status
     if approval_status.blank? || invited
-      set_approval_status_to_pending_email_verification
+      set_approval_status_to_pending_approval
```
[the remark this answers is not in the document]
peggles2: Added this to from_omniauth method in the user model so that we set the approval status after the user is authenticated.

features/admin_center_dashboard.feature
```diff
@@ -122,9 +122,9 @@ Feature: Dashboard
     When I go to the usagov's Dashboard page
     And I follow "Manage Users"
     Then I should see the following table rows:
-      | Affiliate Manager affiliate_manager@fixtures.org            |
-      | Jane Admin admin2@fixtures.gov (Pending email verification) |
-      | John Admin admin1@fixtures.gov (Pending email verification) |
+      | Affiliate Manager affiliate_manager@fixtures.org  |
+      | Jane Admin admin2@fixtures.gov (Pending approval) |
+      | John Admin admin1@fixtures.gov (Pending approval) |
```
peggles2: This may be changed to (Approved) when I start working on SRCH-891 - Ensure admin can invite user. But for now leaving it the way it currently behaves.

app/models/user.rb
```diff
@@ -226,15 +201,14 @@ def perishable_token_expired?
...
-  end
-
   def self.from_omniauth(auth)
     find_or_create_by(email: auth.info.email).tap do |user|
       user.update(uid: auth.uid)
+      if user.requires_manual_approval?
+        user.set_approval_status_to_pending_approval
+      else
+        user.set_approval_status_to_approved
+      end
```
MothOnMars: This logic should not run in the `from_omniauth` method, because we don't need to run this logic every time a user logs in. The logic should probably be in the `set_initial_approval_status` method, or one of the related methods that only runs during account creation.

features/admin_center_dashboard.feature
MothOnMars: Per our discussion, the expected status for each of these users is "Approved".

app/models/user.rb
peggles2: Let's chat. I still want to leave this till I dig into that section.

features/admin_center_dashboard.feature
peggles2: Per our discussion this remains blank does not show (Approved) when approved.

features/admin_center_dashboard.feature
MothOnMars: Just noting per our discussion that the approval status for approved users does not appear on the "Manage Users" page, so the scenario no longer checks for the approval status on the page.

peggles2: Just checks the users are there.

spec/models/user_spec.rb
```diff
@@ -213,10 +180,19 @@
...
+        expect(user.is_approved?).to be true
+      end
+    end
+
+    it 'should set approval status to pending approval' do
+      %w[aff@agency.COM aff@anotheragency.com admin.gov@agency.org anotheradmin.MIL@agency.ORG
+         escape_the_dot@foo.xmil].each do |email|
         user = User.create!(@valid_affiliate_attributes.merge(email: email))
-        expect(user.is_pending_email_verification?).to be true
+        expect(user.is_pending_approval?).to be true
```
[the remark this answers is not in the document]
peggles2: @MothOnMars  Yes this was prior code that I was just tweaking. So it had 'should'. I removed it but we don't need to worry about them as its deprecated in rails 5.1 and all that will be cleaned up.
Need to focus on finishing this up. Thanks!

MothOnMars:
LGTM 👍

PR #385, opened 2019-10-15
SRCH-955 pass an array of hosts to Elasticsearch client

MothOnMars:
This PR updates the configuration for the ES client used by the custom indices. This allows us to pass all the ES hosts directly to the client, rather than a single haproxy listener.

I am adding a "Do Not Merge" label, as this depends on a corresponding update in the cookbooks.

### DIFF
--- a/config/elasticsearch_client.yml
+++ b/config/elasticsearch_client.yml
@@ -1,6 +1,6 @@
 default: &DEFAULT
   log: false
-  randomize_connections: true
+  randomize_hosts: true
   reload_connections: true
   reload_on_failure: true
   retry_on_failure: 1
--- a/config/elasticsearch_client.yml
+++ b/config/elasticsearch_client.yml
@@ -1,0 +1,16 @@
+default: &DEFAULT
+  log: false
+  randomize_connections: true
+  reload_connections: true
+  reload_on_failure: true
+  retry_on_failure: 1
+
+development:
+  <<: *DEFAULT
+  retry_on_failure: false
+
+test:
+  <<: *DEFAULT
+
+production:
+  <<: *DEFAULT
--- a/config/secrets.yml.dev
+++ b/config/secrets.yml.dev
@@ -78,15 +78,15 @@ dev_settings: &DEV_SETTINGS
   custom_indices:
     elasticsearch:
       reader:
-        host: http://localhost:9200
+        hosts:
+          - http://localhost:9200
         user: elastic
         password: changeme
-        log: false
       writers:
-        - host: http://localhost:9200
+        - hosts:
+            - http://localhost:9200
           user: elastic
           password: changeme
-          log: false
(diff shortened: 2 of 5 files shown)


config/elasticsearch_client.yml
```diff
@@ -0,0 +1,16 @@
+default: &DEFAULT
+  log: false
+  randomize_connections: true
+  reload_connections: true
+  reload_on_failure: true
+  retry_on_failure: 1
```
peggles2: why is retry_on_failure here 1 but below you set it to false. To be consistent can we use either true, false or 1,0 ?  or is that the number of retries? In that case should retry_on_failure be 0?

config/secrets.yml.dev
```diff
@@ -78,15 +78,15 @@ dev_settings: &DEV_SETTINGS
...
       reader:
-        host: http://localhost:9200
+        hosts:
+          - http://localhost:9200
         user: elastic
         password: changeme
-        log: false
       writers:
-        - host: http://localhost:9200
+        - hosts:
```
peggles2: does this need to be a list since we only have one for writters?

config/elasticsearch_client.yml
MothOnMars: The `retry_on_failure` value can either be a boolean or an integer indicating the # of times to retry. I am setting it to `1` to retry once in production, because I don't want it retrying the default 3 times (due to the potential for overloading the ES servers with retried queries when there are system issues, as we've seen with i14y). I've left the default value in test to mirror the prod values, but I set it to `false` in dev, because when developing I'd like a fast failure.

config/secrets.yml.dev
MothOnMars: There is only one host in development & test environments, but in production there will be an array of hosts.

config/elasticsearch_client.yml
peggles2: would 0 give you the same result? It would be better to keep it consistent.

config/secrets.yml.dev
peggles2: correct but you are making - hosts: a list for writers.  - http://localhost:92000 is a list for the hosts.

MothOnMars: We currently only using a single writer in all environments, but I will most likely add another writer when we upgrade to ES 6. Each writer will have an array of hosts. That will (theoretically) allow us to temporarily write to two clusters, enabling us to rebuild our indices in ES 6 while keeping data in sync between ES 5 and ES 6.

config/elasticsearch_client.yml
MothOnMars: The ES client code behaves differently depending on whether `retry_on_failure` is `false` vs. an integer (even if that integer is zero), so I'd like to keep it as-is.

PR #387, opened 2019-10-17
Feature/system use notification

peggles2:
This pull request requires users to visit the system notification and accept the terms before proceeding to login.
- Add system notificaiton
- Update CSS to make the warning message more 508
- Add omniauth-rails_csrf_protection gem that will restrict a user doing a get request to /auth/login.gov
- Rspec and Cucumber tests.


app/views/user_sessions/security_notification.html.haml
```diff
@@ -0,0 +1,31 @@
...
+  %ul.nav.signup
+    %li= link_to 'Sign up', signup_path
+
+#user-session-form
+  %center
+    %h2 Security Notification
+    %h2 Search.gov is a service of GSA
+    %p for managing search experiences on federal government websites.
+    %p
+      %strong Members of the general public
```
MothOnMars: Remove extra space after "public". This appears on the page as `Members of the general public , please visit ...`.

app/views/user_sessions/security_notification.html.haml
```diff
@@ -0,0 +1,31 @@
+= content_for :title, 'Search.gov Login'
+= content_for :navigation do
+  %ul.nav.signup
+    %li= link_to 'Sign up', signup_path
```
MothOnMars: What is the purpose of including the "Sign up" link on this page? That link goes to the old signup form, which I assume we'll be removing. Won't the signup path and the login path eventually be the same, taking the user to this page?

app/views/user_sessions/security_notification.html.haml
```diff
@@ -0,0 +1,31 @@
...
+          no expectation of privacy is to be assumed. Individuals found performing unauthorized
+          activities are subject to disciplinary action including criminal prosecution.
+        *******************************************
+    = form_tag '/auth/logindotgov', method: 'post', id: 'user-session-form' do
+      = render_flash_message(false)
+      = submit_tag 'Accept and proceed'
+      = link_to 'Cancel and return to Search.gov', 'https://search.gov'
+      <br/>
+      = link_to 'Go to USA.gov', 'https://usa.gov'
+
```
MothOnMars: omit extra line

app/views/user_sessions/security_notification.html.haml
peggles2: So i'm following what we currently have.  If a user goes to the /login on production you will see the Sign up link there.  The path will eventually be redirected. But for now just focusing on getting the system notification out there.

MothOnMars: But won't it eventually be redirected to the same page? I'm wondering what the purpose is of adding it now if we're just going to remove it later.

app/views/user_sessions/security_notification.html.haml
peggles2: ah it was not an extra space after public. Was a haml syntax.  (one day...  I will change this all to erb)

app/views/user_sessions/security_notification.html.haml
peggles2: It was not added. This was the /login page and it is there currently on production. I'll focus on that when I get to that Jira ticket.

spec/support/omniauth-rails_csrf_protection.rb
```diff
@@ -0,0 +1,20 @@
+# Make sure that https://nvd.nist.gov/vuln/detail/CVE-2015-9284 is mitigated
+RSpec.describe 'CVE-2015-9284', type: :request do
```
MothOnMars: I just noticed a couple of other issues:
 - this spec should be moved from `spec/support/` to `spec/requests/`, and renamed to `omniauth-rails_csrf_protection_spec.rb` (note the addition of `_spec` at the end - without that, the spec won't be run)
- once it's in `spec/requests/`, you can remove `type: :request`, because that will be inferred from the spec directory

https://relishapp.com/rspec/rspec-rails/docs/request-specs/request-spec

MothOnMars:
LGTM 🌮

PR #388, opened 2019-10-22
SRCH-903 - clear session for user to logout

peggles2:
This pull request clears the session so the user can log out.

### DIFF
--- a/app/controllers/user_sessions_controller.rb
+++ b/app/controllers/user_sessions_controller.rb
@@ -1,9 +1,12 @@
 class UserSessionsController < ApplicationController
-  before_filter :reset_session, only: [:destroy]
-  before_filter :require_no_user, :only => [:new, :create]
-  before_filter :require_user, :only => :destroy
 
-  def security_notification; end
+  def new
+    construct_user_session
+  end
+
+  def security_notification
+    redirect_to(account_path) if current_user_session
+  end
 
   def create
     construct_user_session(user_session_params)
@@ -11,13 +14,13 @@ class UserSessionsController < ApplicationController
     if @user_session.save
       redirect_back_or_default redirection_path
     else
-      redirect_to(login_path)
+      render action: :new
     end
   end
 
   def destroy
     current_user_session.destroy
-    redirect_back_or_default login_url
+    redirect_to login_path
   end
 
   private
--- a/features/admin.feature
+++ b/features/admin.feature
@@ -2,8 +2,6 @@ Feature:  Administration
   Background:
     Given I am logged in with email "affiliate_admin@fixtures.org"
 
-  # to be updated in SRCH-903 for login.gov
-  @wip
   Scenario: Visiting the admin home page as an admin
     When I go to the admin home page
     Then I should see the browser page titled "Super Admin"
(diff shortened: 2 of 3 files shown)


app/controllers/omniauth_callbacks_controller.rb
```diff
@@ -6,7 +6,8 @@ def login_dot_gov
     @user = User.from_omniauth(auth)
     return unless @user.persisted?
 
-    UserSession.create(@user)
+    @user_session = UserSession.create(@user)
```
MothOnMars: This should not need to be assigned to an instance variable...?

app/controllers/user_sessions_controller.rb
```diff
@@ -1,9 +1,15 @@
...
-  def security_notification; end
+  def new
+    construct_user_session
+  end
+
+  def security_notification
+    redirect_to(account_path) if current_user
+  end
 
   def create
```
MothOnMars: `create` is one of several methods in this controller that do not appear to be in use anymore. We should purge those, but we should also ensure that the logic in those methods is implemented somewhere or captured in a story (if still needed). For example, `redirection_path` determines which page a user should land on after logging in, which is not currently implemented anywhere. We should review that and determine if it's still necessary. Also, the `construct_user_session`, which is no longer in use, included a line to secure the session:
```
@user_session.secure = Rails.application.config.ssl_options[:secure_cookies]
```
That seems like something we need to retain, as it ensures that the cookie will only be sent over SSL connections.

app/controllers/user_sessions_controller.rb
```diff
@@ -1,9 +1,15 @@
...
+  before_action :require_no_user, only: %w[new create]
+  before_action :require_user, only: :destroy
 
-  def security_notification; end
+  def new
+    construct_user_session
+  end
+
+  def security_notification
+    redirect_to(account_path) if current_user
```
MothOnMars: Specs?

app/controllers/user_sessions_controller.rb
peggles2: so I saw that too. But was afraid taking it out would break something.

app/controllers/user_sessions_controller.rb
peggles2: cucumber spec caught it.

app/controllers/user_sessions_controller.rb
peggles2: yeah we don't need to keep the redirection_path. Omniauth will return that. See https://github.com/omniauth/omniauth/wiki/Saving-User-Location.  There is already a jira ticket to ensure user continues to land on the page they were trying to visit after logging in. That would address this. 

I was going to not touch much of this unused code  but I'll try to remove lets see what happens if its going to be time suck to cleanup rspec tests that are tied to this, I will suggest we wait till after ATO.

app/controllers/user_sessions_controller.rb
peggles2: There is already a spec in user_session_controller. Also cucumber is testing it to redirect to the account_path.

peggles2:
@MothOnMars this is ready for review again

peggles2:
@MothOnMars this is ready for review. Please review this before reviewing any other new pull requests.

PR #393, opened 2019-10-23
Rails 5.1 deprecations

MothOnMars:
This PR resolves 3 low-hanging fruit deprecations:
- SRCH-631 resolve deprecation warning re. association reloading
- SRCH-644 replace before/after_filter with before/after_action
- SRCH-630 resolve deprecration warning re. using `distinct` vs `uniq`


MothOnMars:
@peggles2 , I just noticed that Code Climate identified a few issues. Please hold off on review until I've resolved those.

MothOnMars:
@peggles2 , this is ready for review.

app/controllers/concerns/query_routable_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+
+  def index; end
+
+  private
+
+  def handle_query_routing
+    return unless @search_options[:affiliate].present? && @search_options[:query].present?
+
+    if routed_query.present? &&
+       !matching_urls_for(routed_query.url).include?(request.referer)
```
peggles2: may want to look at using exclude instead of negating an include. It's easier to read and better practice.

app/controllers/concerns/query_routable_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+    u.scheme = u.scheme == 'http' ? 'https' : 'http'
+    [url, u.to_s]
+  end
+
+  def routed_query
+    @routed_query ||= @search_options[:affiliate].routed_queries.
+                        joins(:routed_query_keywords).
+                        where(
+                          routed_query_keywords: { keyword: @search_options[:query] }
+                        ).first
```
peggles2: This should be handled in a scope in the model routed_query.rb.

app/controllers/concerns/query_routable_controller.rb
MothOnMars: I’m trying to avoid extra refactoring of existing code. This file is not new code; it just appears that way because I moved the file. I cleaned up the rubocop issues, but we decided to skip extra refactoring of existing code for the Rails upgrade.

app/controllers/concerns/query_routable_controller.rb
MothOnMars: see above

PR #396, opened 2019-10-25
SRCH-975 addreesed hash deprecation warnings

PatKelsh:
SRCH-975 - This pull request fixes the following deprecation warnings: 
- each_value parameters
- inject deprecation warning
- symbolized_keys deprecation warning

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -113,7 +113,7 @@ class ApplicationController < ActionController::Base
   end
 
   def search_options_from_params(*param_keys)
-    h = permitted_params.slice(*param_keys)
+    h = permitted_params.to_h.slice(*param_keys)
     h.merge! affiliate: @affiliate,
              file_type: permitted_params[:filetype],
              page: permitted_params[:page],
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,8 +1,8 @@
 # frozen_string_literal: true
 
 class ApplicationController < ActionController::Base
-  before_filter :set_default_locale
-  after_filter :set_response_headers
+  before_action :set_default_locale
+  after_action :set_response_headers
   helper :all
   helper_method :current_user_session, :current_user, :permitted_params
   protect_from_forgery with: :exception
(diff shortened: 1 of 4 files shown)


app/controllers/application_controller.rb
```diff
@@ -1,8 +1,8 @@
 # frozen_string_literal: true
 
 class ApplicationController < ActionController::Base
-  before_filter :set_default_locale
-  after_filter :set_response_headers
+  before_action :set_default_locale
+  after_action :set_response_headers
```
MothOnMars: There's nothing wrong w/ this change, but don't worry about fixing other deprecations that you encounter while testing. We've batched the deprecations into tickets, so that each PR will address a specific type of deprecation to simplify code reviews.

lib/api/docs_search_options.rb
```diff
@@ -6,6 +6,7 @@ class Api::DocsSearchOptions < Api::CommercialSearchOptions
                         message: 'must be present'
 
   def initialize(params = {})
+    params = params.to_h unless params.is_a?(Hash)
```
MothOnMars: This is another case where it would be better to convert the params to a hash further up the code path. In this case, the params being passed to this initialization are assembled in Api::V2::SearchesController#search_params. Converting the params to a hash in that method will resolve the deprecation warning in this file, as well as any other potential downstream issues:
```ruby
#app/controllers/api/v2/searches_controller.rb

      def search_params
                                         :affiliate,
                                         :dc,
                                         :api_key,
                                         :cx,
                                         :enable_highlighting,
                                         :format,
                                         :limit,
                                         :offset,
                                         :query,
                                         :sort_by,
                                         :sc_access_key,
                                         :routed,
                                         :query_not,
                                         :query_quote,
                                         :query_or,
                                         :filetype,
                                         :filter
                                        ).to_h

      end
```

app/controllers/application_controller.rb
```diff
@@ -113,7 +113,7 @@ def permitted_params
   end
 
   def search_options_from_params(*param_keys)
-    h = permitted_params.slice(*param_keys)
+    h = permitted_params.to_h.slice(*param_keys)
```
MothOnMars: We can kill these two birds with one stone by converting the params to a hash in the `permitted_params` method itself:
```ruby
  def permitted_params
    @permitted_params ||= params.permit(*PERMITTED_PARAM_KEYS).to_h
  end
```

MothOnMars:
Ignore the review comments - moving these to https://github.com/GSA/search-gov/pull/397.

PR #397, opened 2019-10-25
SRCH 975 addressed hash deprecation warnings

PatKelsh:


MothOnMars:
@PatKelsh , this appears to be a duplicate of https://github.com/GSA/search-gov/pull/396

app/controllers/application_controller.rb
```diff
@@ -1,8 +1,8 @@
 # frozen_string_literal: true
 
 class ApplicationController < ActionController::Base
-  before_filter :set_default_locale
-  after_filter :set_response_headers
+  before_action :set_default_locale
+  after_action :set_response_headers
```
MothOnMars: There's nothing wrong w/ this change, but don't worry about fixing other deprecations that you encounter while testing. We've batched the deprecations into tickets, so that each PR will address a specific type of deprecation to simplify code reviews.

app/controllers/application_controller.rb
```diff
@@ -113,7 +113,7 @@ def permitted_params
   end
 
   def search_options_from_params(*param_keys)
-    h = permitted_params.slice(*param_keys)
+    h = permitted_params.to_h.slice(*param_keys)
```
MothOnMars: We can kill these two birds with one stone by converting the params to a hash in the `permitted_params` method itself:
```ruby
  def permitted_params
    @permitted_params ||= params.permit(*PERMITTED_PARAM_KEYS).to_h
  end
```

app/controllers/sites/rss_feeds_controller.rb
```diff
@@ -70,7 +70,7 @@ def assign_rss_feed_urls(attributes)
     existing_rss_feed_urls = []
     new_urls = []
 
-    rss_feed_urls_attributes = attributes || {}
+    rss_feed_urls_attributes = attributes.to_h || {}
```
MothOnMars: This fixes the issue on the line identified by the deprecation warning, but it would be preferable to convert the params to a hash in the method that extracts the params:
```ruby
  def rss_feed_params
    params.require(:rss_feed).permit(:name,
                                     :show_only_media_content,
                                     { rss_feed_urls_attributes: [:url] }).to_h
  end
```
That will ensure that the `rss_feed_params` are already a hash whenever they are used in the controller.

app/models/affiliate.rb
```diff
@@ -61,7 +61,7 @@ def make_unavailable(template_ids)
 
   has_many :users, -> { order 'contact_name' }, through: :memberships
   has_many :default_users, class_name: 'User', foreign_key: 'default_affiliate_id', dependent: :nullify
-  has_many :rss_feed_urls, -> { uniq }, through: :rss_feeds
+  has_many :rss_feed_urls, -> { distinct }, through: :rss_feeds
```
MothOnMars: See comment above re. ignoring other deprecations (but feel free to keep these changes as-is).

lib/api/docs_search_options.rb
```diff
@@ -6,6 +6,7 @@ class Api::DocsSearchOptions < Api::CommercialSearchOptions
                         message: 'must be present'
 
   def initialize(params = {})
+    params = params.to_h unless params.is_a?(Hash)
```
MothOnMars: This is another case where it would be better to convert the params to a hash further up the code path. In this case, the params being passed to this initialization are assembled in Api::V2::SearchesController#search_params. Converting the params to a hash in that method will resolve the deprecation warning in this file, as well as any other potential downstream issues:
```ruby
#app/controllers/api/v2/searches_controller.rb

      def search_params
                                         :affiliate,
                                         :dc,
                                         :api_key,
                                         :cx,
                                         :enable_highlighting,
                                         :format,
                                         :limit,
                                         :offset,
                                         :query,
                                         :sort_by,
                                         :sc_access_key,
                                         :routed,
                                         :query_not,
                                         :query_quote,
                                         :query_or,
                                         :filetype,
                                         :filter
                                        ).to_h

      end
```

app/controllers/application_controller.rb
```diff
@@ -130,7 +130,7 @@ def query_search_options
                                                  :'query-not',
                                                  :'query-or',
                                                  :'query-quote')
-    query_search_params.inject({}) do |hash, kv|
+    query_search_params.to_h.inject({}) do |hash, kv|
```
MothOnMars: This change is unnecessary, as `query_search_params` are based on `permitted_params`, which now returns a hash.

MothOnMars:
Looks good. Merging to the release branch.

PR #407, opened 2019-10-31
SRCH-973 updated shoulda-matcher gem to work with rails 5 and address…

PatKelsh:
… deprication warnings


MothOnMars:
As these changes are spec-only, I'm merging directly to master.

PR #409, opened 2019-11-01
SRCH 981 addressed deprecation warning for 'find'

PatKelsh:
The `find` search will only handle being passed in parameters but not full active record models. Fixed the tests that were triggering this warning


MothOnMars:
@PatKelsh , can you clean up this PR so that it only contains the commit for SRCH-981?

PR #414, opened 2019-11-04
SRCH-917 - When user is not persisted, authorized, redirect to error …

peggles2:
This pull request will redirect the user to the access denied error page if they are not persisted,  if their account was not approved, or they are not authorized to access the page.

### DIFF
--- a/app/controllers/omniauth_callbacks_controller.rb
+++ b/app/controllers/omniauth_callbacks_controller.rb
@@ -5,7 +5,7 @@ class OmniauthCallbacksController < ApplicationController
     @user = User.from_omniauth(request.env['omniauth.auth'])
 
     if @user.persisted? && @user.approval_status != 'not_approved'
-      #reset_session
+      reset_session
       set_user_session
 
       redirect_to(admin_home_page_path)
--- a/app/controllers/omniauth_callbacks_controller.rb
+++ b/app/controllers/omniauth_callbacks_controller.rb
@@ -5,7 +5,7 @@ class OmniauthCallbacksController < ApplicationController
     @user = User.from_omniauth(request.env['omniauth.auth'])
 
     if @user.persisted? && @user.approval_status != 'not_approved'
-      reset_session
+      #reset_session
       set_user_session
 
       redirect_to(admin_home_page_path)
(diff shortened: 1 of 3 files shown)


app/controllers/omniauth_callbacks_controller.rb
```diff
@@ -2,14 +2,16 @@
...
 class OmniauthCallbacksController < ApplicationController
   def login_dot_gov
-    auth = request.env['omniauth.auth']
-    @user = User.from_omniauth(auth)
-    return unless @user.persisted?
+    @user = User.from_omniauth(request.env['omniauth.auth'])
 
-    reset_session
-    set_user_session
+    if @user.persisted? && @user.approval_status != 'not_approved'
```
MothOnMars: As this is adding logic for a `not_approved` user, there should be a corresponding spec with the context `when a user is not_approved`.

spec/controllers/omniauth_callbacks_controller_spec.rb
```diff
@@ -73,8 +73,18 @@
...
-        expect { get_login_dot_gov }.to raise_error(ActionController::UnknownFormat)
+      it 'redirects to access-denied page' do
+        expect(get_login_dot_gov).to redirect_to('https://search.gov/access-denied')
+      end
+    end
+
+    context 'when a user is not approved' do
+      let(:email) { 'affiliate_manager_with_not_approved_status@fixtures.org' }
+      let(:user) { users(:affiliate_manager_with_not_approved_status) }
+      let(:auth) { mock_user_auth(user.email, email) }
```
MothOnMars: You could simplify this by omitting the `let(:email)` line (because `email` is equivalent to `user.email`), and using the default `uid` parameter in `mock_user_auth`:
```ruby
let(:auth) { mock_user_auth(user.email) }
```
Not a big deal, so I'm going to approve this and leave it up to you.

peggles2: oh yeah thats right. will do

peggles2: actually default uid has to be uniq.  and the default has already been created. I just named is not_approved12345

MothOnMars: 👍

PR #420, opened 2019-11-07
SRCH-627 delete all deprecation

PatKelsh:


MothOnMars:
Note: when merging, update commit message w/ correct ticket #, SRCH-627

lib/fast_delete_from_db_and_es.rb
```diff
@@ -8,7 +8,7 @@ def fast_delete(ids)
       return if ids.blank?
 
       "Elastic#{name}".constantize.delete(ids)
-      delete_all(id: ids)
+      where(:id => ids).delete_all unless (where(:id => ids) == nil)
```
MothOnMars: A couple of changes:
1) use Ruby 1.9 hash syntax: `where(id: ids).delete_all`
2) The `unless` clause isn't necessary, because if no records are found, then no records will be deleted.

spec/models/news_item_spec.rb
```diff
@@ -124,10 +124,10 @@
 
   describe '#fast_delete' do
     it 'delete from mysql and elasticsearch' do
-      ids = %w(1, 2).freeze
+      ids = [news_items(:item1).id, news_items(:item2).id].freeze
       expect(ElasticNewsItem).to receive(:delete).with(ids)
-      expect(NewsItem).to receive(:delete_all).with(id: ids)
       NewsItem.fast_delete(ids)
+      expect(NewsItem.where(:id => ids)).to be_empty
```
MothOnMars: use ruby 1.9 hash syntax: `where(id: ids)`

PR #423, opened 2019-11-08
SRCH-464 Update logstash queries to run on Elasticsearch 5.6

MothOnMars:
This is the first batch of changes to the Logstash queries. I batched the first set of changes as they involved some exploration and shared spec setup. The next set of changes will be more granular.   This PR:
- sets up some shared specs for logstash queries to ensure that they don't cause errors on the specified cluster
- configures the analytics specs to run against the 5.6 cluster
- updates the queries & specs with the correct query format for ES 5.6
- comments out the currently failing specs, which will be fixed & re-enabled in subsequent PRs

### DIFF
--- a/app/models/logstash_queries/top_n_modules_query.rb
+++ b/app/models/logstash_queries/top_n_modules_query.rb
@@ -4,6 +4,7 @@ class TopNModulesQuery < TopNQuery
   def booleans(json)
     json.must do
       json.child! { json.term { json.set! 'params.affiliate', @affiliate_name } }
+      modules_must(json)
       additional_musts(json)
     end
     json.must_not do
@@ -16,4 +17,6 @@ class TopNModulesQuery < TopNQuery
   def additional_musts(json); end
 
   def additional_must_nots(json); end
+
+  def modules_must(json); end
 end
--- a/app/models/logstash_queries/top_n_modules_query.rb
+++ b/app/models/logstash_queries/top_n_modules_query.rb
@@ -2,7 +2,7 @@
 
 class TopNModulesQuery < TopNQuery
   def booleans(json)
-    json.must do
+    json.filter do
       json.child! { json.term { json.set! 'params.affiliate', @affiliate_name } }
       modules_must(json)
       additional_musts(json)
--- a/app/models/logstash_queries/top_n_modules_query.rb
+++ b/app/models/logstash_queries/top_n_modules_query.rb
@@ -1,22 +1,19 @@
-class TopNModulesQuery < TopNQuery
+# frozen_string_literal: true
 
+class TopNModulesQuery < TopNQuery
   def booleans(json)
     json.must do
-      json.child! { json.term { json.affiliate @affiliate_name } }
-      modules_must(json)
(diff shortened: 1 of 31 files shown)


spec/models/logstash_queries/count_query_spec.rb
```diff
@@ -1,10 +1,25 @@
 require 'spec_helper'
 
-describe CountQuery, "#body" do
+describe CountQuery do
   let(:query) { CountQuery.new('affiliate_name') }
+  let(:expected_body) do
+    {
```
MothOnMars: Note: I'm using string keys here with double quotes to reflect the actual JSON we send to Elasticsearch. This also simplifies investigation and debugging, because queries in this format can be pasted directly into Kibana.

app/models/logstash_queries/top_n_modules_query.rb
```diff
@@ -1,22 +1,22 @@
...
+    json.filter do
+      json.child! { json.term { json.set! 'params.affiliate', @affiliate_name } }
       modules_must(json)
       additional_musts(json)
     end
     json.must_not do
-      json.child! { json.term { json.set! "useragent.device", "Spider" } }
-      json.child! { json.term { json.raw "" } }
+      json.child! { json.term { json.set! 'useragent.device', 'Spider' } }
+      json.child! { json.term { json.set! 'params.query.raw', '' } }
```
peggles2: @MothOnMars It is not recommended  to use .raw on params. It can open us up to security vulnerabilities.  sanitize is a safer way to go.

features/admin_center_watchers.feature
```diff
@@ -3,6 +3,8 @@ Feature: Watchers (aka Analytics Alerts)
   As a site customer
   I want to set up and manage various alerts
 
+  # SRCH-1038
+  @wip
   Scenario: View watchers
     Given the following Affiliates exist:
       | display_name | name       | contact_email   | contact_name |
```
peggles2: Just a suggestion add the name of your release branch will make it easy to search for all the commented out tests at the end.  As there will be other cucumber tests that are commented out for other reasons it would save you time to just look for the ones for your release.

app/models/logstash_queries/top_n_modules_query.rb
MothOnMars: The `params` in this line are not ActionController params. They refer to specific fields in an Elasticsearch document:
```
"params": {
            "query": "mars curiosity rover",
```

features/admin_center_watchers.feature
MothOnMars: That's the purpose of the ticket # in the comment.

peggles2: Yes but what I mean is, there will be other ticket numbers. So then you have to look up every ticket number to see if it's related to your release. As suppose to lookup all the ticket numbers for the release.  I'll explain in slack.

app/models/logstash_queries/top_n_modules_query.rb
peggles2: ok in the future best to rename that something different than params. Maybe query_params

PR #428, opened 2019-11-09
SRCH-1053 resolve deprecation related to association reloading

MothOnMars:
@peggles2 , this resolves a deprecation warning I missed in SRCH-631 as it only appeared in a cucumber test:
```
DEPRECATION WARNING: Passing an argument to force an association to reload is now deprecated and will be removed in Rails 5.1. Please call `reload` on the result collection proxy instead. (called from after_profile_deleted at /Users/Moth/src/GSA/search-gov-GSA/app/controllers/sites/youtube_profiles_controller.rb:15) 
```

### DIFF
--- a/app/controllers/sites/youtube_profiles_controller.rb
+++ b/app/controllers/sites/youtube_profiles_controller.rb
@@ -12,7 +12,7 @@ class Sites::YoutubeProfilesController < Sites::SetupSiteController
   end
 
   def after_profile_deleted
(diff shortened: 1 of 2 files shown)


MothOnMars:
Closing to resolve cucumber failure.

MothOnMars:
Re-opening. The cucumber failure was an unrelated transient failure.

PR #429, opened 2019-11-11
SRCH-1023 SRCH-1042 update Sparkline queries for Elasticsearch 5.6

MothOnMars:
SRCH-1023 update OverallSparklineQuery for Elasticsearch 5.6
SRCH-1042 update ModuleSparkline Query

### DIFF
--- a/app/models/logstash_queries/module_sparkline_query.rb
+++ b/app/models/logstash_queries/module_sparkline_query.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 class ModuleSparklineQuery
   include AnalyticsDSL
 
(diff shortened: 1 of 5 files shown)


app/models/logstash_queries/module_sparkline_query.rb
```diff
@@ -26,7 +28,7 @@ def top_level_histogram_agg(json)
       json.agg do
         json.terms do
           json.field 'modules'
-          json.size 0
+          json.size 100
```
MothOnMars: `"size": 0` used to specify that the aggregations include all possible buckets. That was [removed](https://github.com/elastic/elasticsearch/issues/18838), so aggregations must now specify a `size` as the maximum # of buckets to return. This particular query aggregates on the `modules` field. We currently have about 20 modules, so an upward limit of 100 should be plenty even if we add some additional modules later.

MothOnMars:
Closing to make one more change.

PR #436, opened 2019-11-12
Feature/welcome email

peggles2:
SRCH-953 -Update welcome email so if the user is blank use the users email address.
Update the non .gov users if they are approved the flash message should not show up anymore. I also added a cucumber test to test that.

### DIFF
--- a/features/users.feature
+++ b/features/users.feature
@@ -79,6 +79,16 @@ Feature: Users
     Then I should be on the user account page
     And I should see "Because you don't have a .gov or .mil email address, we need additional information."
 
+  @javascript
+  Scenario: Logging in as a new approved affiliate user without government affiliated email address
+    Given the following Users exist:
+      | contact_name | email             | approval_status |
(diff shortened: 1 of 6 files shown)


features/users.feature
```diff
@@ -79,6 +79,21 @@ Feature: Users
...
+      | Joe Schmo    | jschmo@random.com | approved        |
+
+    And I am logged in with email "jschmo@random.com"
+    Then I should be on the user account page
+    And I should not see "Because you don't have a .gov or .mil email address, we need additional information."
+
+  Scenario: Failing registration as a new affiliate user
+    Given I am on the sign up page
+    And I press "Sign up"
+    Then I should see "Email can't be blank"
```
MothOnMars: This doesn't seem like a valid scenario, because users no longer have to fill in an email address to sign up.

peggles2: Yeah thats removed in previous pr. It was not merged with this PR let me pull it out.

PR #439, opened 2019-11-12
Feature/config updates

peggles2:


config/initializers/omniauth.rb
```diff
@@ -1,10 +1,10 @@
 Rails.application.config.middleware.use OmniAuth::Builder do
   provider :login_dot_gov, {
     name: :logindotgov,
-    client_id: 'urn:gov:gsa:openidconnect.profiles:sp:sso:gsa:search',
-    idp_base_url: 'https://idp.int.identitysandbox.gov',
+    client_id: Rails.application.secrets.login_dot_gov['client_id'],
```
MothOnMars: You could DRY this file up by loading the secrets once into a hash:
```ruby
def login_dot_gov_config
  Rails.application.secrets.login_dot_gov.freeze
end

Rails.application.config.middleware.use OmniAuth::Builder do
  provider :login_dot_gov, {
    name: :logindotgov,
    client_id: login_dot_gov_config['client_id'],
...etc
```
I'll approve, so you can take that suggestion or leave it.

Also, if you don't want to have to update this again for Rails 5.1, which requires symbol keys for secrets, you could use:
```ruby
def login_dot_gov_config
  Rails.application.secrets.login_dot_gov.with_indifferent_access.freeze
end
```
That would allow it to work with string or symbol keys.

MothOnMars: Well, never mind, then:
```
2.1 Top-level HashWithIndifferentAccess is soft-deprecated
If your application uses the the top-level HashWithIndifferentAccess class, you should slowly move your code to instead use ActiveSupport::HashWithIndifferentAccess.

It is only soft-deprecated, which means that your code will not break at the moment and no deprecation warning will be displayed, but this constant will be removed in the future.
```
Let's just update to symbol keys as needed.

peggles2: That is in Rails 5.1.  The keys have to be string for now.

MothOnMars: A `HashWithIndifferentAccess` allows access with either string or symbol keys, so it's backward & forward compatible. Something we can keep in mind on the off-chance we need to set up more secrets before we hit 5.1.

PR #441, opened 2019-11-14
Update Rails to 5.1

MothOnMars:
This PR:
- upgrades Rails to 5.1
- updates and adds the necessary configuration files (based on `rails app:update`)
- updates the keys to the secrets from strings to symbols
- resolves a couple of remaining deprecations
- skips the remaining failing specs


MothOnMars:
@peggles2 , I added one more commit to remove the `thin` gem, which is no longer supported, and now unneeded with the addition of `puma`: 
https://github.com/GSA/search-gov/pull/441/commits/894408ba3dc5d7b819a3a15ca3901abf1cc49026

PR #445, opened 2019-11-14
SRCH-1039 SRCH-1040 update DateRangeTopN queries to run on ES 5.6

MothOnMars:
- update DateRangeTopNMissingQuery
- update DateRangeTopNFieldQuery
- replace Timecop with ActiveSupport::Testing::TimeHelpers

### DIFF
--- a/app/models/logstash_queries/date_range_top_n_field_query.rb
+++ b/app/models/logstash_queries/date_range_top_n_field_query.rb
@@ -3,17 +3,18 @@ class DateRangeTopNFieldQuery < DateRangeTopNQuery
 
   def initialize(affiliate_name, start_date, end_date, filter_field, filter_value, agg_options = {})
     super(affiliate_name, start_date, end_date, agg_options)
-    @filters = { 'affiliate' => @affiliate_name, filter_field => filter_value }.compact
+    @filters = { 'params.affiliate' => @affiliate_name,
+                 filter_field => filter_value }.compact
   end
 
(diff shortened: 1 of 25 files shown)


app/models/logstash_queries/date_range_top_n_field_query.rb
```diff
@@ -3,17 +3,18 @@ class DateRangeTopNFieldQuery < DateRangeTopNQuery
 
   def initialize(affiliate_name, start_date, end_date, filter_field, filter_value, agg_options = {})
     super(affiliate_name, start_date, end_date, agg_options)
-    @filters = { 'affiliate' => @affiliate_name, filter_field => filter_value }.compact
+    @filters = { 'params.affiliate' => @affiliate_name,
+                 filter_field => filter_value }.compact
```
peggles2: since you are already factoring this you could also replace the hash rockets. But I will leave it up to you since it was already that way.

MothOnMars: As `params.affiliate` is a string, and `filter_field` is a variable pointing to a string, the recommended style is to use hash rockets:
https://rubystyle.guide/#no-mixed-hash-syntaces

peggles2: It still can be done :)
 { "params.affiliate": @affiliate_name,
   "#{filter_filed}": filter_value }

PR #447, opened 2019-11-14
SRCH-1062 - add omniauth protocol

peggles2:
SRCH-1062 - update all non dev servers to use https: the rest http.

### DIFF
--- a/config/initializers/omniauth.rb
+++ b/config/initializers/omniauth.rb
@@ -1,5 +1,6 @@
 Rails.application.config.middleware.use OmniAuth::Builder do
-  protocol = ENV['RAILS_ENV'] == 'development' ? 'http://' : 'https://'
+  protocol = Rails.env.development? ? 'http://' : 'https://'
+
   provider :login_dot_gov, {
     name: :logindotgov,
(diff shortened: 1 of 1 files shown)


config/initializers/omniauth.rb
```diff
@@ -1,10 +1,11 @@
 Rails.application.config.middleware.use OmniAuth::Builder do
+  protocol = ENV['RAILS_ENV'] == 'development' ? 'http://' : 'https://'
```
MothOnMars: You could simplify this with:
```
protocol = Rails.env.development? ? 'http://' : 'https://'
```

It might also be clearer if the `protocol` was simply `http` or `https` rather than the protocol followed by the colon & trailing slashes. `http` is a `protocol`, while `http://` is a protocol followed by some punctuation. 

As those are just stylistic niggles, I'll approve and leave to your discretion.

peggles2: its more dry to include the colon and trailing slashes thats why I did it.
But I'll add this protocol = Rails.env.development? ? 'http://' : 'https://'

PR #453, opened 2019-11-18
Feature/update blanks in email

peggles2:
This pull request does the following

Fixes emails so if the user name is not defined it uses their email address
Makes the email message more gender neutral
Updates rspec tests that testes the message so that it reflects the gender neutral change. 

**Deployment note:**  In rails console must run EmailTemplate.load_default_templates


peggles2:
> LGTM. Note, however, that the change to the .erb file ONLY affects dev & test environments; you'll need to update that template in the production database in tandem with this code change.

Yeah I discovered this on the weekend when I was trying to test it on my local laptop

peggles2:
@MothOnMars found some more templates that needed to be updated.

app/mailers/emailer.rb
```diff
@@ -31,13 +38,14 @@ def new_feature_adoption_to_admin
...
     @email_verification_url = email_verification_url(user.email_verification_token)
-    @user_contact_name = user.contact_name
+    @user_contact_name = user.contact_name.presence || user.email
     @user_email = user.email
     generic_user_html_email(user, __method__)
   end
 
   def user_approval_removed(user)
     @user = user
+    @user_contact_name = user.contact_name.presence || user.email
```
MothOnMars: These could be DRY'd up by adding a `contact_name(user)` method to this class (or even to the `User` class, if we would need it elsewhere). That said, these will no longer be needed once we require contact names, so it's up to you.

app/mailers/emailer.rb
```diff
@@ -31,13 +38,14 @@ def new_feature_adoption_to_admin
 
   def user_email_verification(user)
```
MothOnMars: Can't this email and this method be removed entirely?

app/mailers/emailer.rb
peggles2: I agree and would have done that. But just as we had discussed, right going to just follow the current pattern so we can get this fix. Cleanup will have to happen later.

app/mailers/emailer.rb
peggles2: Thats being handled in SRCH-1076. I am focusing on the requirements for this ticket.

MothOnMars:
LGTM 🌮

PR #456, opened 2019-11-18
SRCH-1082 - Knollfear/papertrail

knollfear:
This PR adds papertrail to the project.  It is implemented only on the users model at this time, but it can be easily added to other models.  This PR contains a migration script so remember to run that script when testing locally.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -109,6 +109,7 @@ gem 'jquery-rails', git: 'https://github.com/GSA/jquery-rails', ref: 'c0c56208f5
 gem 'therubyracer', '~> 0.12.3'
 gem 'yui-compressor', '~> 0.12.0'
 gem 'twitter-typeahead-rails', '~> 0.11.1'
+gem 'paper_trail'
 # Why do we have two versions of Font Awesome?
 # One is for general use around the app, in places where
 # web fonts are expected to work...
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -688,4 +688,28 @@ describe User do
       it { is_expected.to eq false }
     end
   end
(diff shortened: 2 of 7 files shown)


Gemfile
```diff
@@ -109,6 +109,7 @@ gem 'jquery-rails', git: 'https://github.com/GSA/jquery-rails', ref: 'c0c56208f5
 gem 'therubyracer', '~> 0.12.3'
 gem 'yui-compressor', '~> 0.12.0'
 gem 'twitter-typeahead-rails', '~> 0.11.1'
+gem 'paper_trail'
```
peggles2: include the version of paper trail here as well. `gem 'paper_trail', '~>10.3.1`

spec/models/user_spec.rb
```diff
@@ -687,4 +687,28 @@
...
+      end
+      it 'check version count' do
+        expect(user.versions.size).to eq(3)
+      end
+      it 'check updating increases version count' do
+        user.update!(contact_name: 'Jim Four', password: 'JimFour44!')
+        expect(user.versions.size).to eq(4)
+      end
+    end
+  end
```
peggles2: It be cleaner to make this into a helper. You can call it papertrail_helper.rb and place it in the helper directory. The user_spec.rb will auto load this.

peggles2:
Also we need to create the migration as a separate pull request and deploy that first before pushing the code.

spec/models/user_spec.rb
peggles2: This is actually correct. I thought this was a different file. so ignore.

peggles2:
Closing this for now since its not an ATO requirement

PR #459, opened 2019-11-19
Knollfear/containerization

knollfear:
This adds a docker-compose file for the test dependencies.  With this change there will be no need to install ElasticSearch or run a MYSQL db locally.  Instead users can simply bring up the necessary containers with docker-compose.  This should speed up the time it takes for new people to successfully run the tests and have a functional workspace.


peggles2:
closing for now and will prioritize after ATO

PR #495, opened 2019-12-09
SRCH-1102 validate SiteFeedUrl#rss_url

MothOnMars:
- validate that SiteFeedUrl#rss_url is a valid URL
- use `URI.parse(url).open` instead of `Kernel.open`


spec/models/site_feed_url_spec.rb
```diff
@@ -4,9 +4,14 @@
...
-  describe "Creating new instance" do
+  describe 'associations' do
     it { is_expected.to belong_to :affiliate }
+  end
+
+  describe 'validations' do
     it { is_expected.to validate_presence_of :rss_url }
+    it { is_expected.to allow_value('http://some.site.gov/url').for(:rss_url) }
+    it { is_expected.not_to allow_value('not a URL').for(:rss_url) }
   end
```
peggles2: can you also test inline javascript in the url as well?

MothOnMars:
Closing to split this PR in two.

PR #507, opened 2019-12-30
Feature/inverse

peggles2:
SRCH-615 - Ensure bi-directional ActiveRecord associations have an "inverse_of" association

### DIFF
--- a/app/models/affiliate.rb
+++ b/app/models/affiliate.rb
@@ -59,7 +59,9 @@ class Affiliate < ApplicationRecord
 
   has_many :available_templates, through: :affiliate_templates, source: :template
   has_many :users, -> { order 'contact_name' }, through: :memberships
-  has_many :default_users, class_name: 'User', foreign_key: 'default_affiliate_id', dependent: :nullify
+  has_many :default_users, class_name: 'User',
+           foreign_key: 'default_affiliate_id', dependent: :nullify,
+           inverse_of: :affiliate
   has_many :rss_feed_urls, -> { distinct }, through: :rss_feeds
   has_many :url_prefixes, :through => :document_collections
   has_many :twitter_profiles, -> { order 'twitter_profiles.screen_name ASC' }, through: :affiliate_twitter_settings
@@ -68,7 +70,7 @@ class Affiliate < ApplicationRecord
   has_many :i14y_drawers, -> { order 'handle' }, through: :i14y_memberships
   has_many :routed_query_keywords, -> { order 'keyword' }, through: :routed_queries
   belongs_to :agency
-  belongs_to :language, foreign_key: :locale, primary_key: :code
+  belongs_to :language, foreign_key: :locale, primary_key: :code, inverse_of: :affiliate
   belongs_to :template, inverse_of: :affiliates
 
   AWS_IMAGE_SETTINGS = { styles: { :large => "300x150>" },
--- a/app/models/affiliate.rb
+++ b/app/models/affiliate.rb
@@ -19,31 +19,42 @@ class Affiliate < ApplicationRecord
   MAX_NAME_LENGTH = 33.freeze
 
   with_options dependent: :destroy do |assoc|
-    assoc.has_many :affiliate_feature_addition
-    assoc.has_many :affiliate_twitter_settings
-    assoc.has_many :boosted_contents
-    assoc.has_many :connections, -> { order 'connections.position ASC' }
-    assoc.has_many :connected_connections, :foreign_key => :connected_affiliate_id, :source => :connections, :class_name => 'Connection'
-    assoc.has_many :document_collections, -> { order 'document_collections.name ASC, document_collections.id ASC' }
-    assoc.has_many :excluded_domains, -> { order 'domain ASC' }
-    assoc.has_many :excluded_urls
-    assoc.has_many :featured_collections
-    assoc.has_many :features, :through => :affiliate_feature_addition
-    assoc.has_many :flickr_profiles, -> { order 'flickr_profiles.url ASC' }
-    assoc.has_many :i14y_memberships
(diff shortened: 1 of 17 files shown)


app/models/affiliate.rb
```diff
@@ -19,31 +19,46 @@ class Affiliate < ApplicationRecord
...
-    assoc.has_many :routed_queries
-    assoc.has_many :rss_feeds, -> { order 'rss_feeds.name ASC, rss_feeds.id ASC' }, as: :owner
-    assoc.has_many :sayt_suggestions
-    assoc.has_many :site_domains, -> { order 'domain ASC' }, inverse_of: :affiliate
-    assoc.has_one :site_feed_url
-    assoc.has_many :superfresh_urls
-    assoc.has_one :alert
-    assoc.has_many :watchers, -> { order 'name ASC' }
-    assoc.has_many :tag_filters, -> { order 'tag ASC' }
+    assoc.has_many(:affiliate_feature_addition)
```
MothOnMars: As `has_many` is part of the internal Rails DSL, parentheses should not be used around the args:
https://github.com/rubocop-hq/ruby-style-guide#methods-that-are-part-of-an-internal-dsl

app/models/affiliate.rb
```diff
@@ -19,31 +19,46 @@ class Affiliate < ApplicationRecord
...
+    assoc.has_many(:affiliate_twitter_settings)
+    assoc.has_many(:boosted_contents)
+    assoc.has_many(:connections, -> { order 'connections.position ASC' },
+                   inverse_of: :affiliate)
+    assoc.has_many(:connected_connections,
+                   foreign_key: :connected_affiliate_id,
+                   source: :connections,
+                   class_name: 'Connection',
+                   inverse_of: :affiliate)
```
MothOnMars: This should be `inverse_of: :connected_affiliate`

spec/models/affiliate_spec.rb
```diff
@@ -76,35 +76,60 @@
     it { is_expected.to validate_inclusion_of(:search_engine).in_array(%w( Google BingV6 BingV7 SearchGov )) }
 
     it { is_expected.to have_many :boosted_contents }
+    it { is_expected.to have_many(:connections).inverse_of(:affiliate) }
+    it { is_expected.to have_many(:connected_connections).inverse_of(:affiliate) }
```
MothOnMars: Should be `inverse_of(:connected_affiliate)`

app/models/affiliate.rb
```diff
@@ -19,31 +19,46 @@ class Affiliate < ApplicationRecord
...
+    assoc.has_many(:i14y_memberships)
+    assoc.has_one(:image_search_label)
+    assoc.has_many(:indexed_documents)
+    assoc.has_many(:memberships)
+    assoc.has_many(:navigations,
+                   -> { order 'navigations.position ASC, navigations.id ASC' },
+                   inverse_of: :affiliate)
+    assoc.has_many(:routed_queries)
+    assoc.has_many(:rss_feeds, -> { order 'rss_feeds.name ASC, rss_feeds.id ASC' },
```
MothOnMars: This needs a spec for the `inverse_of`

app/models/affiliate.rb
```diff
@@ -19,31 +19,46 @@ class Affiliate < ApplicationRecord
...
+                   as: :owner,
+                   inverse_of: :owner)
+    assoc.has_many(:sayt_suggestions)
+    assoc.has_many(:site_domains, -> { order 'domain ASC' }, inverse_of: :affiliate)
+    assoc.has_one(:site_feed_url)
+    assoc.has_many(:superfresh_urls)
+    assoc.has_one(:alert)
+    assoc.has_many(:watchers, -> { order 'name ASC' }, inverse_of: :affiliate)
+    assoc.has_many(:tag_filters, -> { order 'tag ASC' }, inverse_of: :affiliate)
```
MothOnMars: needs a spec for the `inverse_of`

app/models/affiliate.rb
peggles2: ahh right. rubcop was complaining about them.

MothOnMars: We could add the Rails association methods to the list of ignored methods for that cop:
```
Style/MethodCallWithArgsParentheses:
  Description: 'Use parentheses for method calls with arguments.'
...
  IgnoredMethods: ['require']
```
We could also consider disabling that cop entirely. It's disabled by default because "There are just too many false positives at the moment". See https://github.com/rubocop-hq/rubocop/issues/6564#issuecomment-445700075

app/models/affiliate.rb
```diff
@@ -19,31 +19,46 @@ class Affiliate < ApplicationRecord
...
+    assoc.has_one(:image_search_label)
+    assoc.has_many(:indexed_documents)
+    assoc.has_many(:memberships)
+    assoc.has_many(:navigations,
+                   -> { order 'navigations.position ASC, navigations.id ASC' },
+                   inverse_of: :affiliate)
+    assoc.has_many(:routed_queries)
+    assoc.has_many(:rss_feeds, -> { order 'rss_feeds.name ASC, rss_feeds.id ASC' },
+                   as: :owner,
+                   inverse_of: :owner)
```
peggles2: It has an inverse of. Its polymorphic association so the inverse is owner.

app/models/affiliate.rb
peggles2: I think disabling it may be a good idea. And just put parenthesis as needed.  I'll do that now.

spec/models/affiliate_spec.rb
peggles2: I'm guessing you mean the test line below that which is testing  ` it { is_expected.to have_many(:connected_connections).inverse_of(:connected_affiliate) }`

PR #508, opened 2020-01-02
SRCH-1165 - Youtube rake tasks should run successfully

peggles2:
This pull request removes uniq and places it with distinct as uniq is deprecated in rails 5.0 and was removed in rails 5.1.

### DIFF
--- a/spec/fixtures/affiliates.yml
+++ b/spec/fixtures/affiliates.yml
@@ -45,6 +45,7 @@ another_affiliate:
   display_name: Another Gov Site
   name: another.gov
   api_access_key: another_key
+  youtube_profiles: nps
 
 admin_affiliate:
   <<: *DEFAULTS
--- a/spec/fixtures/youtube_profiles.yml
+++ b/spec/fixtures/youtube_profiles.yml
@@ -5,3 +5,7 @@ whitehouse:
 nps:
   channel_id: natlparkservice_channel_id
   title: natlparkservice
+
+no_affiliate:
+  channel_id: noaffiliate_channel_id
+  title: noaffiliate_title
--- a/spec/models/youtube_profile_spec.rb
+++ b/spec/models/youtube_profile_spec.rb
@@ -7,8 +7,28 @@ describe YoutubeProfile do
   it { is_expected.to validate_presence_of :channel_id }
   it { is_expected.to have_one(:rss_feed).dependent :destroy }
   it { is_expected.to have_and_belong_to_many :affiliates }
-  it { is_expected.to validate_uniqueness_of(:channel_id).
-                with_message(/has already been added/) }
+  it do
+    is_expected.to validate_uniqueness_of(:channel_id).
+      with_message(/has already been added/)
+  end
+
+  it do
+    is_expected.to have_many(:youtube_playlists).
+                   dependent(:destroy).
+                   inverse_of(:youtube_profile)
+  end
+
+  describe 'Gets the active YoutubeProfiles' do
+    let (:profiles) { YoutubeProfile.active }
+
+    it 'gets the active youtube profiles' do
+      expect(YoutubeProfile.active.count).to equal(2)
+    end
+
+    it 'is expected to have uniq values' do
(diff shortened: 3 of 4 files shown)


spec/fixtures/affiliates.yml
```diff
@@ -45,6 +45,7 @@ another_affiliate:
   display_name: Another Gov Site
   name: another.gov
   api_access_key: another_key
+  youtube_profiles: nps
```
peggles2: needed nps youtube to have 2 affiliates to test out .distinct (replacement for .uniq)

spec/fixtures/youtube_profiles.yml
```diff
@@ -5,3 +5,7 @@ whitehouse:
 nps:
   channel_id: natlparkservice_channel_id
   title: natlparkservice
+
+no_affiliate:
+  channel_id: noaffiliate_channel_id
+  title: noaffiliate_title
```
peggles2: Need to have a youtube_profile that does not have an affiliate to test the .active scopes out correctly.

spec/models/youtube_profile_spec.rb
```diff
@@ -7,8 +7,23 @@
...
+    let (:profiles) { YoutubeProfile.active }
+
+    it 'gets the active youtube profiles' do
+      expect(YoutubeProfile.active.count).to equal(2)
+    end
+
+    it 'is expected to have uniq values' do
+      expect(YoutubeProfile.active.count).to equal(YoutubeProfile.active.distinct.count)
+    end
+  end
```
MothOnMars: A few spec changes:
- use the scope method name as the description: `describe '.active' do`
- It would be preferable to explicitly create a duplicate profile to confirm the uniqueness, rather than rely on a fixture. Example:
```ruby
    context 'when the same profile belongs to multiple affiliates' do
      let(:profile) { youtube_profiles(:nps) }

      before do
        profile.affiliates << affiliates(:another_affiliate)
      end

      it 'returns unique Youtube channels' ...etc
```
That adds some extra documentation and helps keep our fixtures free of data required for a single spec.
- Add a separate spec for the behavior `it 'does not include profiles that do not belong to an affiliate'`. No spec currently tests the omission of the `noaffiliate_channel_id` profile; the specs will still pass if that fixture does not exist. Example:
```ruby
    it 'does not include profiles that do not belong to an affiliate' do
      expect(YoutubeProfile.active).not_to include(youtube_profiles(:no_affiliate))
    end
```
  

- add a spec that verifies the bugfix. `YoutubeProfile.active` on its own works fine in production. The failure only arises when that scope is combined with the `stale` scope:
```ruby
irb(main):022:0> YoutubeProfile.active.count
=> 898
irb(main):023:0> YoutubeProfile.active.stale.count
Traceback (most recent call last):
        1: from (irb):23
NoMethodError (undefined method `stale' for #<Array:0x0000556a5cbef188>)
```
Apparently the `uniq` is forcing the scope to return an array rather than a relation, which prevents it from being combined with other scopes. The simplest way to confirm the desired behavior is by verifying the class returned by the scope: 
```ruby
    it 'returns a relation' do
      expect(YoutubeProfile.active).to be_an(ActiveRecord::Relation)
    end
```

peggles2: Currently the bug is that it is using uniq rather than distinct. I agree this code does need a lot more rspec tests than it currently has. (did not have any for the scopes). But due to limited time, and other higher priority tickets, I am only adding minimum rspec test just to test out using distinct instead of uniq bug fix. We can create a ticket to address additional tests later.

peggles2: Also uniq is deprecated and removed in rails 5.0 so the test of using distinct is in this rspec test.

peggles2: https://riptutorial.com/ruby-on-rails/example/11652/-distinct--or--uniq-

peggles2: https://blog.driftingruby.com/ruby-on-rails-5-1-0-deprecations/

PR #509, opened 2020-01-02
replace Kernel.open which lets data spawn shells with URI.parse to pr…

cryptofilegsa:
…otect against RCE

avoiding remote code execution possibilities, remove Kernel.open which lets data spawn shells, does URI.parse handle the image url as expected?


MothOnMars:
This was not a production vulnerability, as that code is only run during cucumber tests, but it's definitely a good thing to clean up worst-practice code from anywhere in our code base. Thanks for the PR, @cryptofilegsa! Merging.

PR #516, opened 2020-01-15
SRCH-1191 - create not_active_for scope that passes time parameter

peggles2:
This pull request is to create a scope in the User model to collect all users that have been not active for a given number of days. 

For example if you pass the number 76 you can get back all the users that were not active for exactly 76 days.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -39,6 +39,16 @@ class User < ApplicationRecord
                 90.days.ago)
         }
 
+  scope :not_active_for,
+        lambda { |days|
+          where('current_login_at >= ? AND current_login_at <= ? OR
+                (current_login_at IS NULL AND created_at >= ? AND created_at <=? )',
+                days.beginning_of_day,
+                days.end_of_day,
+                days.beginning_of_day,
+                days.end_of_day)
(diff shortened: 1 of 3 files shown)


peggles2:
@MothOnMars  this is ready for review

app/models/user.rb
```diff
@@ -39,6 +39,16 @@ class User < ApplicationRecord
                 90.days.ago)
         }
 
+  scope :not_active_for,
```
MothOnMars: Per our discussion rename the method and/or args to read more logically

app/models/user.rb
```diff
@@ -39,6 +39,16 @@ class User < ApplicationRecord
...
+        lambda { |days|
```
MothOnMars: The SQL scope could be simplified by passing the method an instance of `Date` rather than time (i.e. `76.days.ago.to_date`) and using MySQL's `DATE`:
```ruby
where('DATE(current_login_at) = ? OR
  (current_login_at IS NULL AND DATE(created_at) = ?)', date, date)
```

peggles2:
@MothOnMars this is ready for review

PR #519, opened 2020-01-22
Feature/email template deactivated account

peggles2:
Add email template for warning users that their account is going to be de-activated.


spec/mailers/emailer_spec.rb
```diff
@@ -6,6 +6,19 @@
   include EmailSpec::Matchers
   fixtures :affiliates, :users, :features, :memberships
 
+  describe '#account_deactivation_warning' do
```
MothOnMars: Class methods should be prepended with a period: `describe '.account_deactivation_warning'` (I see the other class methods in this spec are prepended with a `#`, but that ain't right)
http://www.betterspecs.org/#describe

spec/mailers/emailer_spec.rb
```diff
@@ -6,6 +6,19 @@
...
+    let(:user) { users(:not_active_76_days) }
+    let(:message) do
+      'automatically deactivate your account in the next 14 days due to inactivity.'
+    end
+
+    subject(:deactivate_email) { Emailer.account_deactivation_warning(user, 76.days.ago.to_date) }
```
MothOnMars: When describing a method, it is best to use a subject name consistent with the method being tested:
```ruby
subject(:account_deactivation_warning) do
  Emailer.account_deactivation_warning(user, 76.days.ago.to_date)
end
```

spec/mailers/emailer_spec.rb
peggles2: ok I was trying to just be consistent with what is currently there.

MothOnMars:
LGTM

PR #527, opened 2020-01-29
Feature/notify user inactivity rake task

peggles2:
This pull request creates a rake task to send out the warning that a user is going to be de-activated.

Users who have been inactive for more than 90 days, have their account de-activated. The purpose of this is to notify users in advanced that their account will be de-activated.  The code to do this is now on production. The final piece was this rake task that will call the method that sends email out to the user to warn them. 

A date is passed to the rake task which will be used to collect all users inactive since that given date and sends an email out.

There is also another existing rake task that deactivates users that have been inactive for 90 days. That rake task has been updated to notify users that their account has been de-activated. 

Deployment Notes:
no migrations
2 cron jobs must be setup for  warn_set_to_not_approved
usasearch:user:warn_set_to_not_approved[date -d '76 days ago']
usasearch:user:warn_set_to_not_approved[date -d '86 days ago']


spec/lib/tasks/user_spec.rb
```diff
@@ -87,4 +88,24 @@
       @rake[task_name].invoke
     end
   end
+
+  describe 'usasearch:user:warn_set_to_not_approved' do
```
PatKelsh: wouldn't be a bad idea to include a test to make sure the warning is not sent prematurely. Maybe a similar test with a date of 75 days ago and using `expect(...)not.to`

PR #530, opened 2020-02-05
Feature/require account fields

peggles2:
This pull request is to require contact_name and organization_name for the user editing their account.

Contact name and and organization name were required before.  After we migrated to login.gov for authentication we could not require these fields on the creation of the user because login.gov will only pass the UID and email of the user. It does not forward any other information.  So the registration process has to be broken down into 2 steps.  The user will have to create an account through login.gov, login to their account, their account gets created, then user is directed to updated their account information. 

This is the first step. Next pull requests will handle redirecting the user to fill in their contact name and organization name after signing up. 

To validate the user during update and not during create, I followed this article: https://www.justinweiss.com/articles/a-lightweight-way-to-handle-different-validation-situations/

In the controller a new method called update_account handles users updating their account. 
The contact name and organization names will be required.

Any updates to the user that are done by rake tasks, cron jobs are not impacted by this change.


app/controllers/users_controller.rb
```diff
@@ -1,6 +1,7 @@
 class UsersController < ApplicationController
   layout 'sites'
   before_action :require_user, :only => [:show, :edit, :update]
+  before_action :set_user, except: :create
```
peggles2: added this to DRY up the code (Do not Repeat Yourself)

app/mailers/emailer.rb
```diff
@@ -80,7 +80,6 @@ def welcome_to_new_user_added_by_affiliate(affiliate, user, current_user)
     @added_user_email = user.email
     @affiliate_display_name = affiliate.display_name
     @affiliate_site_url = site_url(affiliate)
-    @complete_registration_url = site_url(affiliate)
```
peggles2: we are not completing registrations anymore. This was left over from when a user account had to be validated. Login.gov handles this part.

features/users.feature
```diff
@@ -16,53 +16,28 @@ Feature: Users
     Then I should see "Security Notification"
     And I should see "This is a U.S. General Services Administration Federal Government computer system"
 
-  # to be updated in SRCH-862 for login.gov
+  # to be updated in SRCH-1237 for login.gov
   @wip
   @javascript
   Scenario: Registering as a new affiliate user who is a government employee or contractor with .gov email address
```
peggles2: keeping this commented out for now but I did cleanup some of the test.

spec/fixtures/users.yml
```diff
@@ -105,6 +105,11 @@ user_without_uid:
   email: user_without_uid@fixtures.org
   uid: ~
 
+not_valid_user:
+  email: not_valid_user@fixtures.org
+  persistence_token: cedf2dac3e2d0e7223d8d1ae126c5a18e021ad9bc6dd0a2f01677e907b3e0a1802d82273fe3ac061b5fd80209ea39632c0a9e3081a1d43bb7ced9884d7754b62
+  approval_status: approved
+
```
peggles2: A not valid user does not have contact name or organization name

app/views/complete_registration/edit.html.haml
```diff
@@ -1,10 +0,0 @@
-= content_for :title, 'Complete the sign up process'
```
peggles2: we are not using complete registration anymore. Login.gov handles that part.

app/models/user.rb
```diff
@@ -5,7 +5,8 @@ class User < ApplicationRecord
 
   validates :email, presence: true
   validates :approval_status, inclusion: APPROVAL_STATUSES
-
+  validates :contact_name, presence: true, on: :update_account
```
PatKelsh: seems like a neat way to do this

PR #537, opened 2020-02-14
change it so that the rake task only passes the number of days not the date

peggles2:
This pull request is so that our chef cookbooks would only pass a number rather than figuring out the date 76 days ago or 86 days ago. Instead it will pays the number 76 and the code will interpret that part.


peggles2:
@PatKelsh had to add 1 more  change.  I added to_i to ensure that the number of days is an integer.

PR #542, opened 2020-02-25
Add first last name to form

peggles2:
This pull request replaces anywhere that has reference to contact_name to first_name and last_name.


peggles2:
@PatKelsh one more tweak to the error message when user logs in and does not have their first and last name filled out. Thanks!

PR #546, opened 2020-03-03
SRCH-1282 - raketask populate first last name

peggles2:
This pull request is a rake task that when passed the path of a csv file it will parse it and populate first name and last name for the given email addresses.

### DIFF
--- a/lib/tasks/populate_user.rake
+++ b/lib/tasks/populate_user.rake
@@ -1,13 +1,15 @@
 # frozen_string_literal: true
+
 namespace :usasearch do
   namespace :populate_user_fields do
     desc 'populate first and last name in the user table'
     task :update_first_last_name, [:path] => [:environment] do |_t, args|
       CSV.foreach(args.path, headers: true) do |row|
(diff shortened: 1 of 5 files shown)


lib/tasks/populate_user.rake
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+namespace :usasearch do
+  namespace :populate_user_fields do
+    desc 'populate first and last name in the user table'
+    task :update_first_last_name, [:path] => [:environment] do |_t, args|
+      CSV.foreach(args.path, headers: true) do |row|
+        record = row.to_h
+        @user = User.find_by(email: record['Email'])
```
PatKelsh: There is no warning for when an email doesn't match with an existing user or if the found user already has a first and last name. Not sure if this is beyond the scope of the ticket

peggles2: Yeah this is a quick one time use to just populate those fields.  But good observation!

PR #551, opened 2020-03-19
90 deactivation email clarity 

peggles2:
In this pull request, the email warning wording is tweaked so costumers understand that even if their account is de-activated their sites will work just fine. 

Deployment notes
Templates need to be reloaded in the rails console on production to reflect the wording change:
EmailTemplate.load_default_templates

### DIFF
--- a/app/mailers/emailer.rb
+++ b/app/mailers/emailer.rb
@@ -32,7 +32,8 @@ class Emailer < ApplicationMailer
 
   def account_deactivation_warning(user, date)
     @user = user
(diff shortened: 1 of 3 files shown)


app/mailers/emailer.rb
```diff
@@ -32,7 +32,8 @@ def new_feature_adoption_to_admin
 
   def account_deactivation_warning(user, date)
     @user = user
-    @remaining_days = (date - 90.days.ago.to_date).to_i
+    days_remaining = (date - 90.days.ago.to_date).to_i
```
PatKelsh: This looks a little more clear now

PR #556, opened 2020-03-27
Changes made to find user by UID

peggles2:
This pull request allows a user to be looked up by the unique identifier passed from login.gov. 
Sometimes users are logging in with their other email account from login.gov. Sometimes they changed their email address on login.gov.  Currently the app looks up a user by email. If the user does not exist it creates them.  Additional code is added for the following scenarios:
1) user logs in with different .gov email that does not exist in our system. 
   (uid has 1 user associated in search.gov)
    User account is looked up by uid and that account is pulled up. Email has to be manualy changed by admin for now
2) user logs in with different email that is non gov and does not exists in the system. 
    (uid has 1 user associated in search.gov)
  User account is not looked up by uid because this user is non gov it is looked up by email. If email is not there the new user is created pending approval
3) user logs has multiple gov emails in search.gov and in their login.gov account.
    (uid has 2+ user associated in search.gov)
  User account is looked up by email. If it exists user logs in. If not new account created. 

This will be one of many steps towards updating the app to allow users to be able to change their email address or confirm that they wanted to login with the unsaved email address.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -125,6 +125,9 @@ class User < ApplicationRecord
   end
 
   def self.from_omniauth(auth)
+    user = find_by(uid: auth.uid)
+    return user if user.present?
+
     find_or_create_by(email: auth.info.email).tap do |user|
       user.update(uid: auth.uid)
     end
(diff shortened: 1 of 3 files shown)


peggles2:
Added some additional code that will need to be reviewed. Search.gov system may have 2 or more users with same UID.  If its just 1 user, then we look up by UID, otherwise look up by email.

PR #563, opened 2020-04-09
SRCH-1384 update LinkPopularity query search by `type` field

MothOnMars:
This PR updates the `ElasticLinkPopularityQuery` to filter by `"type": "click"`, which is necessary after passing the type to the client was deprecated in ES 6.

### DIFF
--- a/app/models/link_popularity.rb
+++ b/app/models/link_popularity.rb
@@ -1,11 +1,16 @@
+# frozen_string_literal: true
+
 class LinkPopularity
   extend LogstashPrefix
+
   def self.popularity_for(url, days_back)
     link_popularity_query = ElasticLinkPopularityQuery.new(url, days_back)
-    total = ES::ELK.client_reader.count(index: indexes_to_date(days_back, true), type: 'click', body: link_popularity_query.body)["count"]
+    total = ES::ELK.client_reader.count(
+      index: indexes_to_date(days_back, true),
+      body: link_popularity_query.body
+    )['count']
     [Math.log10(total), 1.0].max
-  rescue Elasticsearch::Transport::Transport::Errors::NotFound => e
+  rescue Elasticsearch::Transport::Transport::Errors::NotFound
     1.0
   end
-
 end
--- a/app/models/logstash_queries/elastic_link_popularity_query.rb
+++ b/app/models/logstash_queries/elastic_link_popularity_query.rb
@@ -15,6 +15,7 @@ class ElasticLinkPopularityQuery
           json.filter do
             json.bool do
               json.must do
+                json.child! { json.term { json.type 'click' } }
                 json.child! do
                   json.terms do
                     json.set! 'params.url', links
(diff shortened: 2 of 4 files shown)


app/models/link_popularity.rb
```diff
@@ -1,11 +1,16 @@
+# frozen_string_literal: true
```
MothOnMars, line 1: Added to freeze strings, which will be the default in later Ruby versions:
https://rubocop.readthedocs.io/en/latest/cops_style/#stylefrozenstringliteralcomment

app/models/link_popularity.rb
```diff
@@ -1,11 +1,16 @@
...
+
 class LinkPopularity
   extend LogstashPrefix
+
   def self.popularity_for(url, days_back)
     link_popularity_query = ElasticLinkPopularityQuery.new(url, days_back)
-    total = ES::ELK.client_reader.count(index: indexes_to_date(days_back, true), type: 'click', body: link_popularity_query.body)["count"]
```
MothOnMars, line 5: The only change here (other than wrapping the long line) is the removal of `type: 'click'` from the args passed to the ES client search, which is deprecated in ES 6.

app/models/link_popularity.rb
```diff
@@ -1,11 +1,16 @@
...
+    total = ES::ELK.client_reader.count(
+      index: indexes_to_date(days_back, true),
+      body: link_popularity_query.body
+    )['count']
     [Math.log10(total), 1.0].max
-  rescue Elasticsearch::Transport::Transport::Errors::NotFound => e
```
MothOnMars, line 7: Removed unused variable assignment, per Rubocop recommendation

app/models/logstash_queries/elastic_link_popularity_query.rb
```diff
@@ -15,6 +15,7 @@ def body
           json.filter do
             json.bool do
               json.must do
+                json.child! { json.term { json.type 'click' } }
```
MothOnMars, line 18: This adds the `"type": "click"` filter to the ES query itself, which is necessary now that we cannot pass the type as an argument to the ES client.

spec/models/logstash_queries/elastic_link_popularity_query_spec.rb
```diff
@@ -4,23 +4,28 @@
...
+              "must": [
                 {
-                  terms: {
-                    'params.url': [
-                      'https://search.gov',
-                      'https://search.gov/'
+                  "term": {
+                    "type": "click"
+                  }
+                },
```
MothOnMars, line 16: I updated the query JSON to include the new `type` filter. I also converted the test hash to use double-quotes instead of single quotes. We've done that throughout the logstash query specs, as it makes it possible to copy & paste the query into Kibana for testing.

spec/models/link_popularity_spec.rb
```diff
@@ -1,13 +1,39 @@
...
+      it 'returns the base 10 log of the click count' do
+        expect(popularity_for).to eq(1.6989700043360187)
+      end
+
+      context 'when the base 10 log of the click count is < 1' do
+        let(:click_count) { 5 }
+
+        it { is_expected.to eq(1.0) }
+      end
+    end
```
MothOnMars, line 26: added some missing specs

PR #573, opened 2020-04-21
API documentation for existing routed queries

ondrae:
This PR adds separate commits for each feature:
- instructions for how to use the routed queries feature in the API
- returns `route_to` instead of `redirect` for the routed query url attribute
- Removes the need for a param for checking for routed queries

This is a Draft PR as I expect we will want to make changes to how this feature works and the instructions, before releasing.

### DIFF
--- a/spec/controllers/api/v2/searches_controller_spec.rb
+++ b/spec/controllers/api/v2/searches_controller_spec.rb
@@ -100,7 +100,7 @@ describe Api::V2::SearchesController do
       it { is_expected.to respond_with :success }
 
       it 'returns search JSON' do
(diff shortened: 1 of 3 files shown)


spec/controllers/api/v2/searches_controller_spec.rb
```diff
@@ -512,6 +493,7 @@
       before do
         affiliate = mock_model(Affiliate, api_access_key: 'usagov_key', locale: :en, search_engine: 'BingV6')
         expect(Affiliate).to receive(:find_by_name).and_return(affiliate)
+        allow(affiliate).to receive_message_chain(:routed_queries, :joins, :where, :first)
```
ondrae, line 496: Stub the `search_engine` on the fixture affiliate, instead of the message chains.

PR #589, opened 2020-06-17
Bump gems 1514

MothOnMars:


features/support/paths.rb
```diff
@@ -23,7 +23,7 @@ def path_to(page_name)
     when /^(.*)'s strictui search page$/
       search_path(:affiliate => $1, :strictui => "1")
     when /^(.*)'s search page with unsanitized "([^\"]*)" query$/
-      search_path(:affiliate => $1, :query => "<script>#{$2}</script>")
+      search_path(:affiliate => $1, :query => "<b>#{$2}</b><script>script</script>")
```
jmax-fearless, line 26: Why the difference in query text here? I'm not following what you're doing.

spec/controllers/image_searches_controller_spec.rb
```diff
@@ -7,7 +7,7 @@
   describe '#index' do
     context 'when searching on legacy affiliate and the query is present' do
       let(:affiliate) { affiliates(:basic_affiliate) }
-      let(:query) { '<script>thunder & lightning</script>' }
+      let(:query) { '<b>thunder & lightning</b>' }
```
jmax-fearless, line 10: Also don't follow what you're doing here.

jmax-fearless:
There are various places where you replace <script> tags with <b> tags in the specs; can you comment on what you're doing there and why?

spec/lib/sanitizer_spec.rb
```diff
@@ -0,0 +1,56 @@
+require 'spec_helper'
+
+describe Sanitizer, ".sanitize" do
+  subject(:sanitize) { Sanitizer.sanitize(string) }
```
jmax-fearless, line 4: I'm not a huge fan of RSpec's subject feature, so this may reflect my ignorance. Why do you use `expect(sanitize).to` in some specs, and `is_expected.to` in others?

spec/models/boosted_content_spec.rb
```diff
@@ -66,16 +74,6 @@
...
 
-    it 'should not allow duplicate url' do
-      url = 'http://search.gov/post/9866782725/did-you-mean-roes-or-rose'
-      BoostedContent.create!(valid_attributes.merge(url: url))
-      expect { BoostedContent.create!(valid_attributes.merge(url: url)) }.to raise_error
-    end
-
-    it 'should not allow blank description' do
-      expect { BoostedContent.create!(valid_attributes.merge(description: '&nbsp;')) }.to raise_error
-    end
```
jmax-fearless, line 77: Do we cover this elsewhere? Or is this no longer needed?

lib/sanitizer.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+module Sanitizer
+  def self.sanitize(str, decode = true)
```
jmax-fearless, line 4: I think invocations of this method would be a lot clearer if `decode` was a keyword parameter. The default case would still be `Sanitizer.sanitize(foo)`, but the exceptional case would change from `Sanitizer.sanitze(foo, false)` to `Sanitizer.sanitze(foo, decode: false)`

spec/lib/sanitizer_spec.rb
MothOnMars: I use the `is_expected` syntax for simple one-liners like:
```ruby
it { is_expected.to eq '' }
```
That's the equivalent of:
```ruby
it 'is nil' do
  expect(sanitize).to be_nil
end
```
I use the former in cases where no further explanation of the behavior seems required, and the latter much more frequently in order to clarify the behavior. I'll often use the one-liner syntax for a base case spec, and the wordier syntax for the more detailed specs. (I don't have a particular rule for using one or the other beyond a gut check of when I feel like some extra verbiage would help.)

We also use the [Shoulda](http://matchers.shoulda.io/) gem for similar one-line matchers for models, controllers, etc. Those generally require a `subject`.

spec/models/boosted_content_spec.rb
MothOnMars: Whoops - I meant to call this out. I replaced the duplication spec with a Shoulda spec (the new `is_expected.to validate_uniqueness_of...` spec). That old spec just jumped out at me for refactoring, because I prefer the Shoulda specs for readability and consistency.

The `should not allow blank description` spec was a bit of a misnomer, because it was testing two pieces of behavior: 1) that a description such as '&nbsp;' would be sanitized to `""`, and 2) the resulting empty string would be invalid. We have an existing spec for the presence validation (`it { is_expected.to validate_presence_of :description }`), and I added the new, separate spec to test the sanitization: `it_behaves_like 'a record that sanitizes attributes', [:title, :description]`.

lib/sanitizer.rb
MothOnMars: 👍

PR #590, opened 2020-06-24
Added comments for finding old code in srch-1525

ondrae:
As part of removing the [legacy SERPs](https://cm-jira.usa.gov/browse/SRCH-368), these comments can help remove some of the code used for the old way click tracking used to work. I would also look at the methods in [results_helper.rb](https://github.com/GSA/search-gov/blob/master/app/helpers/results_helper.rb), besides the first two.


ondrae:
I would ignore these CodeClimate suggestions. This PR is to help delete the code causing them.

PR #592, opened 2020-06-30
SRCH-1534 send ApplicationJob stats to Datadog

MothOnMars:
This PR is intended to ensure that the job classes that inherit from ApplicationJob send their job statistics to Datadog.


MothOnMars:
This is going to require some additional work. Closing until that's done.

PR #594, opened 2020-07-07
SRCH-1524 - Stream sitemaps instead of loading them into memory.

ondrae:
This PR:
- Replaces the Sitemaps gem with the Saxerator xml streaming gem.
- Streams sitemap xml files instead of loading them into memory first.

Details are in [the jira ticket](https://cm-jira.usa.gov/browse/SRCH-1524) but this uses basically no memory and is four times as fast.


app/models/sitemap_indexer.rb
```diff
@@ -12,28 +14,30 @@ def initialize(sitemap_url:)
   end
 
   def index
-    sitemaps.any? ? enqueue_sitemaps : process_entries
+    enqueue_sitemaps
+    process_entries
```
MothOnMars, line 18: @ondrae, it's not the end of the world if these methods are called sequentially, but it's obscuring the fact that any given sitemap will *either* be a sitemap *or* a sitemap index, but never both. In keeping with the original logic, we could add a method to check to see if the sitemap is a sitemap index (as identified by the `<sitemapindex>` tag), in which case we would `enqueue_sitemaps`, otherwise `process_entries`.

app/models/sitemap_indexer.rb
```diff
@@ -12,28 +14,30 @@ def initialize(sitemap_url:)
...
   private
 
-  def sitemaps
-    @sitemaps ||= Sitemaps.parse(sitemap).sitemaps
-  end
-
-  def sitemap_entries
-    # Eventually we might add an option to the Sitemaps gem to limit the URLS
-    # to those strictly adhering to the sitemap protocol, but this should suffice for now
-    # https://www.pivotaltracker.com/story/show/157485118
```
MothOnMars, line 27: Please move this comment above the section where we ensure that the domain of the URL matches the domain of the sitemap:
```
if URI(entry['loc'].strip).host == domain
```

app/models/sitemap_indexer.rb
MothOnMars: P.S. Maybe I'm overthinking this - we could also solve this with a comment...

ondrae: I went with 
```
  def index
    # A sitemap has either a sitemapindex or a urlset
    sitemapindex? ? enqueue_sitemaps : process_entries
  end

  private

  def sitemapindex?
    Saxerator.parser(sitemap).for_tag('sitemapindex').any?
  end
```

ondrae:
If you aren't using the `sitemap_entries_found: count` logline for anything, pulling it out will solve the CodeClimate suggestions.

app/models/sitemap_indexer.rb
MothOnMars: 👍

PR #601, opened 2020-07-14
fix DocumentFetcher redirection spec

MothOnMars:
This PR fixes a spec that has been failing intermittently.


jmax-fearless:
Definitely an improvement.

PR #603, opened 2020-07-20
SRCH-1511 SRCH-1560 improve error handling in SitemapIndexer

MothOnMars:
This PR:
- rescues some known errors in the SitemapIndexer, and improves resiliency from unknown errors
- moves the observer classes into the new `observers` directory
- disables the Rubocop cop that requires rescues to specify an error class
- re-enables the `Sitemap` model specs

(Note: I am holding off on moving the SitemapIndexer & other service classes into a `services` directory, to make it clearer what is changing in this commit. That reorganization will happen in a subsequent story.)


spec/models/sitemap_spec.rb
```diff
@@ -1,43 +1,41 @@
 require 'spec_helper'
 
 describe Sitemap do
-  pending('until the sitemapindexer gem file is updated for rails 5 ') do
```
MothOnMars, line 4: There are actually no changes to this file other than the removal of the `pending` block, and the corresponding changes in indentation. (These specs were supposed to have been re-enabled after the Rails 5 upgrade, but that apparently never happened.)

PR #616, opened 2020-08-27
Remove json

jmax-fearless:
Turns out, per a suggestion from @MothOnMars, that it was simplest to just pull the JSON gem and let the code use Ruby's native JSON support. Worked on the first try, astonishingly.


jmax-fearless:
SRCH-1588

jmax-fearless:
@MothOnMars - The changes to db/structure.sql are actually unrelated to this work, but seem like a good thing to me. We can discuss further if you disagree, or if you'd just rather that they be pulled out to their own ticket. They happened when I ran the latest migration locally; after investigating the changes, I approve of them and want to keep them, one way or another.

jmax-fearless:
Got rid of the offending dependency; the remaining uses of the JSON gem are all in development gems, and thus do not pose a risk in production.

jmax-fearless:
@lsamuels-fearless - Added you as a reviewer. We need to get this one into production by 9/22, otherwise I'd wait for @MothOnMars  to return. If you're not comfortable reviewing/signing off on this, then we need to pull in another developer.

MothOnMars:
@jmax-fearless , the `recaptcha` gem also depends on `json`; that's a production gem, so that will still be an issue. However, `recaptcha` doesn't specify a `json` version, so now you should be able to `bundle update json` to bump it to a secure version.

jmax-fearless:
@lsamuels-fearless - Hold off on reviewing this, please; something interesting has surfaced after I addressed the last comment from @MothOnMars 

@MothOnMars - Remember that intermittent resque spec failure we've been chasing for a bit? Well, coming up to json version 2.3.1 makes it fail consistently. I'm not sure that it's really the same issue, though, because it seems to be failing differently than I remember for that spec. The spec is to check that we don't enqueue duplicate rss_fetcher jobs, and the failure is on creation of the _first_ job - the one we expect to succeed.

Investigating.

jmax-fearless:
@MothOnMars - Ready for re-review.

MothOnMars (approved):
LGTM

MothOnMars:
P.S. The commit message should be prefixed with the ticket id: `SRCH-123...`. You can either update the commit message now or update it when this is merged.

PR #624, opened 2020-10-06
SRCH-1678 format click module as a string in drilldown CSV

MothOnMars:
This PR ensures that the click module in the downloaded CSV appears as a string (`BWEB`), regardless of whether it is stored in Elasticsearch as an array or a string.

### DIFF
--- a/app/controllers/sites/click_drilldowns_controller.rb
+++ b/app/controllers/sites/click_drilldowns_controller.rb
@@ -33,7 +33,7 @@ class Sites::ClickDrilldownsController < Sites::SetupSiteController
     record << doc['request']
     record << doc['referrer']
     record << doc['vertical']
-    record << doc['modules']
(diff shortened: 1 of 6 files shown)


app/controllers/sites/click_drilldowns_controller.rb
```diff
@@ -33,7 +33,7 @@ def document_mapping(doc)
     record << doc['request']
     record << doc['referrer']
     record << doc['vertical']
-    record << doc['modules']
+    record << Array(doc['modules']).first
```
jmax-fearless, line 36: After a minute with irb, I see what you're doing here. Could you move it out to a function, so the function name makes that obvious? 'force_click_module_to_string' or something, perhaps.

jmax-fearless (changes requested):

jmax-fearless (approved):

PR #633, opened 2020-10-23
SRCH-1695 add CORS support to click API

MothOnMars:
- fix flakey human sessions controller spec (for real this time)
- robocop-based spec refactoring

### DIFF
--- a/lib/middlewares/filtered_cors.rb
+++ b/lib/middlewares/filtered_cors.rb
@@ -1,5 +1,9 @@
+# frozen_string_literal: true
+
+# This class should eventually be replaced and its specs refactored
+# in favor of configuration based on the rack-cors gem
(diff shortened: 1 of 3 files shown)


lib/middlewares/filtered_cors.rb
```diff
@@ -1,5 +1,9 @@
+# frozen_string_literal: true
+
+# This class should eventually be replaced and its specs refactored
+# in favor of configuration based on the rack-cors gem
```
MothOnMars, line 4: This isn't a change worthy of a ticket, but I'm leaving the comment as a signpost that there's an opportunity for refactoring if we ever need to make deeper CORS changes.

jmax-fearless (approved):

PR #638, opened 2020-11-20
SRCH-1725 upgrade Rails to 5.2.4.4

MothOnMars:
This PR resolves https://github.com/advisories/GHSA-cfjv-5498-mph5


jmax-fearless (approved):
Verified that all gem bumps are reasonable for the minor version upgrade.

PR #647, opened 2020-12-18
Turned on secure cookies.

jmax-fearless:
Disabled force_ssl for /healthcheck. Our ELB's use that path as a
health check, and won't work right if we try to force SSL for it.


jmax-fearless:
Wrong tip version.

PR #655, opened 2021-02-02
SRCH-1547 fix CTRs in analytics downloads

MothOnMars:
- bump cucumber
- add cucumber test for query downloads
- rubocop cleanup


features/admin_center_clicks_and_queries.feature
```diff
@@ -26,3 +26,8 @@ Feature: Clicks and Queries stats
     Then I should see "Referrers"
     And I should see "Your site has not received any queries with referrers yet."
 
+    When I follow "Monthly Reports"
+    Then I should see "Change Month"
+    When I follow "csv"
+    Then I should get a CSV download
+    And the downloaded file should include "Search Term"
```
MothOnMars, line 33: We don't have any analytics data in our spec environment (and generating it is a very old backlog item), so this test is intended as a simple "make sure nothing blows up" sanity check.

jmax-fearless (approved):

PR #656, opened 2021-02-08
SRCH-1771 block html injection

jmax-fearless:
Creating to see if Codeclimate has any complaints. Will be ready for review after addressing any such issues.


jmax-fearless:
@MothOnMars - This is ready for review.

spec/models/routed_query_spec.rb
```diff
@@ -1,75 +1,127 @@
...
+          }
+        )
+      end
+
+      let(:rq) { affiliate.routed_queries.build(attributes_with_html) }
+
+      before { rq.valid? }
+
+      it 'strips the html from the keyword' do
+        expect(rq.routed_query_keywords[0].keyword).to eq('some keyword phrase')
```
MothOnMars, line 110: These specs should be moved to spec/models/routed_query_keyword_spec.rb. We should test the model instance directly, rather than through an associated object.

MothOnMars (changes requested):

spec/models/routed_query_spec.rb
jmax-fearless: Moved.

jmax-fearless:
@MothOnMars - Comment addressed; ready for re-review.

spec/models/routed_query_keyword_spec.rb
```diff
@@ -34,6 +34,18 @@
     end
   end
 
+  describe 'creating a keyword which contains html' do
```
MothOnMars: Two more minor tweaks: 
- move this spec within the `Creating new instance` block
- use `context` rather than `describe`:
```ruby
context 'when the keyword contains html' do
```

MothOnMars (changes requested):

spec/models/routed_query_keyword_spec.rb
jmax-fearless: Done.

jmax-fearless:
@MothOnMars Ready for re-re-review.

MothOnMars (approved):

MothOnMars:
LGTM

PR #660, opened 2021-02-09
SRCH-1905 resolve intermittent cucumber failures

MothOnMars:
- bump test gems
- bump Nokogiri to resolve https://github.com/advisories/GHSA-vr8q-g5c7-m54m
- simplify CircleCi gem installation
- tweak VCR/Webmock settings


MothOnMars:
@jmax-fearless , this has already passed CI multiple times, but I'm going to keep re-running it until it's merged. So don't be surprised to see test runs in progress...

Gemfile.lock
```diff
@@ -258,8 +258,8 @@ GEM
       activerecord (>= 4.2)
       activesupport (>= 4.2)
       after_commit_action (~> 1.0)
-    crack (0.4.3)
-      safe_yaml (~> 1.0.0)
+    crack (0.4.5)
+      rexml
```
jmax-fearless, line 262: That's an interesting change. They replaced a YAML library with an XML one.

spec/support/shared_vcr_setup.rb
```diff
@@ -36,15 +36,13 @@
...
 
-  config.ignore_hosts 'example.com', 'codeclimate.com'
+  config.ignore_hosts 'example.com', 'codeclimate.com', '127.0.0.1'
   config.ignore_hosts *DRIVER_HOSTS
   config.ignore_request do |request|
     /codeclimate.com/ ===  URI(request.uri).host
   end
 
   config.ignore_request { |request| URI(request.uri).port.between?(9200,9299) } #Elasticsearch
-  config.ignore_request { |request| URI(request.uri).port == 9515 } # Selenium Webdriver
```
jmax-fearless, line 46: Why aren't these necessary any more?

jmax-fearless (approved):

Gemfile.lock
MothOnMars: It only appears that way because of the bump from 0.4.3 to 0.4.5 - they removed `safe_yaml` years ago, and just recently added `rexml`:
https://github.com/jnunemaker/crack/commits/master/crack.gemspec

spec/support/shared_vcr_setup.rb
MothOnMars: Those are now covered by `config.ignore_hosts ... '127.0.0.1'`

PR #665, opened 2021-02-23
SRCH-1926 update CircleCi config to use updated ruby 2.5.5 image

MothOnMars:
This PR updates the Dockerfile used to create our our custom Ruby 2.5.5 image: https://hub.docker.com/repository/docker/searchgov/ruby

It also  updates our CircleCi config to use the updated image. This lays the groundwork necessary to run our search-gov specs against multiple Ruby versions in CircleCi. I

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,8 +12,8 @@ executors:
     working_directory: ~/search-gov
 
     docker:
-      # using custom image, see .circleci/images/primary/Dockerfile
-      - image: searchgov/circleci:20190417
+      # using custom images, see .circleci/images/primary/
+      - image: searchgov/ruby:2.5.5
         environment:
           RAILS_ENV: test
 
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -104,7 +107,8 @@ jobs:
       - restore_cache:
           key: repo-{{ .Environment.CIRCLE_SHA1 }}
 
-      - ruby/install-deps
+      - ruby/install-deps:
```
jmax-fearless, line 110: This and similar lines below are grabbing a version-specific set of gems, correct?

jmax-fearless (approved):

jmax-fearless (approved):

.circleci/config.yml
MothOnMars: Yes. See https://circleci.com/developer/orbs/orb/circleci/ruby#commands-install-deps for what that is doing under the hood.

PR #671, opened 2021-03-17
SRCH-1915 fix Tweet.expire method

MothOnMars:
This PR fixes the broken `Tweet.expire` method. It also adds a meaningful spec for that method, adds an index on the `published_at` column to support the query used by that method, and resolves several Rubocop offenses.

### DIFF
--- a/spec/models/tweet_spec.rb
+++ b/spec/models/tweet_spec.rb
@@ -39,7 +39,7 @@ describe Tweet do
     is_expected.to validate_uniqueness_of :tweet_id
   end
 
-  it 'should sanitize tweet text' do
+  it 'sanitizes tweet text' do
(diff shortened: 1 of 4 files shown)


spec/models/tweet_spec.rb
```diff
@@ -1,17 +1,15 @@
-require 'spec_helper'
+# frozen_string_literal: true
 
 describe Tweet do
-  fixtures :affiliates
-
-  before do
-    allow(Twitter).to receive(:user).and_return double('Twitter', id: 12345, name: 'USASearch', profile_image_url: 'http://some.gov/url')
```
MothOnMars, line 7: I removed this `allow(Twitter).to receive(:user)`... line, as it was related to code that is no longer used.

jmax-fearless (approved):
LGTM

PR #672, opened 2021-03-19
SRCH-1989 bundle update gems

MothOnMars:
- temporarily lock omniauth_login_dot_gov ref

This PR is the result of running `bundle update`, which succeeds after locking the ref for the `omniauth_login_dot_gov`. `bundle update` bumped enough of the gems to new minor or major versions that I'll plan to have these changes smoke-tested and released separately. Going forward, we should be able to run `bundle update` on our feature branches.


jmax-fearless (approved):
LGTM

jmax-fearless (approved):
LGTM

PR #673, opened 2021-03-19
SRCH-1982 merge "IgnoredMethods" in Rubocop config

MothOnMars:
This PR ensures that `IgnoredMethods` for certain cops as specified by the `searchgov_style` gem are merged with `IgnoredMethods` specific to this repo. It also removes the existing `IgnoredMethods: ['require']` line for the `Style/MethodCallWithArgsParentheses` cop, as that is specified in searchgov_style:
https://github.com/GSA/searchgov_style/blob/main/.default.yml#L112

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -6,6 +6,7 @@ inherit_from:
 inherit_mode:
   merge:
     - Exclude
+    - IgnoredMethods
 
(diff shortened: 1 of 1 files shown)


jmax-fearless (approved):
LGTM

PR #680, opened 2021-03-30
SRCH-2014 run legacy SERP features against responsive SERP

MothOnMars:
This PR is the first in a series intended to get all our SERP tests, most of which currently run against the "legacy SERP", running against our current, responsive SERP. When reviewing, you can mostly ignore the auto-generated files in `features/vcr_cassettes`, which are recordings of external requests made during the feature runs. The new `features/searches.feature` is mostly a straight copy and paste from `features/legacy_search.feature`. It has been slightly tweaked to run the features against the current SERP. Failing scenarios have been tagged with `@wip`, and will be updated and re-enabled in subsequent tickets. (The `features/searches.feature` will be cleaned up and refactored going forward, but for now this is intended as a warts-and-all, copy and paste dump intended to get our features running against the responsive SERP as quickly as possible.)

### DIFF
--- a/features/searches.feature
+++ b/features/searches.feature
@@ -1,0 +1,1161 @@
+# This feature file has been copied and MINIMALLY updated from the original
+# legacy_search.feature file. Once these tests are all passing against the current,
+# responsive SERP, we should consolidate these with the other features in
+# responsive_search.feature and mobile_searches.feature
+Feature: Search
+  In order to get government-related information from specific affiliate agencies
+  As a site visitor
+  I want to be able to search for information
+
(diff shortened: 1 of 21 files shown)


jmax-fearless (approved):
Can you comment on the JIRA ticket about the wip'd tests, please? I'm guessing that they were that way in the original, but positive confirmation would be nice. Or an explanation if they're freshly wip'd.

Those are reasonable tests, if a bit lengthy; too many things being checked in one scenario for my taste. When we integrate the two SERPs' test suites, we should maybe see if we can clean up a little too. 

LGTM.

PR #687, opened 2021-04-01
SRCH-1929 - Make CircleCI use a Ruby version matrix

jmax-fearless:


MothOnMars:
I agree that we can think about streamlining this in the future. For an MVP, this will do just fine. LGTM.

MothOnMars (approved):

PR #691, opened 2021-04-01
SRCH-2031 remove feature re. invalid Lucene characters

MothOnMars:
This PR removes the `When a searcher enters a query with invalid Lucene character` cucumber feature. That feature was added following a fix to workaround a Solr bug in 2012:
https://github.com/GSA/search-gov/commit/b2ee1e37b490a89e26531487402103a15698a6c9

We retired Solr in 2014, so we can probably live without an integration test for this case anymore.


jmax-fearless (approved):
LGTM

PR #700, opened 2021-04-06
SRCH-2040 update and enable searching with sitelimit feature

MothOnMars:
This PR updates and enables the "Searching for site specific results using sitelimit" to run against the responsive SERP. I have simplified the feature significantly, as most of the sitelimit behavior that was tested against the legacy SERP (namely, the sitelimit persisting from search to search) has not been implemented for the responsive SERP.


jmax-fearless (approved):
LGTM

PR #703, opened 2021-04-07
SRCH-2049 remove feature re. searching with custom Dublin core mappings

MothOnMars:
This PR removes the `Searching news items with custom dublin core mappings` feature, which is not relevant to the current SERP.


jmax-fearless (approved):
LGTM

PR #704, opened 2021-04-07
SRCH-2017 remove feature re. searching with Dublin Core mappings

MothOnMars:
This feature was specific to the legacy SERP, so I'm removing it rather than updating it for the responsive SERP.


jmax-fearless (approved):
LGTM

PR #707, opened 2021-04-07
SRCH-2016 remove duplicate feature re. searching RSS items

MothOnMars:
This behavior is already tested thoroughly against the responsive SERP via the `News Search` and `Custom date range news search` features in `features/mobile_searches.feature`.


jmax-fearless (approved):
LGTM

PR #714, opened 2021-04-14
SRCH-2037 Update to Ruby 2.6.6

jmax-fearless:
Updated .ruby-version.
CircleCi now tests against both 2.5.5 and 2.6.6


MothOnMars (approved):
LGTM. (I briefly considered whether we even want to run the coverage report on multiple versions, and then I remembered that one time I lost a day because a Ruby upgrade actually changed the code coverage calculation. So yeah.)

PR #719, opened 2021-04-29
SRCH-2058 remove admin center controls for legacy SERP images

MothOnMars:
This PR:
- removes admin center controls for legacy SERP images
- prevents legacy images from being cloned
- bundle updated gems

The CodeClimate failures are all pre-existing, related to the `SiteCloner` service. It's flagging them as new because I moved that file from `app/models` to `app/services`. (I opted not to do a lot of Rubocop cleanup as part of this PR as there were already a lot of changes.)


jmax-fearless (approved):
LGTM

PR #724, opened 2021-05-07
SRCH-2221 Remove Ruby 2.5 from search-gov CircleCI

jmax-fearless:
Also added a nice name display for the matrixed build versions.


MothOnMars (approved):
LGTM

PR #735, opened 2021-06-03
SRCH-2308 add specs for NewsSearchHelper#current_time_filter_description

MothOnMars:
This PR adds specs for a method that was previously only tested through another method that will be removed along with the legacy SERP code.


jmax-fearless:
LGTM

jmax-fearless (approved):

PR #739, opened 2021-06-09
SRCH-2331 the basic_affiliate fixture uses the default search engine

MothOnMars:
This PR updates the `basic_affiliate` fixture to use the default search engine (currently BingV7) instead of the deprecated Google search engine. It also includes updates to the VCR cassettes and some Rubocop fixes. The remaining Rubocop failures are preexisting issues that I am not refactoring as part of this PR.

### DIFF
--- a/spec/models/web_search_spec.rb
+++ b/spec/models/web_search_spec.rb
@@ -116,23 +116,6 @@ describe WebSearch do
         described_class.new(@valid_options).send(:search)
       end
     end
-
-    context 'when Google is the engine' do
-      before do
-        @affiliate = affiliates(:basic_affiliate)
-        @valid_options = {query: 'government', affiliate: @affiliate}
-        google_search = GoogleWebSearch.new(@valid_options)
-        allow(GoogleWebSearch).to receive(:new).and_return google_search
-        allow(google_search).to receive(:execute_query)
-      end
-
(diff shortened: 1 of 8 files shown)


spec/models/web_search_spec.rb
```diff
@@ -67,149 +65,133 @@
...
+      expect(described_class.new(valid_options).cache_key).to eq("government (site:gov OR site:mil):{:query=>\"government\", :page=>5, :affiliate_id=>#{affiliate.id}}:BingV6")
     end
   end
 
   describe 'instrumenting search engine calls' do
     context 'when BingV6 is the engine' do
       before do
-        @valid_options = {query: 'government', affiliate: affiliate}
-        bing_search = BingV6WebSearch.new(@valid_options)
+        valid_options = { query: 'government', affiliate: affiliate }
```
jmax-fearless, line 86: I don't think this is right; `valid_options` should be a context `let` here, not a local variable, yes?

jmax-fearless (changes requested):
Only one comment, but I think it's a real issue; please address or convince me it's not.

spec/models/web_search_spec.rb
MothOnMars: Good catch - that was an oversight. Fixing...

jmax-fearless (approved):
LGTM

PR #736, opened 2021-06-04
SRCH-243 SRCH-1525 SRCH-2109 remove legacy SERP code

MothOnMars:
SRCH-243 SRCH-1525 SRCH-2109 remove legacy SERP code
- remove legacy views, helpers & click-tracking code
- remove "Compare Search Results" super admin feature
- remove advanced-firstgov logic
- remove deprecated or duplicated cucumber features
- remove legacy seed & fixtures
- remove legacy spec & feature helpers
- redirect video news search path to search path
- disable duplication check for specs

PR #742, opened 2021-06-15
SRCH-2336 deprecate the /api/search endpoint

MothOnMars:
- add comments re. the deprecated Agencies API

The CodeClimate failure is an `RSpec/DescribeClass` failure for the new `spec/requests/api/v1/search_spec.rb ` ("The first argument to describe should be the class or module being tested"). That cop isn't valid for request specs, so I've put in a [PR](https://github.com/rubocop/rubocop-rspec/pull/1161) to change this in `rubocop-rspec`.


app/controllers/api/v1/search_controller.rb
```diff
@@ -0,0 +1,13 @@
...
+  module V1
+    class SearchController < ApplicationController
+      DEPRECATION_MESSAGE = 'This API endpoint has been deprecated. Please refer to https://search.gov/developer/ for information on current Search.gov APIs.'
+
+      def search
+        render plain: DEPRECATION_MESSAGE, status: :not_found
+      end
+    end
+  end
+end
```
MothOnMars, line 13: This file is actually just `app/controllers/api_controller.rb`, which I've moved (to get it out of the top-level controllers directory), renamed, and cleaned up.

lib/middlewares/reject_invalid_request_uri.rb
```diff
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
```
MothOnMars, line 1: If these changes look familiar, it's because they were already approved in https://github.com/GSA/search-gov/pull/736/files#diff-1b78e73a5db3338d94d2e1b5d61435284fab90dd3e1c67279d66a9f7e824d727. Adding the magic comment to request/controller files & specs always broke this, to it will be good to get it into master sooner rather than later.

jmax-fearless (approved):
LGTM

(Thank you for the comment about `search_controller.rb`; it saved a couple minutes of spelunking.

PR #745, opened 2021-06-23
SRCH-2260 fix affiliate rss feed urls

jmax-fearless:
SRCH-2260


jmax-fearless:
The remaining two Code Climate issues are existing code; requesting approval without addressing them.

app/models/affiliate.rb
```diff
@@ -722,17 +734,13 @@ def set_json_fields
...
   def clear_existing_attachments
-    if mobile_logo? and !mobile_logo.dirty? and mark_mobile_logo_for_deletion == '1'
-      mobile_logo.clear
-    end
-
-    if header_tagline_logo? and !header_tagline_logo.dirty? and mark_header_tagline_logo_for_deletion == '1'
-      header_tagline_logo.clear
-    end
+    mobile_logo.clear if mobile_logo? && !mobile_logo.dirty? && mark_mobile_logo_for_deletion == '1'
+    header_tagline_logo.clear if header_tagline_logo? && !header_tagline_logo.dirty? && mark_header_tagline_logo_for_deletion == '1'
```
MothOnMars, line 743: Let's revert this change. The original is more readable, and the `Style/IfUnlessModifier` failure for the original is actually a false positive caused by our disabling of the line length cop. The recommendation is to use multiple lines if the condition would be too long for a single line, i.e.
```
if long_condition_that_prevents_code_fit_on_single_line
  do_something_with_a_long_name(arg)
end
```
https://docs.rubocop.org/rubocop/cops_style.html#styleifunlessmodifier

MothOnMars (changes requested):
One style tweak.

app/models/affiliate.rb
jmax-fearless: Reverted.

app/models/affiliate.rb
```diff
@@ -739,8 +739,13 @@ def load_css_properties
...
   def clear_existing_attachments
-    mobile_logo.clear if mobile_logo? && !mobile_logo.dirty? && mark_mobile_logo_for_deletion == '1'
-    header_tagline_logo.clear if header_tagline_logo? && !header_tagline_logo.dirty? && mark_header_tagline_logo_for_deletion == '1'
+    if mobile_logo? and !mobile_logo.dirty? and mark_mobile_logo_for_deletion == '1'
+      mobile_logo.clear
+    end
+
+    if header_tagline_logo? and !header_tagline_logo.dirty? and mark_header_tagline_logo_for_deletion == '1'
+      header_tagline_logo.clear
+    end
```
MothOnMars: Whoops - I should have clarified that we can keep the change from `and` to `&&`.

jmax-fearless: Done.

I originally did this fix fairly mechanically, but now that I look and think, I'm minded to pull those two conditions out to their own predicates (`#need_to_clear_mobile_logo` and `#need_to_clear_header_tagline_logo`), just for the additional clarity. Objections?

MothOnMars: Let's just leave it. Additional refactoring is out of scope for this ticket.

MothOnMars (approved):

MothOnMars:
LGTM, pending specs passing CI.

MothOnMars (changes requested):
@jmax-fearless , apologies for requesting changes after approving this, but I had a change of heart about this change:
https://github.com/GSA/search-gov/pull/745/commits/13fc6e20af271981aad3df64efffa7c83dd02385

Let's back out that change and disable the `Style/ClassAndModuleChildren` cop (just in search-gov). There are a few reasons:
- the change is style-only, and the existing style isn't dangerous
- the existing style isn't getting in our way or cluttering up our code
- if we enforce this style, we'll end up having to make similar changes in a bunch of other controllers
- ...all of which I'd like to avoid, because the result is that the indentation change across the entire file renders `git blame` mostly useless for those files. 

I'm open to discussion if you feel otherwise.

app/controllers/sites/rss_feeds_controller.rb
```diff
@@ -1,108 +1,106 @@
...
-module Sites
-  class RssFeedsController < Sites::SetupSiteController
-    include ActionView::Helpers::TextHelper
-    before_action :setup_rss_feed, only: [:show, :edit, :update]
-    before_action :setup_non_managed_rss_feed, only: [:destroy]
-
-    def index
-      @rss_feeds = @site.rss_feeds.order(:name)
-    end
+class Sites::RssFeedsController  < Sites::SetupSiteController
```
MothOnMars, line 3: - remove the extra space before the `<`
- disable the Style/ClassAndModuleChildren cop in `.rubocop.yml`

MothOnMars (approved):

PR #750, opened 2021-07-14
SRCH-2106 remove legacy assets

MothOnMars:
SRCH-2106 remove legacy assets
- update asset-related gems

As noted, there will be another round of asset removal for a subsequent ticket, but this is a start...


jmax-fearless (approved):
LGTM

PR #758, opened 2021-07-27
SRCH-2243 change bulk upload email error message order

jmax-fearless:
SRCH-2243 change bulk upload email error message order.

Make 'Url already taken' errors appear after all other errors.


MothOnMars:
@jmax-fearless , please squash the commits into a single commit that includes the Jira ticket # in the commit message.

spec/mailers/bulk_url_upload_results_mailer_spec.rb
```diff
@@ -93,6 +76,15 @@
...
       end
+
+      it 'shows the urls with error "url already taken" after all the other errors' do
+        first_error_message_position = mail_body.index(first_error_message)
+        second_error_message_position = mail_body.index(second_error_message)
+        already_taken_error_message_position = mail_body.index(already_taken_error_message)
+
+        expect(already_taken_error_message_position).to be > first_error_message_position
+        expect(already_taken_error_message_position).to be > second_error_message_position
+      end
```
MothOnMars, line 87: @jmax-fearless , this is not a valid spec, because it passes without any of the changes in `spec/services/bulk_url_uploader_spec.rb`. The `before` block sets the errors in the correct order, so the spec is testing errors that are already in the expected order:
```ruby
      before do
        results.add_error(first_error_message, first_bad_url)
        results.add_error(second_error_message, second_bad_url)
        results.add_error(already_taken_error_message, duplicate_url)
      end
 ```

Please add a valid spec using TDD. As the results are being ordered in `BulkUrlUploader::Results`, you could add a unit test for that class, rather than testing this via the mailer.

File: app/services/bulk_url_uploader.rb [hunk 2]
```diff
@@ -35,7 +36,9 @@ def total_count
     end
 
     def error_messages
-      @errors.keys
+      already_taken = ->(error) { error == URL_ALREADY_TAKEN_MESSAGE }
+      errors = @errors.keys
+      errors.reject(&already_taken) + errors.select(&already_taken)
```
MothOnMars, line 41: To avoid future head-scratching, please add a comment about why the re-ordering is necessary.

MothOnMars (changes requested):

app/services/bulk_url_uploader.rb
jmax-fearless: Comment added.

spec/mailers/bulk_url_upload_results_mailer_spec.rb
jmax-fearless: I'm not comfortable removing the mailer spec, but I agree that a unit test for this would be a fine thing. So I added one, and also re-arranged the order of `add_error` calls per your comment. Both test fail if you remove the changes to `#error_messages`

spec/services/bulk_url_uploader/results_spec.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+describe BulkUrlUploader::Results do
+  subject(:results) { described_class.new('test bulk uploader') }
+
+  describe '#errors' do
+    context 'when there are "url not found" errors' do
```
MothOnMars, line 7: This context should be `when there are "Url has already been taken" errors`

spec/services/bulk_url_uploader/results_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+      before do
+        results.add_error('Validation failed: Url has already been taken', 'junk')
+        results.add_error('Validation failed: SearchgovDomain is not a valid SearchgovDomain', 'junk')
```
MothOnMars, line 10: Please change `junk` to a URL, i.e. `https://alreadytaken.gov`/`https://invaliddomain.gov`. That will help document the behavior of the code, whereas a future developer looking at this spec would not know what `'junk'` might be.

MothOnMars (changes requested):

spec/services/bulk_url_uploader/results_spec.rb
jmax-fearless: Changed

spec/services/bulk_url_uploader/results_spec.rb
jmax-fearless: Changed.

MothOnMars (approved):
LGTM

PR #751, opened 2021-07-19
SRCH-2111 remove additional legacy SERP code

MothOnMars:
- remove unused image assets and stylesheets
- removed unused cucumber helpers

PR #763, opened 2021-08-03
SRCH-2423 return assets used for affiliate type-ahead

MothOnMars:
This PR returns the SAYT assets that are used to style the search boxes on our affiliate websites. I realized in preliminary testing on the feature branch that these are still needed, after removing them in https://github.com/GSA/search-gov/pull/751. The files being resurrected are unchanged from master.


jmax-fearless (approved):
LGTM

PR #765, opened 2021-08-11
SRCH-2410 Remove tweetstream and replace with streaming from the twitter gem.

jmax-fearless:
SRCH-2410 Remove tweetstream and replace with streaming from the twitter gem.


jmax-fearless:
Requesting review despite the codeclimate complaints, which break down as follows:
-The two complaints about specs for the `refresh_lists` and `refresh_lists_statuses`, as well as the two issues in twitter_profiles_helper_spec are existing issues that CC has flagged as new.
- The complaints in twitter_streamer_spec about using `to receive` are for a spec that I do not believe can be written in `to have_recieved` form
- The complaints in twitter_streamer_spec about stubbing `sleep` on the class under test - the only other way I found to accomplish this is to use allow_any_instance_of(Object), which is even uglier. I welcome any suggestions on cleaning this up.

MothOnMars:
Please update the commit message with the Jira ticket # (and add me as a reviewer :) )

MothOnMars:
Adding "do not merge" label as a reminder that this release needs to go out in tandem with https://github.com/GSA/usasearch-cookbooks/pull/521

PR #771, opened 2021-08-24
SRCH-2324 update newrelic_rpm

jmax-fearless:
- Bumped newrelic_rpm to 6.15.0 and ran `bundle update newrelic_rpm`

- Set backport_fast_active_record_connection_lookup to true.  As of
  6.5, newrelic_rpm defaults this to false, but recommends setting it to
  true unless that causes problems with other gems' monkey-patching of
  ActiveRecord.


MothOnMars (approved):
LGTM

Fool256 (approved):

PR #776, opened 2021-08-30
SRCH-2251 upgrade authlogic to >= 6

jmax-fearless:
SRCH-2251 upgrade authlogic to >= 6

## Summary
- Update the authlogic gem

#### Functionality Checks

- [x] Code is functional.

#### Process Checks


app/models/user.rb
```diff
@@ -8,9 +8,10 @@ class User < ApplicationRecord
   validates :first_name, presence: true, on: :update_account
   validates :last_name, presence: true, on: :update_account
   validates :organization_name, presence: true, on: :update_account
+  validates :email, presence: true, uniqueness: { case_sensitive: false }
```
MothOnMars, line 11: We should continue to validate the email length & format, which authlogic used to do for us. Please add those validations with corresponding specs. See https://github.com/binarylogic/authlogic/blob/master/doc/use_normal_rails_validation.md

MothOnMars (changes requested):

MothOnMars:
P.S. as an alternative to the regex used by Authlogic, we could use Ruby's built-in pattern: `URI::MailTo::EMAIL_REGEXP`

app/models/user.rb
jmax-fearless: Added validation and test therof.

spec/models/user_spec.rb
```diff
@@ -39,6 +39,7 @@
 
     it { is_expected.to validate_presence_of :email }
     it { is_expected.to validate_uniqueness_of(:email).case_insensitive }
+    it { is_expected.to_not allow_value('not_a_valid_email').for(:email) }
```
MothOnMars, line 42: This should be `not_to` instead of `to_not`. (Code Climate *should* be flagging that, as our Rubocop configuration enforces the `not_to` format. I have a support ticket open w/ Code Climate to investigate the false negative.)

MothOnMars (changes requested):

spec/models/user_spec.rb
jmax-fearless: Changed.

MothOnMars (approved):
LGTM

MothOnMars (approved):
LGTM

PR #866, opened 2022-02-04
SRCH-2760 specify 'include_type_name: true' when creating indices

MothOnMars:
## Summary
This PR is for forward-compatibility with ES7. This is part of a multi-version shift away from using types within ES indices:
https://www.elastic.co/guide/en/elasticsearch/reference/6.8/removal-of-types.html
https://www.elastic.co/blog/moving-from-types-to-typeless-apis-in-elasticsearch-7-0

The `include_type_name: true` parameter actually has no effect in ES6, but in ES7, it will ensure the creation of indices will work the same as in ES6. It will be removed when we upgrade to ES8.
 
### Checklist

#### Functionality Checks

- [x] Code is functional.

- [x] If your changes will be tested manually, you have run `bundle update` and committed your changes to Gemfile.lock. - This is another change I suggest we skip testing on staging. This change is already in production in I14y: https://github.com/GSA/i14y/blob/29ab4e86935c7ce836ca315e8f0c53f125cf94ea/app/controllers/api/v1/collections.rb#L54
 
#### Process Checks


lorawoodford (approved):
LGTM & confirm safe to push without testing

PR #869, opened 2022-02-09
SRCH-2764 Retire PhantomJS, adopt Selenium

lorawoodford:
## Summary
Work necessary to retire PhantomJS/Poltergeist, in favor of headless chrome, including:
- Numerous minor test adjustments to address 1) the (in)visibility of many of our links and 2) the accurate text/presence of modals;
- Updating `ruby_version`s in our cirlceci config to use the new non-legacy 2.7.5 image introduced in #862 ;
- Updating circleci config to install chrome and chromedriver (see: https://circleci.com/docs/2.0/browser-testing/ and https://circleci.com/developer/orbs/orb/circleci/browser-tools); and,
- Removing references to PhantomJS in the readme.
 
### Checklist

#### Functionality Checks

- [x] Code is functional.

Though not to be tested manually, bundle update was run as part of the process of removing poltergeist.
 
#### Process Checks


features/step_definitions/web_ext_steps.rb
```diff
@@ -13,17 +13,17 @@
 
 Then /^I should see a link to "([^"]*)" with url (for|that starts with|that ends with) "([^"]*)"$/ do |name, attribute_selector, url|
   operator =
```
MothOnMars, line 15: Because the `operator` variable is no longer returning an "operator", we should rename it to something more accurate and descriptive. It will either return a regexp or a string, so something like `link_matcher` or `link_pattern`...?

features/step_definitions/web_ext_steps.rb
```diff
@@ -13,17 +13,17 @@
...
-      case attribute_selector
-        when 'that starts with' then '^='
-        when 'that ends with' then '$='
-        else '='
-      end
+    case attribute_selector
+    when 'that starts with' then /^#{Regexp.escape(url)}.*$/
+    when 'that ends with' then /.*#{Regexp.escape(url)}$/
```
MothOnMars, line 18: Because the beginning/end anchors are all that's needed, these can be simplified as:
```ruby
    when 'that starts with' then /^#{Regexp.escape(url)}/
    when 'that ends with' then /#{Regexp.escape(url)}$/
```

MothOnMars (changes requested):

Gemfile.lock
```diff
@@ -190,14 +190,14 @@ GEM
       request_store (~> 1.0)
     awesome_print (1.9.2)
     aws-eventstream (1.2.0)
-    aws-partitions (1.551.0)
-    aws-sdk-core (3.125.5)
+    aws-partitions (1.553.0)
+    aws-sdk-core (3.126.0)
```
MothOnMars, line 194: Just noting that a `bundle install` would have been better than a `bundle update`, as this is adding bumps to our production gems for a ticket that won't be manually tested. That said, the bumps are minor enough and limited enough that we can still release this w/o testing.

MothOnMars (approved):
LGTM! 🎉 

OK to merge & move to "done" w/o testing.

Gemfile.lock
lorawoodford: Ah, yea, good catch.  Thanks for pointing this out.

PR #871, opened 2022-02-10
SRCH-2564 Fix users export from super admin

lorawoodford:
## Summary
Fixes User Export failures in Super Admin by adding `affiliate_names` to the ActiveScaffold config columns.  Also:
- Consolidates various `config.actions.exclude` and `config.actions.add`;
- Minor code style, indentation, etc. changes; and,
- Adds a new `spec/system/admin/user_spec.rb`.
 
### Checklist

#### Functionality Checks

- [x] Code is functional.

#### Process Checks


spec/system/admin/user_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+        click_link('Export')
+        click_button('Export')
+      end
+
+      let(:file) { 'users.csv' }
+
+      after { FileUtils.rm_f(file) }
+
+      it 'exports a csv of users' do
+        expect { export }.not_to raise_error
```
MothOnMars, line 23: We should avoid including multiple expectations in a single spec (with some exceptions - we can discuss), so let's move the error testing to its own spec, i.e.:
```ruby
      it 'is successful' do
        expect { export }.not_to raise_error
      end

      it 'exports a csv of users' do
        export...    
```

app/controllers/admin/users_controller.rb
```diff
@@ -1,27 +1,27 @@
+# frozen_string_literal: true
+
 class Admin::UsersController < Admin::AdminController
   active_scaffold :user do |config|
-    config.actions.exclude :create, :delete
-    config.columns = %I[email first_name last_name
+    config.actions.exclude :create, :delete, :search
+    config.columns = %I[email first_name last_name affiliate_names
```
MothOnMars, line 6: This introduces a new bug, as it changes the behavior of the users index page, by adding the "Affiliate names" column:

image

We'll need to fix that and add a spec for the fix. (Note: the "Memberships" column is just a wordier version of the affiliates column, as it displays the affiliate name plus the user's email. I actually think the "Affiliate names" column is preferable, but I leave it up to you as to whether we want to chase down the desired UI behavior at this point in time.)

MothOnMars (changes requested):

app/controllers/admin/users_controller.rb
lorawoodford: Per the JIRA comment, I've proceeded with swapping out memberships for affiliate names.  This work is grouped in the final commit.

spec/system/admin/user_spec.rb
```diff
@@ -17,7 +45,10 @@
 
       let(:file) { 'users.csv' }
 
-      after { FileUtils.rm_f(file) }
+      after do
+        sleep(0.5) unless File.exist?(file)
+        FileUtils.rm_f(file) if File.exist?(file)
```
MothOnMars, line 50: This is fine for now, but eventually we could transfer the file downloading code to a helper file.

spec/system/admin/user_spec.rb
```diff
@@ -8,6 +8,34 @@
   context 'when a super admin is logged in' do
     include_context 'log in super admin'
 
+    describe 'active_scaffold list', :js do
```
MothOnMars, line 11: Minor nit - as this is a system spec, let's focus on the behavior of this page rather than the technical underpinnings, i.e. `describe 'Users list'`...`

spec/models/user_spec.rb
```diff
@@ -412,7 +412,7 @@
     it 'returns all associated affiliate display names' do
       affiliates(:power_affiliate).users << user
       affiliates(:basic_affiliate).users << user
-      expect(user.affiliate_names.split(',').sort).to eq(%w[noaa.gov nps.gov])
+      expect(user.affiliate_names.split(', ').sort).to eq(%w[noaa.gov nps.gov])
```
MothOnMars, line 415: OK as-is, but we could also simplify this spec, which obscures the fact that the method is intended to return a formatted string:
```ruby
expect(user.affiliate_names).to eq ('noaa.gov, nps.gov')
```

app/models/user.rb
```diff
@@ -117,7 +117,7 @@ def self.new_invited_by_affiliate(inviter, affiliate, attributes)
   end
 
   def affiliate_names
-    affiliates.collect(&:name).join(',')
+    affiliates.collect(&:name).join(', ')
```
MothOnMars, line 120: OK, as is, but `affiliates.pluck(:name)` would be slightly more performant

spec/system/admin/user_spec.rb
```diff
@@ -8,6 +8,34 @@
...
+          email-column
+          first_name-column
+          last_name-column
+          affiliate_names-column
+          default_affiliate-column
+          created_at-column
+          updated_at-column
+          approval_status-column
+          actions
+        ]
```
MothOnMars, line 31: Don't spend too much time refactoring (like: 5-10 mins max), but it would be preferable if we could verify the human-readable column headings as they appear on the page, i.e., "Email", "First Name", etc.)

MothOnMars (approved):
I'm approving, but I also made a few suggestions for tweaks (mostly to improve pre-existing code). Your choice if you would like to implement any, all, or none of those.

MothOnMars (approved):
LGTM

PR #907, opened 2022-03-31
SRCH-2849 Fix flaky UsersController#create tests

lorawoodford:
## Summary
- After a lot of trial and effort, it seems the intermittent failures with the UsersController#create tests stem from the `verify_recaptcha` at https://github.com/GSA/search-gov/blob/master/app/controllers/users_controller.rb#L18 returning false **only** when these tests are run after https://github.com/GSA/search-gov/blob/master/spec/system/admin/user_spec.rb
- `verify_recaptcha` should never run in a test environment (https://github.com/ambethia/recaptcha#testing)
- Ultimately the fix is just to ensure that verify_recaptcha returns true no matter what in the UsersController tests

To replicate a failing scenario against master, `rspec spec/controllers/users_controller_spec.rb spec/system/admin/user_spec.rb --seed 23764 --format documentation`.  Note the system spec will throw a deprecation warning which can be ignored for now (see: https://github.com/teamcapybara/capybara/issues/2511)
 
### Checklist

#### Functionality Checks

- [x] Code is functional.
N/A

As discussed, code climate failures are unrelated to this change.  Please accept as is and an interruption ticket will be filed to investigate those cc failures.

 N/A, test changes only

#### Process Checks

### DIFF
--- a/spec/controllers/users_controller_spec.rb
+++ b/spec/controllers/users_controller_spec.rb
@@ -14,6 +14,12 @@ describe UsersController do
   before { activate_authlogic }
 
   describe '#create' do
+    # SRCH-2849 For some reason verify_recaptcha returns false on UsersController#create if, and only
+    # if, these tests run after the js tests in `spec/system/admin/user_spec.rb`. By default, reCAPTCHA
+    # *should* be skipped in the 'test' env (https://github.com/ambethia/recaptcha#testing), so
+    # we're just explicitly returning a true here to guard against sporadic failures.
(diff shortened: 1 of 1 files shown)


MothOnMars (approved):
LGTM. Super weird! I'm dying to dig some more to understand the root cause, but this is a simple enough change, and it sounds like there's a chance we'll be able to pull the recaptcha functionality out of this repo entirely in the not too distant future. So I'm happy to move on...

PR #911, opened 2022-04-01
SRCH-2889 Update support email address

lorawoodford:
## Summary
Updates instances of search@support.digitalgov.gov to the new admin contact email of search@gsa.gov, including minor tweaks to future proof against future email address updates:
- Moving to using the single stored secret wherever possible,
- Small (hopefully inoffensive) update to `features/step_definitions/email_steps.rb` to reflect awareness of these common email addresses set in secrets.

Note, the following are *excluded* from this change (as out of the scope of this ticket):
- 12 email templates stored in the db with this old address
- 2 spec fixtures based on those templates.
Those will need to be resolved with db migrations at a later time.
 
### Checklist

#### Functionality Checks

- [x] Code is functional.

#### Process Checks


app/controllers/omniauth_callbacks_controller.rb
```diff
@@ -11,8 +11,10 @@ def login_dot_gov
     redirect_to('/login')
   rescue OmniauthError => e
     Rails.logger.error e.message
-    flash[:error] = 'Error logging in. Please reach out to' \
-                    ' search@support.digitalgov.gov if the problem persists'
+    flash[:error] = <<~MESSAGE
+      Error logging in. Please reach out to
+      #{Rails.application.secrets.organization[:support_email_address]} if the problem persists.
```
MothOnMars, line 16: Per our discussion, let's use a constant throughout these changed files, to avoid reading it out of the secrets file each time. We can either:
- move all the email constants out of app/mailers/application_mailer.rb to config/initializers/constants.rb

OR

- just set the email addresses in config/initializers/constants.rb if they are used in a variety of places in the app

Coin flip!

features/step_definitions/email_steps.rb
```diff
@@ -34,6 +34,17 @@ def current_email_address
...
+  def searchgov_addresses(address)
+    address = case address
+              when "support"
+                "#{Rails.application.secrets.organization[:support_email_address]}"
+              when "admin"
+                Rails.application.secrets.organization[:admin_email_address]
+              else
+                address
+              end
+  end
```
MothOnMars, line 47: I think we should remove this helper and stick with the hard-coded expectation in the features. This strikes me as inserting the sausage-making process into the test, whereas our integration tests should stick as closely as possible to testing the behavior and not the implementation. For example, you wouldn't write the raw expectation as: `"And #{Rails.application.secrets.organization[:admin_email_address]} should receive an email"`. We're also not referencing those common emails very much in the features, so this wouldn't save us much work if they change in the future.

config/secrets.yml.dev
```diff
@@ -97,7 +97,7 @@ dev_settings: &DEV_SETTINGS
     tos_url: https://search.gov/tos
     page_not_found_url: https://www.usa.gov/search-error
     admin_email_address: usasearchoutbound@search.gov
-    support_email_address: search@support.digitalgov.gov
```
MothOnMars, line 100: Per our discussion, I'll put in a sub-task to make this change in the cookbooks, as this file isn't used in production.

MothOnMars (changes requested):

config/secrets.yml.dev
lorawoodford: See: https://github.com/GSA/usasearch-cookbooks/pull/601

config/initializers/constants.rb
```diff
@@ -1,4 +1,8 @@
 # frozen_string_literal: true
 
-SEARCH_ENGINES = %w[BingV6 BingV7 Google SearchGov].freeze
+ADMIN_EMAIL_ADDRESS = Rails.application.secrets.organization[:admin_email_address].freeze
 DEFAULT_USER_AGENT = Rails.application.secrets.organization[:default_user_agent].freeze
+DELIVER_FROM_EMAIL_ADDRESS = 'no-reply@support.digitalgov.gov'.freeze
```
MothOnMars, line 5: You can remove the `freeze` from the strings, as the `frozen_string_literal: true` setting will freeze them. (Rubocop would normally flag this, except that we've disabled Rubocop for the `config/` directory: https://docs.rubocop.org/rubocop/cops_style.html#styleredundantfreeze)

MothOnMars (changes requested):
One minor tweak. Overall, this feels much cleaner than the original. Yay!

config/initializers/constants.rb
lorawoodford: Good reminder that rubocop doesn't run against everything.  Also added in missing newline.

MothOnMars (approved):
LGTM

MothOnMars:
LGTM

PR #924, opened 2022-04-25
SRCH-1591 strip whitespace around sitemap URL before parsing

jamesmadhoc:
## Summary
- SitemapIndexer#initialize now strips leading/trailing whitespace from its single, sitemap_url argument.
- With this change in place, giving a well formed URL with leading and/or trailing whitespace (which then makes it not well formed) to SitemapIndexer.new() will not cause an exception to be thrown, but instead allow successful creation of a SitemapIndexer object whose uri method returns a value, that when coerced to a string, equals the original argument to new() but with any leading or trailing whitespace removed.  (This observation is the basis of a new spec that is also part of this PR.)
 
### Checklist

#### Functionality Checks

- [x] Code is functional.

#### Process Checks


jamesmadhoc:
Oops -- Rubocop disliked string concatenation in the rspec, so turned it into interpolation as suggested.

And it's the little bit of rspec change that's the only thing remotely interesting about this PR!

jamesmadhoc:
Thanks for the review, @lorawoodford .  Updated the Summary text, ran "bundle update" and committed/pushed the updated Gemfile.lock file.  Please take a look at the result.

lorawoodford (approved):
LGTM

PR #928, opened 2022-05-04
SRCH-2979 reduce "size" of LowCtrQuery aggregation

MothOnMars:
## Summary
This PR reduces the `size` of the `LowCtrQuery` aggregation to resolve the following deprecation warning:

> This aggregation creates too many buckets (10001) and will throw an error in future versions. You should update the [search.max_buckets] cluster setting or use the [composite] aggregation to paginate all buckets in multiple requests.

The current `size` setting of 100,000 buckets is much larger than we need for that query, which is used to display analytics data in the GUI, as well as for the LowCtrQuery watchers. I'm dialing it down to 5000, which should be more than enough for our existing use cases.
 
### Checklist

#### Functionality Checks

- [x] Code is functional.

#### Process Checks


spec/models/logstash_queries/watcher_low_ctr_query_spec.rb
```diff
@@ -28,50 +28,50 @@
           "filter": [
             {
               "term": {
-                "params.affiliate": 'affiliate_name'
```
MothOnMars, line 31: The logstash query specs are an exception to our normal Rubocop rule that enforces single over double quotes. Double-quotes are preferable in these specs as they allow us to copy and paste directly into Kibana for development and troubleshooting.

jamesmadhoc: Didn't remember that ruby allows underscores in number literals -- can completely understand that that wouldn't play nicely with JSON.  Didn't know the latter dislikes single quotes, either.

jamesmadhoc (approved):
Curious about how you arrived at 5000, a 20X reduction in that size.

Other than that, did a little find/greppery and seems you found all the locations for the change, so...

LGTM

MothOnMars:
Good q - the limit needed to be below 10k, as that's the default `max_buckets` in ES 7. However, the limit actually needs to be below the `max_buckets`, as ES pulls additional results form each shard, and then processes the combined result set. See:
- https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-terms-aggregation.html#search-aggregations-bucket-terms-aggregation-shard-size
- https://stackoverflow.com/a/57464229

5k is still somewhat arbitrary, but it gets us well below the threshold, and is still returning more data than we actually display to the user.

PR #944, opened 2022-06-22
SRCH-3101 specify minimum_should_match in ElasticBlendedQuery

MothOnMars:
## Summary
This PR resolves the following deprecation warning from Elasticsearch, generated by our `ElasticBlendedQuery`:

> "Should clauses in the filter context will no longer automatically set the minimum should match to 1 in the next major version. You should group them in a [filter] clause or explicitly set [minimum_should_match] to 1 to restore this behavior in the next major version." 

The logic of that query specifies that condition A must be met, and *either* condition B or C "should" be met. Elasticsearch 7 requires that we specify `minimum_should_match` for the either/or conditions; otherwise, neither condition is required to be met. Our tests were all passing because we weren't testing a case where condition A was met, but neither B or C were met. I have added a cucumber test for that scenario. You can see that the scenario passes on Elasticsearch 6 but fails on Elasticsearch 7:
https://app.circleci.com/pipelines/github/MothOnMars/search-gov/2257/workflows/580f1af7-2db8-44a8-99b6-37f2be2d1b07

 
### Checklist

#### Functionality Checks

- [x] Code is functional.

- [x] If your changes will be tested manually, you have run `bundle update` and committed your changes to Gemfile.lock. - N/A. We have made a similar change in previous PRs without manual testing.
 
- [x] If your target branch is NOT `main`, specify the reason: - N/A
 
#### Process Checks

### DIFF
--- a/app/models/custom_index_queries/elastic_blended_query.rb
+++ b/app/models/custom_index_queries/elastic_blended_query.rb
@@ -52,6 +52,8 @@ class ElasticBlendedQuery < ElasticTextFilterByPublishedAtQuery
           json.child! { json.term { json.affiliate_id @affiliate_id } }
           json.child! { json.terms { json.rss_feed_url_id @rss_feed_url_ids } }
         end
+
+        json.minimum_should_match 1
       end
     end
   end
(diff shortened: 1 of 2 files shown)


MothOnMars:
@danswick , hold that thought. It's nagging me that we should have a unit test for this. Closing this up temporarily to add one...

PR #965, opened 2022-07-15
SRCH-3154 searchgov_domain.check_status should not raise errors

jamesmadhoc:
## Summary
- SearchgovDomain#check_status now returns a non-OK_STATUS rather than raising an exception when its domain is unreachable.
- Logic is still a bit tortured; I'm open to suggestions for improvement.
 
### Checklist
 
#### Functionality Checks

- :warning: You have run `bundle update` and committed your changes to Gemfile.lock; depending on this operation having been done by the recently merged [SRCH-3224](https://cm-jira.usa.gov/browse/SRCH-3224)
 
#### Process Checks

### DIFF
--- a/app/models/searchgov_domain.rb
+++ b/app/models/searchgov_domain.rb
@@ -42,7 +42,7 @@ class SearchgovDomain < ApplicationRecord
   end
 
   def available?
-    !!((status || check_status) =~ /^200\b/)
+    (status || check_status) == OK_STATUS
   end
 
   def check_status
@@ -90,7 +90,7 @@ class SearchgovDomain < ApplicationRecord
 
   def failed_response(err)
     update(status: err.message.strip)
-    Rails.logger.warn "#{domain} response error url: #{url} error: #{status}"
+    Rails.logger.error "#{domain} response error url: #{url} error: #{status}"
     nil
   end
 
--- a/app/models/searchgov_domain.rb
+++ b/app/models/searchgov_domain.rb
@@ -21,14 +21,13 @@ class SearchgovDomain < ApplicationRecord
   has_many :sitemaps, dependent: :destroy
 
   attr_readonly :domain
+  attr_reader :response
 
   scope :ok, -> { where(status: OK_STATUS) }
   scope :not_ok, -> { where.not(status: OK_STATUS) }
 
   def delay
-    @delay ||= begin
-      robotex.delay("http://#{domain}/") || 1
-    end
+    @delay ||= (robotex.delay("http://#{domain}/") || 1)
   end
 
   def index_urls
@@ -43,22 +42,13 @@ class SearchgovDomain < ApplicationRecord
   end
 
   def available?
-    /^200\b/ === (status || check_status)
+    !!((status || check_status) =~ /^200\b/)
   end
 
   def check_status
-    record_response_if_non_nil
-
-    save if changed?
+    fetch_response
+    record_response if response
     status
-  end
-
-  def record_response_if_non_nil
-    return if response.nil?
-
-    self.status = response.status
-    self.scheme = response.uri.scheme
-    self.canonical_domain = host unless domain == host
   end
 
   aasm column: 'activity' do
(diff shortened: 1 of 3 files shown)


MothOnMars (changes requested):
@jamesmadhoc , the minimal changes I would make to this are:
- rename `record_response_if_non_nil` to simply `record_response` (the guard clause makes the flow very clear)
- make the `record_response` method private, as there should be no need to call that method directly
- move the `save if changed?` line to `record_response`, because that's where the save is logically relevant

I'm trying not to overthink this, as we're likely to do significant refactoring for the JS indexing project. However, I played with the code a bit to make it more readable (IMHO - 'readable' is often in the eye of the beholder). You can take a look at my tweaked version, and take it or leave it: 
https://github.com/MothOnMars/search-gov/blob/7bd3417564f46e5c3ec06bf4d75d944f036fad12/app/models/searchgov_domain.rb

The `check_status` method now tells a procedural story, returning the `status` string at the end:
```ruby
  def check_status
    get_response
    record_response if response
    status
  end
 ```
 That said, the procedural aspect of it is part of the reason it should be refactored into a service class. But that doesn't have to happen today. So take a look at my changes, but don't spend more than about 10 minutes on it. Let's focus on getting this out the door, with the awareness that any lingering ickiness will be refactored shortly.

**Comment by jamesmadhoc** [comment]
Hi, @MothOnMars -- Your suggestion looks good to me, but it has a behavioral change that I'm curious about:

 ```
 def get_response
    @response = begin
      Retriable.retriable(base_interval: delay) do
        DocumentFetchLogger.new(url, 'searchgov_domain').log
        HTTP.headers(user_agent: DEFAULT_USER_AGENT).
          timeout(connect: 20, read: 60).follow.get(url)
      end
    rescue StandardError => e
      failed_response(e)
    end
  end
```
The existing SearchgovDomain#response makes the assignment to the instance variable, on success, only once though:

```
def response
  @response ||= begin
  #...
```
I don't really like the "if it succeeds once, never check again" semantic, but was hesitant to change it to "always check, regardless of past success/failure".  If I do change it to "always", should there be a spec for it?

jamesmadhoc:
Hello, @MothOnMars -- went with your suggested implementation after all, but with a few changes to mollify rubocop.  The 'status' update is not an issue after all (because the failed case is covered by ``update(status:...'' in failed_response).  The ``always check, regardless of past success/failure'' semantic passes specs and is what I prefer anyway, so it stays in.  Finally, one of the rubocop fixes -- getting rid of the '===' operator in available? -- *is* rather egregious, but it exactly preserves previous behavior, returning booleans rather than truthy/falsey values, which would require the specs to change -- and changing specs blindly gives me the creeps.

app/models/searchgov_domain.rb
```diff
@@ -43,15 +42,12 @@ def index_sitemaps
   end
 
   def available?
-    /^200\b/ === (status || check_status)
+    !!((status || check_status) =~ /^200\b/)
```
MothOnMars, line 45: @jamesmadhoc , this can be improved with the new-ish `.match?` method:
```ruby
(status || check_status).match?(/^200\b/)
```
However, we can simplify it even further as:
```ruby
(status || check_status) == OK_STATUS
```
The pattern matching is a remnant of the early days of Garfield, when I wasn't sure if there would ever be variants of the `200 OK` string. 5 years worth of data later, I'm not concerned about that:
```ruby
> pp SearchgovDomain.pluck(:status).uniq.sort
["200 OK",
 "403 Forbidden",
 "404 Not Found",
 "SSL_connect returned=1 errno=0 state=error: certificate verify failed (Hostname mismatch)",
 "SSL_connect returned=1 errno=0 state=error: certificate verify failed (unable to get local issuer certificate)",
 "execution expired",
 "failed to connect: Connection refused - connect(2) for \"clu-in.org\" port 443",
 "failed to connect: Network is unreachable - connect(2) for \"clintonwhitehouse1.archives.gov\" port 443",
 "failed to connect: Network is unreachable - connect(2) for \"clintonwhitehouse2.archives.gov\" port 443",
 "failed to connect: getaddrinfo: Name or service not known"]
 ```

MothOnMars (changes requested):
see comments

jamesmadhoc:
Made the suggested changes -- please re-review.

jamesmadhoc:
Hello, @danswick -- sending this your way to do the final review; already made the changes last requested by Martha and got them through circleci OK (after an initial intermittent failure of unrelated specs).  Note that there are a bunch of little rubocop-motivated changes sprinkled throughout.

danswick (approved):
Following the comment trail, these changes all look clear to me, @jamesmadhoc 👍

PR #966, opened 2022-07-15
[SRCH-505] remove any database tables or columns related to Instagram

danswick:
## Summary
 
### Checklist
 
#### Functionality Checks

- [ ] You have run `bundle update` and committed your changes to Gemfile.lock.
 
- [ ] You have squashed your commits into a single commit (exceptions: your PR includes commits with formatting-only changes, such as required by Rubocop or Cookstyle, or if this is a feature branch that includes multiple commits).
 
#### Process Checks

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2022_05_12_200543) do
+ActiveRecord::Schema.define(version: 2022_07_15_232531) do
 
   create_table "active_storage_attachments", charset: "utf8mb4", collation: "utf8mb4_unicode_ci", force: :cascade do |t|
     t.string "name", null: false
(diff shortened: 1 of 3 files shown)


db/schema.rb
```diff
@@ -780,7 +768,6 @@
     t.index ["id", "imported_at"], name: "index_youtube_profiles_on_id_and_imported_at"
   end
 
-  add_foreign_key "active_storage_attachments", "active_storage_blobs", column: "blob_id"
```
lorawoodford, line 783: @danswick This seems like an accidental/unrelated change.  Try the following:

1. Rollback this file:
`git checkout main -- db/schema.rb`
2. `rails db:drop`
3. `rails db:create`
4. `rails db:setup`
5.`rails db:migrate`

And then check the changes in the new `db/schema.rb`.  Is this change there?  (For me it isn't).

lorawoodford (changes requested):
@danswick One comment left in `db/schema.rb`.

db/schema.rb
danswick: Thank you for catching that! I'm not sure how that change even happened, but it should be rolled back now.

danswick:
@lorawoodford this should be ready for re-review whenever you're available. The "request re-review" UI isn't present, possibly because there are so many commits since your last review, so tagging you here instead.

lorawoodford (changes requested):
@danswick I've left two inline comments, but long story short just a couple tweaks to ensure that if these migrations were ever reverted we'd be left in the same state we were prior to the table removals.

A good gut check for this is to revert either/both migrations with:

`rails db:migrate:down VERSION=20220715232531` and/or `rails db:migrate:down VERSION=20220715232211`

and then compare the resulting `schema.rb` to the one currently in main.  If they aren't matching, then the down migration isn't entirely reverting the change.

Thanks!

danswick:
Thank you for the suggestions, @lorawoodford! I totally missed that detail and went ahead and incorporated both of your suggested changes. Should be ready for another review whenever you're available.

lorawoodford (approved):
LGTM

PR #970, opened 2022-07-20
SRCH-3158 update parsing logic for tags field

lorawoodford:
## Summary
Primarily serves to update the parsing logic for `keywords` for search-gov crawled urls.
 
- Previously, keywords originated from either the first `metadata['keywords']` or, if not present, the first `dublin_core_data['dc.subject']`; now, all values in any/all of the following fields are gathered, deduplicated, and combined in the final keywords string:
    - dublin_core_data['dc.subject'],
    - dcterms_data['dcterms.subject'],
    - dcterms_data['dcterms.keywords'],
    - metadata['keywords']&.join(', '),
    - metadata['article:tag']&.join(', '),
    - metadata['article:section'];
- Because we wanted to support parallel descriptions (more in inline comment), additional tweaks were made to existing/non-keyword methods;
- The first commit includes required/necessary spec changes, as well as an new spec in `spec/models/application_document_spec.rb` to support current behavior (seems an oversight there wasn't a spec specifically for non-html keyword parsing to begin with);
- A second commit includes low-risk rubocop changes to a couple spec files just to cut down on the noise.  Larger refactoring to appease all rubocop rules has been deferred as it is out of scope for this work.
 
### Checklist
 
#### Functionality Checks

:warning: Automated checks pass. If Code Climate checks do not pass, explain reason for failures:
Many, but not all preexisting code style issues in `spec/models/searchgov_url_spec.rb` have been resolved.  5 instances of `too many memoized helpers` remain to be dealt with another day.
 
#### Process Checks

### DIFF
--- a/app/models/html_document.rb
+++ b/app/models/html_document.rb
@@ -97,9 +97,9 @@ class HtmlDocument < WebDocument
     [dublin_core_data['dc.subject'],
      dcterms_data['dcterms.subject'],
      dcterms_data['dcterms.keywords'],
-     metadata['keywords']&.join(', '),
-     metadata['article:tag']&.join(', '),
-     metadata['article:section']&.join(', ')]
+     metadata['keywords'],
+     metadata['article:tag'],
+     metadata['article:section']].map { |k| k&.join(', ') }
   end
 
   def extract_language
@@ -107,7 +107,7 @@ class HtmlDocument < WebDocument
   end
 
   def extract_created
-    metadata['article:published_time']&.first  || dublin_core_date || dcterms_date
+    metadata['article:published_time']&.first || dublin_core_date || dcterms_date
   end
 
   def extract_changed
--- a/app/models/html_document.rb
+++ b/app/models/html_document.rb
@@ -4,25 +4,25 @@ class HtmlDocument < WebDocument
   include RobotsTaggable
 
   def title
-    titles = [metadata['og:title'], html.title.try(:strip)]
+    titles = [metadata['og:title']&.first, html.title.try(:strip)]
     titles.compact_blank!
     titles.blank? ? url : titles.max_by(&:length)
   end
 
   def description
-    metadata['og:description'] || metadata['description'] || dublin_core_data['dc.description']
+    metadata['og:description']&.first || metadata['description']&.first || dublin_core_data['dc.description']&.first
   end
 
   def keywords
-    metadata['keywords'] || dublin_core_data['dc.subject']
+    extract_keywords.uniq.compact.join(', ')
   end
 
   def audience
-    dcterms_data['dcterms.audience']
+    dcterms_data['dcterms.audience']&.first
   end
 
   def image_url
-    metadata['og:image']
+    metadata['og:image']&.first
   end
 
   def content_type
@@ -32,7 +32,7 @@ class HtmlDocument < WebDocument
   def searchgov_custom(number)
     return if !number.is_a?(Integer) || !number.between?(1, 3)
 
-    metadata["searchgov_custom#{number}"]
+    metadata["searchgov_custom#{number}"]&.first
   end
 
   # Returns client-side redirect url
@@ -87,10 +87,19 @@ class HtmlDocument < WebDocument
     metadata = {}
     meta_nodes = html.xpath('//meta')
     meta_nodes.each do |node|
-      (metadata[node['name'].downcase] = node['content']) if node['name']
-      (metadata[node['property'].downcase] = node['content']) if node['property']
+      property = node['name'] || node['property']
+      (metadata[property.downcase] ||= []) << node['content'] unless property.nil?
     end
     metadata
+  end
+
+  def extract_keywords
+    [dublin_core_data['dc.subject'],
+     dcterms_data['dcterms.subject'],
+     dcterms_data['dcterms.keywords'],
+     metadata['keywords']&.join(', '),
+     metadata['article:tag']&.join(', '),
+     metadata['article:section']&.join(', ')]
   end
 
   def extract_language
@@ -98,15 +107,15 @@ class HtmlDocument < WebDocument
   end
 
   def extract_created
-    metadata['article:published_time'] || dublin_core_date || dcterms_date
+    metadata['article:published_time']&.first  || dublin_core_date || dcterms_date
   end
 
   def extract_changed
-    metadata['article:modified_time']
+    metadata['article:modified_time']&.first
   end
 
   def robots_directives
-    (metadata['robots'] || '').downcase.split(',').map(&:strip)
+    (metadata['robots']&.first || '').downcase.split(',').map(&:strip)
   end
 
   def main_html
@@ -116,11 +125,11 @@ class HtmlDocument < WebDocument
   end
 
   def dublin_core_date
-    dublin_core_data['dc.date'] || dublin_core_data['dc.date.created']
+    dublin_core_data['dc.date']&.first || dublin_core_data['dc.date.created']&.first
   end
 
   def dcterms_date
-    dcterms_data['dcterms.created']
+    dcterms_data['dcterms.created']&.first
   end
 
   def dublin_core_data
--- a/spec/models/html_document_spec.rb
+++ b/spec/models/html_document_spec.rb
@@ -318,9 +318,105 @@ describe HtmlDocument do
     subject(:keywords) { html_document.keywords }
 
     context 'when a Dublin Core subject is available' do
+      let(:raw_document) do
+        <<~HTML
+          <html lang="en">
+            <head>
+              <title>My Title</title>
+              <meta name="dc.subject" content="One DC Subject, Another DC Subject"/>
+            </head>
+          </html>
+        HTML
+      end
+
+      it { is_expected.to eq 'One DC Subject, Another DC Subject' }
+    end
+
+    context 'when a dcterms subject is available' do
+      let(:raw_document) do
+        <<~HTML
+          <html lang="en">
+            <head>
+              <title>My Title</title>
+              <meta name="dcterms.subject" content="One DCTerms Subject, Another DCTerms Subject"/>
+            </head>
+          </html>
+        HTML
+      end
+
+      it { is_expected.to eq 'One DCTerms Subject, Another DCTerms Subject' }
+    end
+
+    context 'when a dcterms keywords is available' do
+      let(:raw_document) do
+        <<~HTML
+          <html lang="en">
+            <head>
+              <title>My Title</title>
+              <meta name="dcterms.keywords" content="One DCTerms Keyword, Another DCTerms Keyword"/>
+            </head>
+          </html>
+        HTML
+      end
+
+      it { is_expected.to eq 'One DCTerms Keyword, Another DCTerms Keyword' }
+    end
+
+    context 'when all dc keyword sources are available' do
       let(:raw_document) { doc_with_dc_data }
 
-      it { is_expected.to eq 'One Subject, Another Subject' }
+      it { is_expected.to include('One DC Subject') }
+      it { is_expected.to include('One DCTerms Subject') }
+      it { is_expected.to include('One DCTerms Keyword') }
+      it { is_expected.to include('Another DC Subject') }
+      it { is_expected.to include('Another DCTerms Subject') }
+      it { is_expected.to include('Another DCTerms Keyword') }
+    end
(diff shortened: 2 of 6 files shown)


app/models/html_document.rb
```diff
@@ -87,26 +87,35 @@ def extract_metadata
     metadata = {}
     meta_nodes = html.xpath('//meta')
     meta_nodes.each do |node|
-      (metadata[node['name'].downcase] = node['content']) if node['name']
-      (metadata[node['property'].downcase] = node['content']) if node['property']
+      property = node['name'] || node['property']
```
lorawoodford, line 90: This is the biggest change, which impacts both the new tags logic, as well as existing metadata parsing for other fields.  At present, `extract_metadata` cannot handle parallel description.  So, something completely valid like:

```
<meta property="article:tag" content="NIAMS">
<meta property="article:tag" content="bone research">
```

Would only result in a tags value of "bone research" (dropping/overwriting "NIAMS").  The `extract_metadata` method has been rewritten to construct an array of for duplicated names or properties.

While the `&.first`s peppered throughout to convert these arrays back to strings aren't great, it at least makes more explicit what we've always been doing - taking only one value even if multiple values are provided.  And, specifically in the case of tags, returning arrays allows us to collect all values from all tag source fields, dedupe, and preserve all values - whether presented in one meta field or sprinkled across many.

Feedback/questions welcome.

lorawoodford:
@danswick Assigning this one to you as you're going to have to approve the CC failures either way.  If your plate is too full, please reassign to whoever.

app/models/html_document.rb
danswick: Thanks for explaining this! Just to make sure I'm understanding: the new behavior you've implemented breaks out the way we treat _keywords_, but leaves other types of metadata the same?

1. Extracts and stores _all_ keywords as a deduped array. 
2. Retains the old behavior (extracting only the first of potentially many values) for all other metadata.

Is that right?

danswick (approved):
Thanks for the detailed description and inline notes, @lorawoodford! I left a couple of questions / suggestions for your consideration inline, but these changes all look good to me either way!

Let me know if you want to discuss anything, otherwise I can hop back in to approve codeclimate's issues whenever you're ready to merge.

app/models/html_document.rb
```diff
@@ -4,25 +4,25 @@ class HtmlDocument < WebDocument
   include RobotsTaggable
 
   def title
-    titles = [metadata['og:title'], html.title.try(:strip)]
+    titles = [metadata['og:title']&.first, html.title.try(:strip)]
```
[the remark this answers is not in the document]
lorawoodford: IMO the code itself is self explanatory.

app/models/html_document.rb
```diff
@@ -87,26 +87,35 @@ def extract_metadata
...
   end
 
+  def extract_keywords
+    [dublin_core_data['dc.subject'],
+     dcterms_data['dcterms.subject'],
+     dcterms_data['dcterms.keywords'],
+     metadata['keywords']&.join(', '),
+     metadata['article:tag']&.join(', '),
+     metadata['article:section']&.join(', ')]
+  end
```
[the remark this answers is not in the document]
lorawoodford: When any of these values do not exist, it just constructs an array of `nil`s.  Then, when it hits the keywords method, the `.compact` dumps all the `nil`s.  In the console:

```
[9] pry(main)> url = 'https://search.gov'
=> "https://search.gov"
[10] pry(main)> doc = HtmlDocument.new(document: open(url).read, url:url)
=> #<HtmlDocument:0x00007fa4a17fbca8 [...]
[11] pry(main)> doc.send(:extract_keywords)
=> [nil, nil, nil, nil, nil, nil]
[12] pry(main)> doc.keywords
=> ""
```

app/models/html_document.rb
lorawoodford: Yes, so the higher level `extract_metadata` method now always returns an array for _everything_, but the lower level, metadata-specific methods (where the `&.first`s have been added) all directly convert those arrays (mostly single-item arrays) back to a string.

On current main against the url `https://search.gov` the `extract_metadata` method would return:

```
{"handheldfriendly"=>"True",
 "mobileoptimized"=>"320",
 "viewport"=>"width=device-width, initial-scale=1.0",
 "msapplication-tilecolor"=>"#da532c",
 "msapplication-config"=>"https://search.gov/assets/img/favicons/browserconfig.xml",
 "theme-color"=>"#ffffff",
 "generator"=>"Jekyll v4.2.2",
 "og:title"=>"Search.gov Home",
 "og:locale"=>"en_US",
 "description"=>
  "Search.gov powers thousands of federal government search boxes. Use it to improve the public’s search experience on your agency’s site.",
 "og:description"=>
  "Search.gov powers thousands of federal government search boxes. Use it to improve the public’s search experience on your agency’s site.",
 "og:url"=>"https://search.gov/",
 "og:site_name"=>"Search.gov",
 "og:type"=>"website",
 "twitter:card"=>"summary",
 "twitter:title"=>"Search.gov Home"}
```

But, on this branch, we get:

```
{"handheldfriendly"=>["True"],
 "mobileoptimized"=>["320"],
 "viewport"=>["width=device-width, initial-scale=1.0"],
 "msapplication-tilecolor"=>["#da532c"],
 "msapplication-config"=>["https://search.gov/assets/img/favicons/browserconfig.xml"],
 "theme-color"=>["#ffffff"],
 "generator"=>["Jekyll v4.2.2"],
 "og:title"=>["Search.gov Home"],
 "og:locale"=>["en_US"],
 "description"=>
  ["Search.gov powers thousands of federal government search boxes. Use it to improve the public’s search experience on your agency’s site."],
 "og:description"=>
  ["Search.gov powers thousands of federal government search boxes. Use it to improve the public’s search experience on your agency’s site."],
 "og:url"=>["https://search.gov/"],
 "og:site_name"=>["Search.gov"],
 "og:type"=>["website"],
 "twitter:card"=>["summary"],
 "twitter:title"=>["Search.gov Home"]}
 ```
 
 But, since I've added in array handling to the specific metadata methods (e.g. `description` or `title`), both branches ultimately return a string. (So, to point 2 in your questions, yes, the old behavior is retained).
 
 Main:
 ```
 [5] pry(main)> doc.description
=> "Search.gov powers thousands of federal government search boxes. Use it to improve the public’s search experience on your agency’s site."
[6] pry(main)> doc.title
=> "Search.gov Home | Search.gov"
 ```
 
 This branch:
 ```
 [3] pry(main)> doc.description
=> "Search.gov powers thousands of federal government search boxes. Use it to improve the public’s search experience on your agency’s site."
[4] pry(main)> doc.title
=> "Search.gov Home | Search.gov"
 ```

File: app/models/html_document.rb (hunk 3 above)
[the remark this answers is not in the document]
Comment (danswick) [reply app/models/html_document.rb:103]: Ah, I see! Thanks for explaining that.

**Comment by lorawoodford** [comment]
@danswick if all of the above makes sense, can you approve (both CC and the overall set of code changes) so I can move into acceptance testing?  Thanks.

**Comment by danswick** [comment]
@lorawoodford apologies! I missed the CC approval bit, so this was still appearing as ❌, but should be all 🍏 now.

**Comment by lorawoodford** [comment]
@danswick For when you return -- Upon setting up acceptance testing I noticed undesirable behavior in two scenarios:

1. when duplicate keywords appeared in an array as well as in parallel description (e.g. a list of comma-separated keywords in a meta keywords tag, and a parallel listing of the same separate keywords in article:tags tags), and;
2. when mashing together keywords from sources handled as an array (e.g. dc keywords) and those handled as concatenated strings (e.g. meta keywords).

If the above descriptions are hard to follow, take a peek at the two new tests added in the new commit, and hopefully it'll become clearer.  Also in that commit I've modified the `extract_keywords` method to resolve this issue.

Sorry for the double review, and thanks for your patience.

File: spec/models/html_document_spec.rb [hunk 4]
```diff
@@ -270,9 +270,142 @@
...
+    context 'when meta keywords and article keywords are available' do
+      let(:raw_document) { read_fixture_file('/html/page_with_og_metadata.html') }
+
+      it { is_expected.to include('this') }
+      it { is_expected.to include('that') }
+      it { is_expected.to include('the other') }
+      it { is_expected.to include('thing') }
+    end
+
+    context 'when parallel keywords duplicate a list of keywords' do
```
danswick, line 373: Are there any scenarios where we would need to dedupe keywords that aren't perfectly replicated in both a comma-separated list and a set of separate tags? 

I believe something like the following (only two tags in the comma-separated list) would fail, but not sure if it's a likely scenario:

```html
<meta name="keywords" content="tag 1, tag 3">
<meta property="article:section" content="tag 1" />
<meta property="article:section" content="tag 2" />
<meta property="article:section" content="tag 3" />
```

danswick (commented):
This looks like a great catch, @lorawoodford! I have one follow-up question noted inline, but I may be misunderstanding the nature of duplicated keywords.

spec/models/html_document_spec.rb
lorawoodford: Yes, you're correct, that would keep the duplicated "tag 3" around:

```
[1] pry(main)> doc =         <<~HTML
          <html lang="en">
            <head>
              <title>My Title</title>
              <meta name="keywords" content="tag 1, tag 3">
              <meta property="article:section" content="tag 1" />
              <meta property="article:section" content="tag 2" />
              <meta property="article:section" content="tag 3" />
            </head>
          </html>
        HTML
=> "<html lang=\"en\">\n  <head>\n    <title>My Title</title>\n    <meta name=\"keywords\" content=\"tag 1, tag 3\">\n    <meta property=\"article:section\" content=\"tag 1\" />\n    <meta property=\"article:section\" content=\"tag 2\" />\n    <meta property=\"article:section\" content=\"tag 3\" />\n  </head>\n</html>\n"
[2] pry(main)> doc = HtmlDocument.new(document: doc, url:'www.example.com')
=> #<HtmlDocument:0x00007fe3d72263f8
 @document=
  "<html lang=\"en\">\n  <head>\n    <title>My Title</title>\n    <meta name=\"keywords\" content=\"tag 1, tag 3\">\n    <meta property=\"article:section\" content=\"tag 1\" />\n    <meta property=\"article:section\" content=\"tag 2\" />\n    <meta property=\"article:section\" content=\"tag 3\" />\n  </head>\n</html>\n",
 @url="www.example.com">
[3] pry(main)> doc.send(:extract_keywords)
=> [nil, nil, nil, "tag 1, tag 3", nil, "tag 1, tag 2, tag 3"]
[4] pry(main)> doc.keywords
=> "tag 1, tag 3, tag 1, tag 2, tag 3"
```

I'm inclined to keep this as is, but welcome feedback.  First, because I agree that we don't know how often (or if ever) we will encounter this.  Across my list of test sites the most common pattern I saw was exact duplication of content across one or more metadata elements (aka https://www.niams.nih.gov/ where meta keywords directly duplicate parallel article:tag elements).  I think this is probably a safe assumption given the wide adoption of content management systems, but it is definitely an _assumption_.  But, also, taking the opposite approach - treating every word offset by commas as a truly unique entity to be deduped - runs the risk of improperly deleting semantically different content (e.g. a list of one element containing "Smith, John" and another containing "Smythe, John" would drop one of the Johns).  Whereas, the risk with the current approach is merely improperly duplicating things (but not deleting things).  

But, this conversation is still helpful because it does raise to my awareness another future pitfall with maintaining the current system of generating tags as a comma-separated list.  As we get deeper into the tags indexing and, more importantly, SERP design, we'll need to be aware that phrases like my example above "Smith, John" are, unfortunately, going to at this particular moment be treated as two different keywords.  Or maybe more likely for the type of content we see, geographic tags like "Washington, D.C." or "Baltimore, MD").

danswick: Oooh, these are very interesting scenarios! In that case, I'm inclined to agree that unintentional deduping seems worse than leaving some duplicate tags in. There also doesn't seem to be a risk, other than requiring infinitesimally more storage, to retaining a couple of duplicate tags for situations like these.

danswick (approved):
Thanks for the explanation re: duplicate tags. @lorawoodford. This looks good to me!

lorawoodford:
@danswick Hopefully last re-review on this.  Merge conflicts in gemfile resolved.

danswick (approved):
Conflict resolutions look good to me. Thanks, @lorawoodford!

PR #974, opened 2022-07-27
SRCH-2893 use Tika service instead of jar when parsing indexed documents

jamesmadhoc:
## Summary
- Core change is to make the IndexedDocument#parse_file method use the Tika service already used in Garfield (in ApplicationDocument, which is used by SearchgovUrl#parse_document); complementary changes were made to the IndexedDocumentFetcher job (which confusingly isn't named "...Job" and whose source is in app/models/ rather than app/jobs/) and to the specs for both of those and for IndexedDocumentsController.  **As of 2 August, I've confirmed that the manual test works.**
- This is *old* ODIE code, and so it's a rubocop minefield.  I started to clean it up but soon realized that some nontrivial refactoring would be required.  As hinted in the "Background/Notes" section of [SRCH-2893](https://cm-jira.usa.gov/browse/SRCH-2893), refactoring is not the purpose of this ticket, so I didn't do it.  Hence, you'll find one codeclimate offense (an 'expect' in a 'before' hook, which was already there but which I altered).
 
### Checklist
 
#### Functionality Checks

- [x] Automated checks pass. If Code Climate checks do not pass, explain reason for failures:  (1) Failure at app/models/indexed_document.rb:106, "Avoid using rescue in its modifier form", existed beforehand, and replacing it with begin/rescue/end construct that then has a 'new' "raise IndexedDocumentError.new(EMPTY_BODY_STATUS)" results in an error to the effect of "provide error message for new exception", presumably because code climate doesn't recognize that a constant evaluating to a string literal is a message.  (2) Failure at spec/controllers/sites/indexed_documents_controller_spec.rb:47 is an "expect" that replaces an older, Resque-specific one that was also in the same "before" hook.
 
#### Process Checks


jamesmadhoc:
Hello, @lorawoodford -- if you haven't started this review, then please do not -- I need to make some more changes, as the. manual testing failed.  Sorry about that.

lorawoodford:
Thanks @jamesmadhoc .  I actually have spent a large part of the day on this because I was seeing issues, so good to know your testing has raised the same.  Happy to help next week, but yes, I have been working away.

jamesmadhoc:
Hi, @lorawoodford -- this code passes auto and manual tests, now; still has one codeclimate violation, as mentioned in the summary above.

jamesmadhoc:
Hi, @lorawoodford  -- incorporating the changes you had on your fork/branch with only one little tweak at spec/models/indexed_document_spec.rb:281 to quell a rubocop complaint.  Change (1) did not raise rubocop's ire, anyway -- which I thought it would, which is why I had that code-smelly ``rescue TikaError'' in the first place.  Change (3) goes right along with this, of course.  The really interesting one was the the use of ``at_least(:once)'' in (2), which I *would* like to discuss tomorrow, if we can squeeze it in.  I tried to do something similar to what you did -- repeating the ``post :create ...'' after the before hook's block, but without success.  It was at that point I just put the expect around the first one and called it a day.

lorawoodford:
@jamesmadhoc Sure, let's talk after standup today.

lorawoodford (approved):
LGTM

PR #975, opened 2022-07-28
SRCH-1599 searchgov domain activity set to idle when domain is unreac…

krbhavith:
## Summary
- calling `done_indexing!` directly since the there is no response code when domain  is not found so that it would move it to idle, when ever standard exception happens on `activity == "indexing"`

### Checklist
 
#### Functionality Checks

#### Process Checks


app/models/searchgov_domain.rb
```diff
@@ -84,6 +84,7 @@ def fetch_response
           timeout(connect: 20, read: 60).follow.get(url)
       end
     rescue StandardError => e
+      done_indexing! if activity == 'indexing'
```
lorawoodford, line 87: This is the right action, but upon looking at the mocked up test Martha created here (https://github.com/GSA/search-gov/compare/main...MothOnMars:search-gov:set_to_idle_1599) this just needs to be moved up into the `check_status` method on ln 48.  So something like:

```
  def check_status
    fetch_response
    if response
      record_response
    elsif activity == 'indexing'
      done_indexing!
    end
    status
  end
  ```
  
(If you want to refactor that into a separate cleaner method or tweak in whatever way, feel free.  Just proof of concept.)

I didn't dig deeply into _why_ we want `check_status` to be the trigger, but trust that Martha had a reason for this when writing up the ticket so feels safer to roll with it.

File: spec/models/searchgov_domain_spec.rb [hunk 2]
```diff
@@ -298,6 +298,17 @@
       end
     end
 
+    context 'when the domain is being indexed' do
```
lorawoodford, line 301: This can be replaced with the spec in Martha's branch (https://github.com/GSA/search-gov/compare/main...MothOnMars:search-gov:set_to_idle_1599) with one minor change.  It appears the defined domain in that branch doesn't error out, so I'd just recommend updating the domain to something that will error out, like:

```
      let(:searchgov_domain) do
        described_class.create!(domain: 'test.gov', activity: 'indexing')
      end
 ```

lorawoodford (changes requested):
@krbhavith Two updates, please:

1. Use the spec provided in the JIRA ticket (with a change to the domain as documented in the comment below), and;
2. Relocate the bit of code you've added to `app/models/searchgov_domain.rb` to the `check_status` method (see comment).

app/models/searchgov_domain.rb
krbhavith: I was trying to error out the defined domain in local machine, but the `test.gov` domain is resulting "200 OK" status all the time in local machine, but throwing an exception in staging. Need to figure that out.

spec/models/searchgov_domain_spec.rb
krbhavith: I was trying to error out the defined domain in local machine, but the test.gov domain is resulting "200 OK" status all the time in local machine, but throwing an exception in staging. Because of this, I was unable write a proper test, need to figure that out.

lorawoodford (approved):
LGTM

PR #990, opened 2022-08-18
SRCH-3134 Store raw documents in db

lorawoodford:
## Summary
Continuation of SRCH-3080 to implement raw storage of searchgov documents in the database.  Of note:
- Setting the environment variable `SEARCHGOV_DOCUMENT_NO_SAVE` to true will disable this functionality;
- The `save_document` method is written to look to the [ETag](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag) HTTP header to determine if the url being fetched has changed prior to saving the document to the SearchgovDocument table.
- For html documents, the raw html is stored in the `body` column, whereas for other document types the somewhat misleadingly named "metadata" value is stored in the db.  For non-html docs, "metadata" is the response from [get_recursive_metadata](https://github.com/GSA/search-gov/blob/main/lib/tika.rb#L4) which contains both the full text of the document and associated file metadata (which will both be needed when we transition to using the db for parsing of document metadata.

:warning: It is intended that this code should not be deployed until after matching cookbooks change to set the environment variable (`SEARCHGOV_DOCUMENT_NO_SAVE`) is deployed.
 
### Checklist
 
#### Functionality Checks

:warning: Automated checks pass. If Code Climate checks do not pass, explain reason for failures:
Code Climate is failing on cyclomatic complexity in the `fetch` method and several instances of too many memoized helpers in `spec/models/searchgov_url_spec.rb`.  I _could_ move the env var check in the fetch method down into `save_document` but I prefer to keep it where it is as it is more prominent and assumed to be a temporary bandaid during this initial roll out.  (And, that method was fairly dense even before these changes).  The memoized helpers would require some sizable refactoring.  I cleared up some by removing the `i14y_document` (which was only used in one location), but most were pre-existing before I got to this file.
 
#### Process Checks


spec/models/searchgov_url_spec.rb
```diff
@@ -150,6 +162,69 @@
...
+      context 'when no searchgov_document exists' do
+        it 'saves the document to searchgov_documents' do
+          expect { fetch }.to change { SearchgovDocument.count }.
+            from(0).to(1)
+        end
+
+        it 'saves html content in searchgov_documents' do
+          fetch
+          doc = SearchgovDocument.find_by(searchgov_url_id: searchgov_url.id)
+          expect(doc.body).to include('This is my content.')
```
MothOnMars, line 188: I would simplify this a bit, but also include the tags to make it clearer that the HTML is actually what's being stored (as opposed to the parsed content):
```ruby
doc = searchgov_url.searchgov_document
expect(doc.body).to include('<p>This is my content.</p>')
```
In fact, it would DRY things up to assign the `searchgov_document` for use anywhere in the fetching specs:
```ruby
let(:searchgov_document) { searchgov_url.searchgov_document }
```
That will allow you to replace all the instances of:
```ruby
SearchgovDocument.find_by(searchgov_url_id: searchgov_url.id)
```
with:
```ruby
searchgov_document
```

spec/models/searchgov_url_spec.rb
```diff
@@ -313,6 +388,14 @@
...
         end
       end
+
+      context 'when destroying' do
+        it 'deletes the searchgov_document' do
+          fetch
+          expect { searchgov_url.destroy }.to change { SearchgovDocument.count }.
+            from(1).to(0)
+        end
+      end
```
MothOnMars, line 398: This spec can be removed, as it will be covered by the association spec I mentioned above:
```ruby
it { is_expected.to have_one(:searchgov_document).dependent(:destroy) }
```

MothOnMars (changes requested):
Apologies again for the delay on this review! LMK if you have questions about any of the suggested changes.

app/models/searchgov_url.rb
```diff
@@ -178,13 +195,28 @@ def url_without_protocol
...
     else
       HtmlDocument.new(document: response.to_s, url: url)
     end
   end
 
+  def skip_save?(doc)
+    if application_document?
+      # If an application document, skip if that document's entity tag is unchanged and if
+      # the document was parsed with our current Tika version
+      (doc&.Etag == response.headers.to_hash['Etag']) && (doc&.tika_version == Tika.tika_version)
```
MothOnMars, line 209: Hmm - apparently the world is moving toward downcased headers: https://blog.yaakov.online/http-2-header-casing/. The `httprb` gem (used for our fetching) returns the camel-cased headers by default, but we can use lowercase, symbol accessors instead of converting them to a hash:
```ruby
response.headers[:etag]
```
Let's use the lowercased version of 'Etag' and any other header keys we need. This will also help with consistency in case we ever swap out `httprb` for a different HTTP client, such as Faraday:
```ruby
> Faraday.get('https://search.gov').headers
=> {"content-type"=>"text/html; charset=utf-8",
 "content-length"=>"7431",
 "connection"=>"close",
 "date"=>"Thu, 15 Sep 2022 13:30:59 GMT",
 "accept-ranges"=>"bytes",
 ...
 ```

File: app/models/searchgov_document.rb [hunk 4]
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class SearchgovDocument < ApplicationRecord
+  validates :body, presence: true
```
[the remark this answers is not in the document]
lorawoodford: Renamed to `web_document` to make connection to that model's parsing clearer.

lib/tika.rb
```diff
@@ -16,6 +16,12 @@ def self.get_recursive_metadata(file)
     JSON.parse(body)
   end
 
+  def self.tika_version
+    response = client.get('/version')
+
+    response.to_s
+  end
```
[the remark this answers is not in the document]
lorawoodford: While adding a test for this method I realized there should at least be some rudimentary error handling for this method (in case the Tika api is unavailable and does not return a version).  I mimicked the existing Tika error used in `get_recursive_metadata` and added a test for this as well.

db/migrate/20220816160655_create_searchgov_documents.rb
```diff
@@ -0,0 +1,13 @@
+class CreateSearchgovDocuments < ActiveRecord::Migration[6.1]
+  def change
+    create_table :searchgov_documents do |t|
+      t.text :body, size: :long
```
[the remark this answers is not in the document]
lorawoodford: Opted for `web_document`

spec/models/searchgov_document_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+    let(:doc) do
+      described_class.new(valid_attributes)
+    end
+
+    it 'assigns an entity tag if provided in the header' do
+      doc.header = { content_type: 'text/plain',
+                     Etag: '123' }
+
+      expect(doc.Etag).to eq('123')
+    end
```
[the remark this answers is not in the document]
lorawoodford: Updated to `context`

spec/models/searchgov_url_spec.rb
```diff
@@ -123,13 +127,21 @@
...
+          body: html,
+          headers: { content_type: 'text/html',
+                     Etag: '123' } }
       end
 
       before do
-        stub_request(:get, url).with(headers: { user_agent: DEFAULT_USER_AGENT }).
-          to_return({ status: 200, body: html, headers: { content_type: 'text/html' } })
-        stub_request(:get, url).with(headers: { 'User-Agent' => DEFAULT_USER_AGENT }).
+        stub_request(:get, url).
```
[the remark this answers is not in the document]
lorawoodford: Updated both these new `stub_request`s as well as other existing/unrelated `stub_request`s to keep things consistent and less redundant.

spec/models/searchgov_url_spec.rb
lorawoodford: 👍 Major improvement.

lorawoodford:
@MothOnMars This should be ready for re-review.

spec/models/searchgov_document_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+                 etag: '123' },
+      searchgov_url_id: 1
+    }
+  end
+  let(:doc) { described_class.new(valid_attributes) }
+
+  describe 'schema' do
+    it { is_expected.to have_db_column(:web_document).of_type(:text) }
+    it { is_expected.to have_db_column(:headers).of_type(:json) }
+    it { is_expected.to have_db_column(:tika_version).of_type(:decimal) }
```
MothOnMars, line 19: We should probably test all the column options:
```ruby
    it { is_expected.to have_db_column(:headers).of_type(:json).with_options(null: false) }

    it do
      is_expected.to have_db_column(:web_document).
        of_type(:text).with_options(null: false, limit: 4_294_967_295)
    end

    it do
      is_expected.to have_db_column(:tika_version).
        of_type(:decimal).with_options(precision: 10, scale: 4, default: nil)
    end
```    
 
 I used to think this was testing overkill until I learned the hard way that migrations silently ignore options that are set incorrectly...

app/models/searchgov_url.rb
```diff
@@ -179,13 +199,28 @@ def url_without_protocol
...
     else
       HtmlDocument.new(document: response.to_s, url: url)
     end
   end
 
+  def skip_save?(doc)
+    if application_document?
+      # If an application document, skip if that document's entity tag is unchanged and if
+      # the document was parsed with our current Tika version
+      (doc&.etag == response.headers[:etag]) && (doc&.tika_version == Tika.tika_version)
```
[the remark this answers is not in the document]
lorawoodford: Addressed on lns 150 and 154 above.

lorawoodford:
@MothOnMars Changes incorporated, eady for re-re-review.

app/models/searchgov_document.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class SearchgovDocument < ApplicationRecord
+  # SRCH-3134: web_document stores either:
+  # 1. the entire raw html of an html document, or
+  # 2. the full metadata and content of an application document of an application document.
```
MothOnMars, line 6: typo: "of an application document of an application document."

MothOnMars (changes requested):
One last tweak!

app/models/searchgov_document.rb
lorawoodford: Whoops, this is why I don't normally work in an office.  Fixed.

MothOnMars (approved):
LGTM!

lorawoodford:
@danswick This will require CodeClimate approval for the 2 documented issues in `app/models/searchgov_url.rb`.

The `Cyclomatic complexity for fetch is too high.` stems from the addition of a _temporary_ flag to toggle this new functionality on/off, and, as such, this failure will only be temporary and doesn't justify additional work at this time.

The `Assignment Branch Condition size for save_document is too high.` failure in the new `save_document` method can also be safely ignored at this time (if you agree). It isn't a great indicator of complexity in this context, and it is likely that this method will continue to be tweaked once this functionality is actually flagged on in prod (which it won't be just yet).

Thanks.

PR #1000, opened 2022-09-07
SRCH-3312: checking no-var in error state

yogesh27:
## Summary
- Updated the eslint file to follow the airbnb standards at https://airbnb.io/javascript/
 
### Checklist
 
#### Functionality Checks

#### Process Checks


yogesh27:
Hi @danswick changes are ready to be reviewed.

danswick:
Thanks, @yogesh27. Looking at the Code Climate report, couple of things come to mind right away. Let me know what you think:

1. Looks like there are at least some items in Code Climate that we could easily fix without having to get into any meaningful code changes. The “space after ‘//’ in comment” ones, for example. I would be totally okay implementing these smaller style changes as part of your eslint work, but we could also scope this as a follow-on task if you think that makes more sense.
2. We should probably ignore vendor scripts, especially minified ones. [app/assets/javascripts/searches/google_analytics.js](https://github.com/GSA/search-gov/blob/main/app/assets/javascripts/searches/google_analytics.js), for example. 
3. Anything that’s left over, we can evaluate and “accept” in bulk. You’ll need to tag me in when you’re ready, but I can click this Approve button to dismiss any eslint messages in CC we don’t intend to address. 

image

yogesh27:
> Thanks, @yogesh27. Looking at the Code Climate report, couple of things come to mind right away. Let me know what you think:
> 
> 1. Looks like there are at least some items in Code Climate that we could easily fix without having to get into any meaningful code changes. The “space after ‘//’ in comment” ones, for example. I would be totally okay implementing these smaller style changes as part of your eslint work, but we could also scope this as a follow-on task if you think that makes more sense.
> 2. We should probably ignore vendor scripts, especially minified ones. [app/assets/javascripts/searches/google_analytics.js](https://github.com/GSA/search-gov/blob/main/app/assets/javascripts/searches/google_analytics.js), for example.
> 3. Anything that’s left over, we can evaluate and “accept” in bulk. You’ll need to tag me in when you’re ready, but I can click this Approve button to dismiss any eslint messages in CC we don’t intend to address.
> 
> image

Hi @danswick I agree, I can take care of the simple ones in this task itself. 
@amanifarooque Let me know if this sounds goods and I will be make the updates.

amanifarooque:
@yogesh27 That approach sounds good to me!

yogesh27:
> @yogesh27 That approach sounds good to me!

Great. I will start making updates.

yogesh27:
> > @yogesh27 That approach sounds good to me!
> 
> Great. I will start making updates.

Hi @danswick I have updated the js code to include eslint-disable rule so that eslint can ignore the rules. One thing to notice is that our rails app doesnt support ES6 yet so I have disabled checking those rules as well. Eg. in the file app/assets/javascripts/uswds.js, I have added /* eslint-disable no-var, prefer-const, prefer-destructuring */

@amanifarooque For the react integration, we will need to have our rails app to work on ES6 and then the above disabled rules can be put back in.

danswick (approved):
This looks good to me, @yogesh27. I was going to say that we should bulk-ignore any vendor JS, but I'm not convinced that doing so would be any clearer or more maintainable than what you've done here given 1) how infrequently vendor files are updated (GA was last updated 9 years ago, for example) 2) the fact that vendor files aren't really colocated in a predictable way.

Thanks for updating our eslint system!

MothOnMars:
@yogesh27 , when you merge this, please update the commit message to something that describes the full set of changes, such as `SRCH-3312 update eslint rules`.

PR #1008, opened 2022-09-22
SRCH-3318 upgrade Tika to 2.4.1

MothOnMars:
## Summary
This PR:
- updates our document metadata parsing for compatibility with Tika 2.4.1
- adds & updates unit tests
 
### Checklist
 
#### Functionality Checks

#### Process Checks


app/models/application_document.rb
```diff
@@ -18,8 +24,9 @@ def extract_metadata
   end
 
   def parse_content
-    content = metadata['X-TIKA:content'] || ''
-    content.gsub!(/\n+/, "\n")
+    content = metadata['X-TIKA:content'] || +''
```
lorawoodford, line 27: It's a preference thing so I'm not going to block on this, but I find:

```
    content = metadata['X-TIKA:content']
    return '' unless content&.match?(/([a-zA-Z])|\d/)

    content.tr("\uFFFD", ' ')&.squish
```

to be slightly more readable since it's rare to see that mutable string construction in the wild and since you've added the return there that can catch the nil content if one exists.  But again, totally a preference call.

lorawoodford (approved):
One comment left in-line, but also approving as-is if you disagree.  LGTM.

app/models/application_document.rb
MothOnMars: good call - will do

lorawoodford (dismissed):
LGTM

MothOnMars:
@lorawoodford , this needs a re-review. I had to tweak the specs after merging `main` to make them compatible with Tika 2.4.

lorawoodford (approved):
Sorry for the delay.  LGTM.

PR #1020, opened 2022-09-30
SRCH-3415 upgrade to Rails 7.0

MothOnMars:
## Summary
This PR:
- bumps Rails to 7.0.x
- updates/adds/deletes files based on running `rails app:update`
- comments out tests that currently fail, with references to the necessary Jira tickets

This is the initial commit into the `feature/rails_7` branch, which we'll use for the fix-it tickets, and eventually full testing on staging.
 
### Checklist
 
#### Functionality Checks

⚠️  Automated checks pass. If Code Climate checks do not pass, explain reason for failures:
As noted above, several tests as well as our coverage check have been commented out, to ensure that subsequent changes can be made against a passing branch. All tests and the coverage will be re-enabled before `feature/rails_7` is merged to `main`.
 
#### Process Checks


bin/rails
```diff
@@ -1,4 +1,4 @@
 #!/usr/bin/env ruby
-APP_PATH = File.expand_path('../config/application', __dir__)
+APP_PATH = File.expand_path("../config/application", __dir__)
```
MothOnMars, line 2: The single- to double-quote changes in these boilerplate files are the result of running `rails app:update`. A follow-up ticket will enforce double-quotes for the stock files in `searchgov-style`, so that Rubocop and Rails are not enforcing conflicting styles.

config/initializers/active_job_uniqueness.rb
```diff
@@ -1,35 +1,37 @@
 # frozen_string_literal: true
 
-ActiveJob::Uniqueness.configure do |config|
-  # Global default expiration for lock keys. Each job can define its own ttl via :lock_ttl option.
-  # Stategy :until_and_while_executing also accept :on_runtime_ttl option.
-  config.lock_ttl = 1.day
+Rails.application.reloader.to_prepare do
```
MothOnMars, line 3: This `to_prepare` line is the only substantial change in this file, per https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#autoloading-during-initialization. The rest of the lines are just indentation shifts.

config/initializers/new_framework_defaults_7_0.rb
```diff
@@ -0,0 +1,136 @@
+# Be sure to restart your server when you modify this file.
```
MothOnMars, line 1: These will be addressed in a follow-up ticket: https://cm-jira.usa.gov/browse/SRCH-3413

spec/support/shared_sites_controller_shared_behavior.rb
```diff
@@ -32,7 +32,7 @@
 
   before do
     UserSession.create current_user
-    expect(User).to receive(:find_by_id).and_return(current_user)
```
MothOnMars, line 35: The expectations in `before` blocks were causing some flakey spec behavior in Rails 7. `before` blocks shouldn't include expectations, so I've changed them all to `allow` and cleaned up some other Rubocop offenses.
https://docs.rubocop.org/rubocop-rspec/cops_rspec.html#rspecexpectinhook

Gemfile
```diff
@@ -38,11 +38,15 @@ gem 'googlecharts', '~> 1.6.12'
 gem 'tweetstream', git: 'https://github.com/GSA/tweetstream'
 gem 'twitter', git: 'https://github.com/GSA/twitter.git', branch: '5-stable'
 gem 'flickraw', '~> 0.9.9'
-gem 'active_scaffold', '~> 3.6.2'
+gem 'active_scaffold', '~> 3.6.0'
```
MothOnMars, line 41: Yes, I had to downgrade `active_scaffold`. This is the result of them changing the rails requirements from `rails (>= 4.2.0)` to `rails (>= 4.2.0, < 6.2)` in 3.6.1:
https://github.com/activescaffold/active_scaffold/commit/c89a670462f3da490177ec28667af3c7455de344#diff-89cade48462044ee1b672dc5f4c3ec250fbd29effcd8932096a23c1283c6731fR13

As tests are passing, and my local smoke-tests of super admin pass, I'm fairly comfortable with the downgrade. It's possible that a full test on staging will surface problems, but we can tackle those at that point.

bin/rails
lorawoodford: SRCH-3424 filed, but marked as a no-go as we already skip these files in rubocop.  Adding the comment here as well to close the loop.

MothOnMars: 👍

Gemfile
```diff
@@ -159,6 +167,11 @@ group :development, :test do
   gem 'pry-rails', '~> 0.3.6'
   gem 'awesome_print'
   gem 'puma', '~> 5.3'
+  # See https://guides.rubyonrails.org/debugging_rails_applications.html#debugging-with-the-debug-gem
+  # We may eventually remove the pry-* gems in favor of 'debug'.
+  # Installation results in an intermittent spec failure, so commenting this out for now:
+  # https://cm-jira.usa.gov/browse/SRCH-3344
+  # gem 'debug', platforms: %i[mri mingw x64_mingw]
```
MothOnMars, line 174: About to merge `main`, which will install `debug` now that SRCH-3344 is resolved.

.gitignore
```diff
@@ -1,29 +1,44 @@
-*.swp
-.DS_Store
-.bundle
+# See https://help.github.com/articles/ignoring-files for more about ignoring files.
+#
+# If you find yourself ignoring temporary files generated by your text editor
+# or operating system, you probably want to add a global ignore instead:
+#   git config --global core.excludesfile '~/.gitignore_global'
```
lorawoodford, line 5: This is a fine/positive change, but when the time comes we should ensure this gets added in to onboarding/early dev setup instructions.  I fear a future where we start seeing accidental commits of DS_Store files because this change occurred somewhat silently and obscured by the larger Rails upgrade.

MothOnMars: I'm not too worried about that, as this is the only one of our repos that has ignored the `.DS_Store` files. Various devs have occasionally committed various types of swap files (not just DS_store), and we've just addressed it at that point.

lorawoodford (approved):
LGTM.

Confirmed follow-on tickets are present and clear, confirmed I can start the application locally, and specs passing (unless set to pending).

PR #1021, opened 2022-09-30
SRCH-3344 Add `debug` gem

ddzz:
## Summary
This PR adds the `debug` gem to the `Gemfile`. One test had to be refactored due to an additional call to `File.exist?` being made by `debug` during test execution.
 
### Checklist
 
#### Functionality Checks

#### Process Checks


spec/models/med_topic_spec.rb
```diff
@@ -260,6 +260,8 @@
       let(:response) { double('http response') }
 
       before do
+        allow(File).to receive(:exist?)
+        allow(File).to receive(:exist?).with(/#{xml_file_path}$/)
```
MothOnMars, line 264: We only need the initial `allow(File).to receive(:exist?)` line plus the line specifying the return value (using `allow` instead of `expect`. We can also specify `allow(File).to receive(:exist?)` once in the surrounding `describe` block to avoid duplication:
```ruby
  describe '.download_medline_xml' do
    before { allow(File).to receive(:exist?) }

    context 'when file is not present in the tmp/medline directory' do
      ...

      before do
        allow(File).to receive(:exist?).with(/#{xml_file_path}$/).and_return(false)
        ...
      end

...

    context 'when file is present in the tmp/medline directory' do
      let(:xml_file_path) { 'tmp/medline/mplus_topics_2012-07-21.xml' }

      before do
        allow(File).to receive(:exist?).with(/#{xml_file_path}$/).and_return(true)
        ...
      end
```
These are old specs with a lot of `expect`s in the context blocks, but it's preferred to use `allow`s in the contexts and `expect`s only in the expectation blocks: https://docs.rubocop.org/rubocop-rspec/cops_rspec.html#rspecexpectinhook . (Fixing the other `expect`s in the `before` blocks is out of scope of this story.)

MothOnMars (changes requested):

spec/models/med_topic_spec.rb
ddzz: Thanks for reviewing. I agree that only the `allow` is needed since a stub is sufficient for the actual test. The PR has been updated.

MothOnMars (approved):
LGTM!

PR #1022, opened 2022-10-03
SRCH 3400 js indexing production

krbhavith:
## Summary
- JS Fetcher library using Selenium Webdriver to fetch JavaScript page content.
- new field `js_renderer` boolean on `SearchgovDomain` to opt-in/out of JS rendering functionality.
 
### Checklist
 
#### Functionality Checks

:warning: had to use new method due to another code climate issue "limitation of lines same method"
 
#### Process Checks


spec/models/searchgov_url_spec.rb
```diff
@@ -380,6 +380,35 @@
       end
     end
 
+    context 'when the searchgov domain has js renderer enabled' do
+      let(:domain_with_js_renderer) do
+        instance_double(SearchgovDomain, check_status: '200 OK', js_renderer: true)
+      end
+
+      it 'fetches page with javascript response' do
```
danswick, line 388: I think this test would pass for any HTML document, right?

krbhavith: Yes, it would pass for any response that would start with `'<head>'`. Cannot determine the exact content type, here making sure JS content is rendered.

lib/js_fetcher.rb
```diff
@@ -0,0 +1,23 @@
...
+
+module JsFetcher
+  def self.fetch(url)
+    options = Selenium::WebDriver::Chrome::Options.new(args: ['headless'])
+    driver = Selenium::WebDriver.for(:chrome, options: options)
+    wait = Selenium::WebDriver::Wait.new(timeout: 5)
+
+    begin
+      driver.get(url)
+      wait.until { document_initialised(driver) }
```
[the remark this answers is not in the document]
krbhavith: The explicit wait did not work while I was testing few sites and then I had to use `sleep` to wait for the page to load.

spec/models/searchgov_url_spec.rb
```diff
@@ -381,6 +381,37 @@
       end
     end
 
+    context 'when the searchgov domain has js renderer enabled' do
+      let(:js_searchgov_url) { described_class.create!(url: 'https://www.medicare.gov/sign-up-change-plans') }
```
lorawoodford, line 385: I'm still mulling through the larger testing philosophy for these changes, but didn't want to hold off my review into the weekend because I'm still thinking.

As this stands, these tests will actually send requests to these urls each time the tests are run, which is something we want to avoid.  See https://github.com/GSA/search-gov/wiki/Editing-Recording-and-Re-recording-API-calls-with-VCR-(WIP)

This may also be overkill for what we're actually trying to test here -- that the JsFetcher module receives a fetch request when SearchgovUrl#fetch is issued on a SearchgovDomain that has js_renderer set to true.  Checking the content of the response here doesn't quite feel right.

The inner logic of what is returned by JsFetcher probably belongs in a JsFetcher spec in `spec/lib`.  Happy to chat more on this while I continue to mull it over next week.

lorawoodford (changes requested):
@krbhavith @danswick Some initial review comments left in-line.  I'm still thinking through what makes sense for the tests, but didn't want to hold back the easier changes on that count.

spec/models/searchgov_url_spec.rb
lorawoodford: So, after quick tinkering, something more like this might do the trick (this should still be cleaned up to reduce redundancy):

```
context 'when the searchgov domain has js renderer enabled' do
      before do
        allow(searchgov_url).to receive(:searchgov_domain).and_return(domain_with_js_renderer)
        allow(JsFetcher).to receive(:fetch)
      end

      let(:url) { 'https://www.medicare.gov/sign-up-change-plans' }
      let(:domain_with_js_renderer) do
        instance_double(SearchgovDomain, check_status: '200 OK', available?: true, js_renderer: true)
      end

      it 'fetches page with javascript fetcher' do
        fetch
        expect(JsFetcher).to have_received(:fetch).with(url)
      end
    end

    context 'when the searchgov domain has js renderer disabled' do
      before do
        allow(searchgov_url).to receive(:searchgov_domain).and_return(domain_without_js_renderer)
        allow(JsFetcher).to receive(:fetch)
      end

      let(:url) { 'https://www.medicare.gov/sign-up-change-plans' }
      let(:domain_without_js_renderer) do
        instance_double(SearchgovDomain, check_status: '200 OK', available?: true, js_renderer: false)
      end

      it 'fetches page without javascript fetcher' do
        fetch
        expect(JsFetcher).not_to have_received(:fetch).with(url)
      end
    end
```

lib/js_fetcher.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
```
danswick, line 1: As @lorawoodford mentioned in a comment in spec/models/searchgov_url_spec.rb, `js_fetcher` probably needs a unit test.

app/models/searchgov_url.rb
```diff
@@ -206,6 +206,15 @@ def parse_document
     Rails.logger.info "[SearchgovUrl] Parsing document for #{url}"
     if application_document?
       ApplicationDocument.new(document: download.open, url: url)
+    else
+      update_document
+    end
+  end
+
+  def update_document
```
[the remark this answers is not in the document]
krbhavith: Updated as requested.

spec/lib/js_fetcher_spec.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+describe 'js fetcher' do
+	context 'when the searchgov domain has js renderer enabled' do
+	  let(:url) { 'https://www.medicare.gov/sign-up-change-plans' }
+	 
+	  it 'fetches page with javascript response' do
+  		js_response = JsFetcher.fetch(url)
```
lorawoodford, line 8:
```suggestion
  		js_response = described_class.fetch(url)
```

lorawoodford (changes requested):
@krbhavith I've added a few little changes throughout, but the bigger issue remains that it appears the vcr cassette does not get consulted during spec runs once `127.0.0.1` is added back to the shared setup.  

If you make a change to the cassette that would cause the tests to fail (for example, remove the `<head>` from the string starting on line 132) and still have `127.0.0.1` removed from `ignore_hosts` in `shared_vcr_setup`, the test will (rightly) fail with the following:

```
Failures:

  1) JsFetcher fetches page with javascript response
     Failure/Error: expect(js_response).to start_with('<head>')
       expected ">\n    <meta charset=\"utf-8\"><script type=\"text/javascript\" src=\"//zn56jl522ec1a1wzf-medicare.g...></div><script><iframe data-gtm-yt-inspected-3=\"true\" style=\"display: none;\"></iframe></script>" to start with "<head>"
```

But, once you return `127.0.0.1` to the `ignore_host` config, the same is not true.  The spec run does not refer to the vcr cassette, but instead pings the live medicare site, and will pass (even though you've changed the vcr cassette to a state which should cause the test to fail).

This becomes more obvious if you uncomment the final debug line in `shared_vcr_setup` and watch the logging.  When `127.0.0.1` is in ignore hosts, the new js cassette is not consulted (`[Cassette: 'js_fetcher']` is only logged once on initialization).  But, when localhost is removed from the `ignore_host`, `[Cassette: 'js_fetcher']` is pinged multiple times.

This remains the big issue at the moment.  Of lesser importance now, but worth mentioning, is that we should revisit @danswick 's initial question about the test expectations.  Returning `not_nil` and `<head>` seem like dubious tests to me, but I don't know the landscape as well as you do to know what can be trusted to be a sign of a success case, and what is a sign of a failed case.  But first we've got to figure out the bigger issue of ensuring the cassettes are working.

Let's try and talk as a team after standup?

spec/lib/js_fetcher_spec.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+describe JsFetcher do
+  # Below live site will pinged on each test run, should be handled with VCR cassette config.
```
lorawoodford, line 4: @krbhavith As mentioned in standup, please update this comment to add the JIRA ticket number for the follow-on ticket you have created/will create for resolving the vcr configuration issues at some future date.  This saves this comment from getting lost in the code.

spec/lib/js_fetcher_spec.rb
```diff
@@ -0,0 +1,12 @@
...
+  let(:url) { 'https://search.gov/javascript-test.html' }
+
+  it 'fetches page with javascript response' do
+    js_response = described_class.fetch(url)
+    expect(js_response).not_to be_nil
```
lorawoodford, line 9: Now that we've got a stronger expectation set on line 10, the not nil expectation on ln 9 seems unnecessary.  Go ahead an remove it, so that there's only one expectation in this block.

spec/models/searchgov_url_spec.rb
```diff
@@ -381,6 +381,42 @@
       end
     end
 
+    context 'when the searchgov domain has js renderer enabled' do
+      before do
+        allow(searchgov_url).to receive(:searchgov_domain).and_return(domain_with_js_renderer)
+        allow(JsFetcher).to receive(:fetch)
+      end
+
+      # Below live site will pinged on each test run, should be handled with VCR cassette config.
```
lorawoodford, line 390: This comment can be removed, as these specs aren't actually pinging a live url.  (If you turned off your wifi and ran this locally, it'd still pass because of the vcr_cassettes added below.)

spec/models/searchgov_url_spec.rb
```diff
@@ -381,6 +381,42 @@
...
+      end
+    end
+
+    context 'when the searchgov domain has js renderer disabled' do
+      before do
+        allow(searchgov_url).to receive(:searchgov_domain).and_return(domain_without_js_renderer)
+        allow(JsFetcher).to receive(:fetch)
+      end
+
+      # Below live site will pinged on each test run, should be handled with VCR cassette config.
```
lorawoodford, line 408: Same as above, remove comment.

spec/vcr_cassettes/js_fetcher.yml
```diff
@@ -0,0 +1,2763 @@
+---
```
lorawoodford, line 1: This entire cassette can be removed as it isn't being used by the `js_fetcher` spec.  You may want to save it to a local branch in your fork if you want to reference it in the vcr config ticket you're creating/will create.  But, leaving it here is unnecessary and potentially confusing in the future.

lorawoodford (changes requested):
@krbhavith A few final (hopefully very small) changes commented in-line, but this looks very, very close.

spec/lib/js_fetcher_spec.rb
krbhavith: Updated comments with the Jira ticket number

lorawoodford (approved):
@krbhavith LGTM.

@danswick This will need a CC approval.  Code Climate's structure engine is complaining about the number of methods in SearchgovUrl, which is (IMHO) not worth worrying about at this point.  SearchgovUrl will continue to see a lot of activity/changes in the coming sprints as work continues on other feature work, so other opportunities will present themselves to refactor this (if we deem it prudent/worthwhile to do so).

PR #1025, opened 2022-10-05
SRCH-3412 upgrade to Rails 7.0

MothOnMars:
## Summary
This PR upgrades Rails to 7.0.4. It contains changes in previously reviewed PRs to this feature branch:
```
SRCH-3404 Fix failing view tests in Rails 7 (#1033)
SRCH-3409 Fix failing affiliate cucumber test by disabling eager loading (#1032)
SRCH-3413 Enable Rails 7 default configuration (#1039)
SRCH-3414 Enable test coverage reporting in Rails 7 branch (#1034)
SRCH-3415 upgrade to Rails 7.0 (#1020)
```
 
### Checklist
 
#### Functionality Checks

#### Process Checks


MothOnMars:
@lorawoodford , apologies, I forgot to bonk the "ready for review" button, AND I had intended to call out the schema.rb issue you mentioned. 

However, your comment did prompt me to do an extra sanity check on the schema change. The changes wouldn't affect anything in production, but I wanted to double-check that we wouldn't end up with a mismatch between our dev/test db and production. 

The only table in production with a specified `precision` for the timestamp columns is the `searchgov_documents` table, which is the only table that was created w/ Rails 6:
```
SELECT table_name, column_name, datetime_precision
  FROM information_schema.columns 
  WHERE (
    table_schema like 'usasearch_%'
    and data_type='datetime'
    and column_name in('created_at','updated_at')
    and datetime_precision != 0
  );
```

table_name | column_name | datetime_precision
|---|---|---|
searchgov_documents | created_at | 6
searchgov_documents | updated_at | 6

All the other timestamp columns have `datetime_precision` set to `0`. The updated schema.rb preserves the existing settings, as `precision: nil` results in `datetime_precision` set to `0`, while the `searchgov_documents` table lacks a specified `precision`, which results in the default MySQL `datetime_precision` of `6`. There *will* be a diff between the `datetime_precision` for the production db vs. dev/test for the `ar_internal_metadata` table, which currently has a precision of `0`, and will have a precision of `6` in our local dbs. As that is an internal Rails table, I don't think we need to worry about a diff between local dbs vs. production.

lorawoodford (approved):
@MothOnMars This all makes sense to me.  Thanks for the additional checking/confirmation.

LGTM.

lorawoodford (approved):
LGTM

PR #1027, opened 2022-10-10
SRCH-3425 Add `exception_notification` gem

ddzz:
## Summary
This PR adds and configures the `exception_notification` and `dogapi` gems, which are used to send exception data from production to Datadog.
 
### Checklist
 
#### Functionality Checks

- [ ] You have run `bundle update` and committed your changes to Gemfile.lock.
 
#### Process Checks


config/environments/production.rb
```diff
@@ -101,6 +101,15 @@
   # Do not dump schema after migrations.
   config.active_record.dump_schema_after_migration = false
 
+  # Datadog API client used to send exception data to Datadog.
+  if !!Rails.application.secrets.datadog[:api_enabled]
+    client = Dogapi::Client.new(Rails.application.secrets.datadog[:api_key], Rails.application.secrets.datadog[:application_key])
```
MothOnMars, line 106: Let's clean this up a bit for readability:
```ruby
datadog_config = Rails.application.secrets.datadog
if datadog_config[:api_enabled] # no need for the !!, as datadog_config[:api_enabled] will evaluate to a boolean
  client = Dogapi::Client.new(datadog_config[:api_key], datadog_config[:application_key])
```
It looks like the optimal way to configure this gem is now via an initializer: 
https://github.com/smartinez87/exception_notification#extras

Unfortunately, they buried that info at the bottom of their README, so I didn't even see it until today. However, we'll need to enable notifications for Resque, so we can make that change when we add an initializer for that story. Will add notes to a follow-up ticket.

MothOnMars (changes requested):

config/environments/production.rb
ddzz: Thanks for reviewing. I had the `!!` to handle `:api_enabled` being `nil`, but since `nil` is evaluated as false it's not needed.

MothOnMars (approved):
LGTM

PR #1029, opened 2022-10-11
SRCH-3423 Fix flaky add/edit/remove Collection test

ddzz:
## Summary
This PR addresses a race condition in this test that has been causing flakiness in CI. I was able to recreate the failure locally, and the fix did not fail when running the test in a loop 40 times.
 
### Checklist
 
#### Functionality Checks

#### Process Checks


MothOnMars (commented):
I left a note in the ticket. We could use a little more detail (in Jira) on the root cause and the fix.

MothOnMars (approved):
LGTM

PR #1032, opened 2022-10-14
SRCH-3409 Fix failing affiliate cucumber test by disabling eager loading

ddzz:
## Summary
This PR disables eager loading in the test environment in the Rails 7 branch, which was causing `features/admin.feature:47` to fail due to`Admin::AffiliatesController` being loaded before the test began. `Language`s were not present in the test database when the controller was loaded, which meant that the locale (a required field) could not be selected for the affiliate. Since the required field was missing, the affiliate could not be updated, which is what the test was trying to do.
 
### Checklist
 
#### Functionality Checks

#### Process Checks


config/environments/test.rb
```diff
@@ -11,10 +11,10 @@
...
   config.cache_classes = true
 
-  # Eager loading loads your whole application. When running a single test locally,
-  # this probably isn't necessary. It's a good idea to do in a continuous integration
-  # system, or in some way before deploying your code.
-  config.eager_load = ENV["CI"].present?
+  # Do not eager load code on boot. This avoids loading your whole application
+  # just for the purpose of running a single test. If you are using a tool that
+  # preloads Rails for running tests, you may have to set it to true.
+  config.eager_load = false
```
MothOnMars, line 17: @ddzz  , the root cause makes sense, but we shouldn't disable eager loading in CI, as that helps catch any loading issues that we might not run into locally. I have been trying a few strategies to work around this, but I think our best bet is to explicitly load the necessary fixtures in CircleCI before we run the cucumber tests, with a comment explaining why. This is what I'm thinking:
```yaml
#.circleci/config.yml

  cucumber:
...
    steps:
....
      - prepare_database

      # In CI, the application is eager loaded. This ensures the db contains the necessary
      # records before the application is loaded.
      - run:
          name: Load Database Fixtures
          command: RAILS_ENV=test rails db:fixtures:load FIXTURES_PATH='./spec/fixtures' FIXTURES=languages,users....#(see the list in features/support/hooks.rb)
```
I'm not in love with that solution either, so I'm open to ideas.

MothOnMars (changes requested):

config/environments/test.rb
ddzz: We could load the fixtures through `db/seeds.rb`. With that approach we don't need an extra step in CircleCI.

MothOnMars (approved):
LGTM

PR #1040, opened 2022-10-27
SRCH-3422 Upgrade newrelic_rpm gem

ddzz:
## Summary
This PR upgrades the `newrelic_rpm` to the latest version, which required multiple configuration changes.

Associated cookbooks PR to remove a deprecated setting: https://github.com/GSA/usasearch-cookbooks/pull/665
 
### Checklist
 
#### Functionality Checks
 
#### Process Checks


config/newrelic.yml
```diff
@@ -1,11 +1,15 @@
+common: &default_settings
+  host: <%= Rails.application.secrets.newrelic[:endpoint] %>
+
 <%= Rails.env %>:
+  <<: *default_settings
   app_name: <%= Rails.application.secrets.newrelic[:app_name] %>
   license_key: <%= Rails.application.secrets.newrelic[:license_key] %>
   log_level: <%= Rails.application.secrets.newrelic[:log_level] || 'info' %>
   monitor_mode: <%= Rails.application.secrets.newrelic[:enabled] ? 'true' : 'false' %>
-  apdex_t: <%= Rails.application.secrets.newrelic[:apdex_t] || 0.5 %>
```
ddzz, line 6: The `apdex_t` config has been deprecated and is now set in the New Relic UI, which is why this was removed.

config/newrelic.yml
```diff
@@ -16,7 +20,9 @@
     explain_threshold: 0.5
   error_collector:
     enabled: true
-    capture_source: true
```
ddzz, line 19: `capture_source` is set to `true` by default.

config/newrelic.yml
```diff
@@ -16,7 +20,9 @@
...
-    ignore_errors: ActionController::RoutingError
-  common: &default_settings
-    host: <%= Rails.application.secrets.newrelic[:endpoint] %>
+    ignore_classes: ['ActionController::RoutingError']
+  application_logging:
+    forwarding:
+      enabled: false
+  distributed_tracing:
+    enabled: false
```
MothOnMars, line 28: Is `distributed_tracing` being disabled due to associated costs? If so, can you add a note to https://cm-jira.usa.gov/browse/SRCH-2796? We were hoping to enable that at some point.

config/newrelic.yml
MothOnMars: Looks like that setting has been removed entirely: https://docs.newrelic.com/docs/apm/agents/ruby-agent/configuration/ruby-agent-configuration/

config/newrelic.yml
MothOnMars: If you haven't done so already, please create a subtask for the cookbooks change and link it here. That should be tested in tandem with these changes.

MothOnMars (changes requested):
I have a few follow-up questions & comments. There are a few other configurations in here that match the defaults and could be removed, but I don't think it hurts to be explicit (especially if the defaults change in the future; I have to assume there's an historical reason we specified those settings).

config/newrelic.yml
ddzz: This setting is still in the full New Relic config; it's just not documented. There are a lot more settings in the full config than what's on that page.

config/newrelic.yml
```diff
@@ -1,11 +1,15 @@
+common: &default_settings
+  host: <%= Rails.application.secrets.newrelic[:endpoint] %>
```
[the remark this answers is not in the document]
ddzz: This was New Relic's suggestion on how to structure the config when using the `gov-collector.newrelic.com` host, which is not the default. When I was testing this locally I had a modified configuration to include a separate configuration for development, and I considered adding more to `common`. But it adds unnecessary complexity right now, so I'll remove it.

config/newrelic.yml
ddzz: Actually, I may be wrong here. I'll do some additional testing to confirm whether this still exists.

config/newrelic.yml
ddzz: `distributed_tracing` is disabled right now, but it's enabled by default starting with 8.x: https://docs.newrelic.com/docs/apm/agents/ruby-agent/getting-started/migration-8x-guide/#distributed_tracing

I'm not sure what the associated cost is.

config/newrelic.yml
ddzz: Confirmed that this setting has been removed entirely from the gem. It was persisting in the New Relic UI when I tested it yesterday, but according to the git log for the gem, this setting was removed in 2014.

config/newrelic.yml
MothOnMars: Related PR: https://github.com/GSA/usasearch-cookbooks/pull/665

MothOnMars (changes requested):
@ddzz , I have been thinking about this some more, and I'm actually going to have you go back to the drawing board on this. I'd like you to generate a fresh `config/newrelic.yml` file using their generator:
```
newrelic install "My application" --force
```
That will generate a boilerplate config file based on their template:
https://github.com/newrelic/newrelic-ruby-agent/blob/dev/newrelic.yml

After you've generated the file, you'll update the necessary values and settings per our existing config. For any settings that use their default, we can leave those commented out. You can also comment out the `staging` section at the bottom, because we use the same config for staging as we do in production.

This will require a bit more work right now, but it should make future upgrades easier, as we'll be able to see what version of the gem generated the config file, and quickly diff changes between versions by re-running the generator. It will also include their documentation on each setting, which should facilitate investigations and changes in the future.

ddzz:
To make this PR easier to review, here's the new config with all comments removed:

```yml
common: &default_settings
  license_key: <%= Rails.application.secrets.newrelic[:license_key] %>
  app_name: <%= Rails.application.secrets.newrelic[:app_name] %>
  host: <%= Rails.application.secrets.newrelic[:host] %>
  log_level: <%= Rails.application.secrets.newrelic[:log_level] || 'info' %>
  application_logging.forwarding.enabled: false
  backport_fast_active_record_connection_lookup: true
  distributed_tracing.enabled: false
  error_collector.ignore_classes: ['ActionController::RoutingError']
  instrumentation.curb: disabled
  monitor_mode: <%= Rails.application.secrets.newrelic[:enabled] ? 'true' : 'false' %>
  transaction_tracer.transaction_threshold: 'apdex_f'
development:
  <<: *default_settings
  app_name: search.gov (Development)

production:
  <<: *default_settings
  ```

File: config/newrelic.yml [hunk 5]
```diff
@@ -1,22 +1,693 @@
...
+# RAILS_ENV or RACK_ENV (as appropriate) is used to determine the environment.
+# If your application has other named environments, configure them here.
+development:
+  <<: *default_settings
+  app_name: search.gov (Development)
+
+# test:
+#   <<: *default_settings
+#   # It doesn't make sense to report to New Relic from automated test runs.
+#   monitor_mode: false
```
MothOnMars, line 685: Please uncomment the `test` section. `monitor_mode: false` prevents data from being sent to NR during test runs, but it will help us catch any faulty configs in the default settings above. The initializer will run in all environments.

config/newrelic.yml
```diff
@@ -1,22 +1,693 @@
...
+
+  # If true, the agent automatically detects that it is running in a Pivotal Cloud Foundry environment.
+  # utilization.detect_pcf: true
+
+# Environment-specific settings are in this section.
+# RAILS_ENV or RACK_ENV (as appropriate) is used to determine the environment.
+# If your application has other named environments, configure them here.
+development:
+  <<: *default_settings
+  app_name: search.gov (Development)
```
MothOnMars, line 680: We should extract this from the secrets:
```ruby
app_name: <%= Rails.application.secrets.newrelic[:app_name] %> (Development)
```

MothOnMars (changes requested):

config/newrelic.yml
ddzz: Good point, thanks for reviewing.

config/newrelic.yml
ddzz: I just checked the cookbooks, and the `app_name` attribute actually already interpolates the environment, so this line can be removed.

ddzz: I added `(Development)` to `secrets.yml.dev` so that we can take out this line entirely.

config/secrets.yml.dev
```diff
@@ -91,10 +91,10 @@ dev_settings: &DEV_SETTINGS
     adapter: :net_http_persistent
   newrelic:
     <<: *NEWRELIC_SECRETS
-    app_name: USA Search (PUT YOUR NAME HERE)
+    app_name: USA Search (PUT YOUR NAME HERE) (Development)
```
MothOnMars, line 94: As long as we're updating this, let's keep it consistent with the cookbooks*:
```
app_name: search.gov (PUT YOUR NAME HERE) (<%= Rails.env %>)
```
We still need to interpolate the environment, as these secrets are used for both test and dev environments.

("USA Search" was our old name, and the old repo name was "usasearch". Those terms are still around, but we're updating them to "search.gov" when possible.)

MothOnMars (changes requested):

config/secrets.yml.dev
ddzz: Ah, I didn't realize these were used in the test environment as well. PR has been updated.

PR #1041, opened 2022-10-28
SRCH-2554 Add bootsnap to dev and test environments

ddzz:
## Summary
This PR adds the bootsnap gem to the dev and test environments, which speed up boot times. Due to concerns around cache maintenance, the gem is not being enabled in production.
 
### Checklist
 
#### Functionality Checks
 
#### Process Checks


Gemfile
```diff
@@ -166,6 +166,7 @@ group :development, :test do
   gem 'awesome_print'
   gem 'puma', '~> 5.3'
   gem 'debug'
+  gem 'bootsnap', '~> 1.13.0', require: 'bootsnap/setup'
```
ddzz, line 169: [The documentation for bootsnap](https://github.com/Shopify/bootsnap#usage) suggests loading bootsnap in the Gemfile as an alternative to loading it in `config/boot.rb`. `Rails.env.development?` and `Rails.env.test?` cannot be accessed in `config/boot.rb`, which is why I've put it in the Gemfile.

MothOnMars: Sounds good - we can keep it simple for now. 

As this is only being used in test/dev environments, we can relax the pessimistic versioning. Let's set it to `'~> 1.13'`.

MothOnMars (changes requested):

PR #1045, opened 2022-11-08
Revert "SRCH 3400 js indexing production"

MyNameIsMissing:
Reverts GSA/search-gov#1022

Reverting because the CircleCI tests have failed twice in a row in `main`.  Speculation is one of the new js specs occasionally hangs and needs to be adjusted.


krbhavith (approved):
Looks good

PR #1060, opened 2022-12-02
SRCH-3604 fix email initializer

MothOnMars:
## Summary
- This PR prevents the email initializer from failing on environments that have specified a `force_to` email address (such as staging).
 
### Checklist
 
#### Functionality Checks
 
#### Process Checks

### DIFF
--- a/config/initializers/email.rb
+++ b/config/initializers/email.rb
@@ -1,18 +1,22 @@
-SearchGovInterceptor = Struct.new(:force_to) do
-  def delivering_email(message)
-    message.to = [force_to]
+# frozen_string_literal: true
+
+Rails.application.config.to_prepare do
+  SearchGovInterceptor = Struct.new(:force_to) do
+    def delivering_email(message)
+      message.to = [force_to]
+    end
   end
-end
 
-email_config = Rails.application.secrets.email || { }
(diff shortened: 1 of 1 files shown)


config/initializers/email.rb
```diff
@@ -1,18 +1,22 @@
-SearchGovInterceptor = Struct.new(:force_to) do
-  def delivering_email(message)
-    message.to = [force_to]
+# frozen_string_literal: true
+
+Rails.application.config.to_prepare do
```
MothOnMars, line 3: The diff looks much bigger than it is due to the change in indentation. The only relevant change is that the code of this initializer is now wrapped in a `to_prepare` block:
```ruby
Rails.application.config.to_prepare do
  # pre-existing code
end
```
This change is required by Rails 7, and was not surfaced while testing because the staging stack used for testing did not specify a `force_to` address. See https://guides.rubyonrails.org/autoloading_and_reloading_constants.html#use-case-1-during-boot-load-reloadable-code for info on the fix.

MyNameIsMissing: Thanks for the ref.

MyNameIsMissing (approved):
LGTM

PR #1064, opened 2022-12-09
SRCH-3541 Add and configure React-Rails and Shakapacker gems

ddzz:
## Summary
This PR adds the `react-rails` and `shakapacker` gems. A basic React component also been created for the SERP redesign. The gems have been configured to use Webpack for bundling JavaScript files for the redesign. TypeScript support has also been added in this PR.
 
### Checklist
 
#### Functionality Checks
 
#### Process Checks


deploy/before_symlink.rb
```diff
@@ -7,13 +7,17 @@
...
 end
 
-# Pre-compile assets. Also, a very small subset of the assets
-# need to be available without digest fingerprints in their
-# filenames - assets that live "in the wild" and can't be
-# updated whenever our asset fingerprints change.
+execute 'Install JavaScript dependencies and pre-compile assets' do
+  cwd release_path
+  environment NODE_ENV: 'production'
+  command "sudo su search -c 'yarn install --production && RAILS_ENV=#{rails_env} bundle exec rake assets:precompile'"
```
ddzz, line 13: JavaScript dependencies are being installed through `yarn install` in this file instead of the cookbooks. As far as I can tell, we'd have to either modify the internals of opsworks_ruby (specifically `deploy[usasearch] action deploy (opsworks_ruby::deploy line 24)`) or modify the order of operations in `before_symlink.rb` in order to run the yarn command in the cookbooks.

MothOnMars: @ddzz , talk to Rob about that. The installation would just need to happen before `opsworks_ruby` runs its deployment code. The [usasearch/recipes/deploy.rb](https://github.com/GSA/usasearch-cookbooks/blob/main/usasearch/recipes/deploy.rb) file contains our custom deployment code. There are several other packages that search-gov depends on that we install via the cookbooks, so there is precedent for the type of installation we'll need to do for `yarn`.  Take a look at:
https://github.com/GSA/usasearch-cookbooks/blob/7ed524ab7d790275095b103cf12d888e29f96f21/usasearch/recipes/packages.rb
https://github.com/GSA/usasearch-cookbooks/blob/main/usasearch/recipes/dependencies.rb

ddzz: It can be done through the cookbooks, and I think it should be done that way, but we'd have to take out the `assets:precompile` from `before_symlink.rb` and run it later during the deploy, which would require its own PR. I think we should keep this as-is for this PR and address this in a future task.

Running `yarn install` is different than the other packages installed through the cookbooks since `search-gov` needs to be cloned first and `yarn install` needs to be run in the repo because it needs the `package.json` in this PR. Right now, `before_symlink` runs immediately after the repo is cloned, and there's no obvious way to add the `yarn install` in between them other than modifying opsworks_ruby itself.

MothOnMars: OK - just leave a comment including the follow-up ticket ID to see if we can move that to the cookbooks.

MothOnMars:
@ddzz , if @yogesh27 approves the suggested Code Climate configuration changes, please make those changes as part of this PR unless there's a reason to make them a separate follow-up story.

MothOnMars (changes requested):
I have some comments about the searches_controller, but I want to get my first round of comments to you before I head out the door. That's the last file I have to noodle a bit.

README.md
```diff
@@ -84,6 +88,13 @@ Use [Bundler](https://bundler.io/) 2.3.8 to install the required gems:
     $ gem install bundler -v 2.3.8
     $ bundle install
 
+### JavaScript dependencies
+
+Use [Yarn](https://classic.yarnpkg.com/en/) to install the required JavaScript dependencies:
+
+    $ npm install --global yarn
+    $ yarn install
```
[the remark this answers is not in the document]
ddzz: I have it in my notes to address this in a future PR.

yogesh27 (commented):
eslint disable changes are looking good to me.

features/searches.feature
```diff
@@ -451,3 +451,11 @@ Feature: Search
...
     Then I should see "Showing results for query"
     And I should see "Search instead for qeury"
+
+  @javascript
+  Scenario: Search with a blank query on an affiliate page (redesign)
+    Given the following Affiliates exist:
+      | display_name     | name             | contact_email         | first_name | last_name |
+      | bar site         | bar.gov          | aff@bar.gov           | John       | Bar       |
+    When I am on bar.gov's redesigned search page with "" query
+    Then I should see "Please enter a search term in the box above."
```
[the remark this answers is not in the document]
ddzz: Thanks for reviewing. I agree that a new feature file should be used. The test will require some tweaking since the search box does not work in the dummy UI; right now, the search query has to be passed in through the URL.

features/support/paths.rb
```diff
@@ -12,6 +12,8 @@ def path_to(page_name)
       search_path
     when /^(.*)'s search page$/
       search_path(:affiliate => $1)
+    when /^(.*)'s redesigned search page with "([^\"]*)" query$/
+      search_path(:affiliate => $1, :query => $2, :redesign => 'true')
```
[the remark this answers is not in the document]
ddzz: This will have to wait until a subsequent story since the search box does not work yet.

app/controllers/searches_controller.rb
```diff
@@ -66,13 +65,15 @@ def advanced
...
-      [I14ySearch, :i14y, :i14y]
+      [I14ySearch, :i14y, :i14y, 'searches']
     elsif @affiliate.gets_blended_results
-      [BlendedSearch, :blended, :blended]
+      [BlendedSearch, :blended, :blended, 'searches']
+    elsif redesign?
+      [WebSearch, :web, :index_redesign, 'searches_redesign']
     else
-      [WebSearch, :web, :index]
+      [WebSearch, :web, :index, 'searches']
```
MothOnMars, line 76: @ddzz , I'm going to approve this for now to facilitate getting this story out the door, but I think we'll want to refactor this method. This method was already doing too many things, and now it's even more overloaded (as-is, it should be renamed to `pick_klass_vertical_template_layout`, as it's now setting 4 different things). I'm thinking that we may want a `layout` method (or `serp_layout`, if `layout` is a protected term). Eventually all the different search engines will be supported by the redesigned serp, so the only logic we should ultimately need is something like:
```ruby
format.html { render template, layout: serp_layout }
...
def serp_layout
  redesign? ? 'searches_redesign' : 'searches'
end
```

ddzz: The plan is to refactor this method as we support more search engines, so these changes are all temporary. I considered an approach similar to the `#serp_layout` method, but I decided against it for now because it would allow the redesigned layout to be used with the old templates for I14y and blended searches.

MothOnMars:
@ddzz , is this ready for re-review? I see additional changes since your re-review request, so I'm wondering if you're still making changes...?

ddzz:
@MothOnMars yes, it's ready for review. I made some changes based on the comments.

app/views/layouts/searches_redesign.html.haml
```diff
@@ -0,0 +1,18 @@
...
+    != "<!-- aid: #{@search.affiliate.id} handle: #{@search.affiliate.name} -->"
+    = favicon_link_tag favicon_url(@search.affiliate.favicon_url)
+
+    %title #{@page_title}
+
+    = javascript_pack_tag 'application'
+    = stylesheet_pack_tag 'application'
+
+  %body
+    #main-content{ tabindex: -1 }= yield
```
MothOnMars, line 18: Per our conversation, please remove any code that isn't explicitly required to satisfy the acceptance criteria for this story. Copying and pasting from the old HTML runs the risk of polluting our new SERP code with outdated/incorrect/untested code. For example, nothing currently tests the display of the favicon, the correct values of the robots meta tag, etc. For now, please pull out anything that's unnecessary and create follow-up stories to address needs (and discuss with @amanifarooque where those stories might land in the PRD). 

As for testing, the cucumber tests are good for testing what the user sees, but either view specs or system specs would be appropriate places to test the under-the-hood elements such as the robots meta tag.

app/helpers/affiliate_helper.rb
```diff
@@ -12,4 +12,8 @@ def render_embed_code_javascript(affiliate)
     JS
     javascript_tag(embed_code)
   end
+
+  def favicon_url(url)
+    url.presence || '/favicon_affiliate.ico'
+  end
```
MothOnMars, line 18: We can leave this for now, but it's a little weird for the `affiliate_helper` to include a method that doesn't involve an affiliate. In my original comment I was thinking we'd pass the affiliate to the method (`favicon_url(affiliate)`). Your method is an improvement, but a better spot for it would be `app/helpers/url_helper.rb`. But again, we can leave it as-is for now.

README.md
```diff
@@ -84,6 +88,13 @@ Use [Bundler](https://bundler.io/) 2.3.8 to install the required gems:
     $ gem install bundler -v 2.3.8
     $ bundle install
 
+### JavaScript dependencies
+
+Use [Yarn](https://classic.yarnpkg.com/en/) to install the required JavaScript dependencies:
+
+    $ npm install --global yarn
```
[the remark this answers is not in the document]
MothOnMars: I added a comment to the ticket - the README changes should be tested by a developer.

MothOnMars (changes requested):
Just cleanup the (currently) unnecessary and untested view code.

app/views/layouts/searches_redesign.html.haml
ddzz: Thanks for pointing this out. I'll bring this up at tomorrow's grooming meeting.

MothOnMars (dismissed):
LGTM, pending CI.

yogesh27 (approved):
LGTM

tsconfig.json
```diff
@@ -0,0 +1,21 @@
+{
```
bmish, line 1: It's a good idea to start off with `"strict": true`.

.circleci/config.yml
```diff
@@ -58,6 +58,14 @@ commands:
             google-chrome --version
             chromedriver --version
           name: Check install
+  install_js_dependencies:
+    description: 'Install JavaScript dependencies'
+    steps:
+      - run: yarn install
```
bmish, line 64: Typically you want to use `yarn install --frozen-lockfile` in CI to ensure no one forgets to check-in lockfile changes.

PR #1065, opened 2022-12-12
SRCH-3614 Add tags filtering to Results API

lorawoodford:
## Summary
- Extends tag-based search filtering to the Results API;
- Individual clarifying comments left in-line in the files, mainly so as to save getting distracted by all the rubocop-inspired changes to existing spec files;
- Instructions on using the Results API at: https://open.gsa.gov/api/searchgov-results/
 
### Checklist
 
#### Functionality Checks
 
#### Process Checks


app/controllers/api/v2/searches_controller.rb
```diff
@@ -93,21 +93,22 @@ def handle_query_routing
       def search_params
         @search_params ||= params.permit(:access_key,
```
lorawoodford, line 94: Alphabetized and left some in-line comments on existing params.  Only real change is addition of `:tags`.

spec/controllers/api/v2/searches_controller_spec.rb
```diff
@@ -276,45 +273,43 @@
...
+                                                      hash_including('query'),
+                                                      be_a_kind_of(ActionDispatch::Request))
 
-        get :i14y, params: search_params
+        get :i14y, params: search_params.merge(tags: 'tag from params')
       end
 
       it { is_expected.to respond_with :success }
 
       it 'passes the correct options to its ApiI4ySearch object' do
```
lorawoodford, line 292: Primary change in the file... many/most other changes are to appease rubocop while I'm in here.

spec/models/i14y_search_spec.rb
```diff
@@ -1,11 +1,39 @@
...
+
 require 'spec_helper'
 
 describe I14ySearch do
-  fixtures :affiliates, :i14y_drawers, :i14y_memberships, :tag_filters
-
   let(:affiliate) { affiliates(:i14y_affiliate) }
   let(:i14y_search) { described_class.new(affiliate: affiliate, query: 'marketplase') }
 
+  describe '#initialize' do
```
lorawoodford, line 9: Primary changes are here and beginning at line 76.  Remaining changes rubocop appeasement.

spec/models/news_search_spec.rb
```diff
@@ -14,16 +14,32 @@
...
     end
 
     it_behaves_like 'an initialized filterable search'
 
     context 'when options does not include sort_by' do
       subject(:search) { described_class.new filterable_search_options }
-      its(:sort_by_relevance?) { should be false }
-      its(:sort) { should eq('published_at:desc') }
+
+      its(:sort_by_relevance?) { is_expected.to be false }
```
lorawoodford, line 27: We weren't currently testing this existing tag-setting behavior, so I added these in to clarify how NewsSearch behaves different from the other FilterableSearches (e.g. i14y and blended).

spec/models/news_search_spec.rb
```diff
@@ -14,16 +14,32 @@
...
+      its(:sort) { is_expected.to eq('published_at:desc') }
+    end
+
+    context 'when show_only_media_content is false' do
```
lorawoodford, line 31: We weren't currently testing this existing tag-setting behavior, so I added these in to clarify how NewsSearch behaves different from the other FilterableSearches (e.g. i14y and blended).

spec/support/filterable_search_behavior.rb
```diff
@@ -1,87 +1,93 @@
+# frozen_string_literal: true
```
lorawoodford, line 1: All changes herein are rubocop-inspired.  I was initially going to add a shared tags param test in here, but discovered the divergence in how NewsSearch set tag filters and moved the tests up into the 3 model specs - `blended_search_spec.rb`, `i14y_search_spec.rb`, and `news_search_spec.rb`.  I didn't want to back out these updates so they remain.

app/models/filterable_search.rb
```diff
@@ -47,6 +49,10 @@ def initialize_date_attributes(options)
     end
   end
 
+  def initialize_facet_attributes(options)
```
ddzz, line 52: Is the plan to expand this method in future PRs to support other facet attributes?

app/models/i14y_search.rb
```diff
@@ -53,6 +51,19 @@ def first_page?
...
+    filter_options[:sort_by_date] = 1 if @sort_by == 'date'
+    filter_options[:min_timestamp] = @since if @since
+    filter_options[:max_timestamp] = @until if @until
+  end
+
+  def included_tags
+    tags = []
+    tags << @affiliate.tag_filters.required.pluck(:tag).sort.join(',') if @affiliate.tag_filters.required.present?
+    tags << @tags if @tags
+    tags.join(',')
```
ddzz, line 64: Does the order of tags matter?

spec/controllers/api/v2/searches_controller_spec.rb
```diff
@@ -276,45 +273,43 @@
...
+        allow(Affiliate).to receive(:find_by_name).and_return(affiliate)
+        allow(ApiI14ySearch).to receive(:new).with(hash_including(query: 'api')).and_return(search)
+        allow(search).to receive(:run)
+        allow(SearchImpression).to receive(:log).with(search,
+                                                      'i14y',
+                                                      hash_including('query'),
+                                                      be_a_kind_of(ActionDispatch::Request))
 
-        get :i14y, params: search_params
+        get :i14y, params: search_params.merge(tags: 'tag from params')
```
ddzz, line 287: Should this be moved to its own context block?

app/models/filterable_search.rb
lorawoodford: Yep, see: https://github.com/GSA/search-gov/pull/1071/files#diff-862beb065f9ac1a20d2b969361d71e64770fd956524da00c39c952cf510e4688 in the other open PR.

app/models/i14y_search.rb
lorawoodford: The order of tags should _not_ matter, but you raise a good question re: why that sort is in there (and in the related expression on ln 35).  For the purposes of this new funcitonality, the goal is just to add the user/searcher driven tag params (`@tags`) to the list of tag filters already present via functionality already in the main codebase (these filters are set globally for all searchers through the admin interface).  I've merely relocated the existing:

`@affiliate.tag_filters.required.pluck(:tag).sort.join(',')`

down into this new method so I can add more terms to it.  I propose leaving this as is for now at least.

spec/controllers/api/v2/searches_controller_spec.rb
lorawoodford: Good call... will refactor.

lorawoodford:
@ddzz This should be ready for re-review.

app/models/i14y_search.rb
ddzz: I think this method needs to be refactored. The `#join` in L62 can be taken out and kept L64; there's no point in calling `#join` twice. Also, in L63, should there be logic to handle `@tags` being a comma-separated string of tags? Is leading/trailing whitespace an issue here?

ddzz (changes requested):

app/models/i14y_search.rb
lorawoodford: I've removed the unnecessary `sort` here and its partner on ln 35, and also removed the redundant `join`.

To the question of white space and further massaging the comma-separated string, that's already (mostly) handled once this query is shipped over to i14y.

Tag params are sent to `lib/serde.rb` (https://github.com/GSA/i14y/blob/main/lib/serde.rb#L9-L11) then to `lib/ext/string` (https://github.com/GSA/i14y/blob/main/lib/ext/string.rb#L3) where the array is split, stripped, and downcased.

ddzz (approved):
LGTM

PR #1076, opened 2022-12-22
SRCH-3556 Disable the twitter list setting in the admin center

krbhavith:
## Summary
- Commented out the twitter list checkbox in admin center, setting cannot be enabled for users in going forward from the view.
 
### Checklist
 
#### Functionality Checks
 
#### Process Checks

### DIFF
--- a/app/views/sites/twitter_profiles/new.html.haml
+++ b/app/views/sites/twitter_profiles/new.html.haml
@@ -11,6 +11,7 @@
   = render_error_messages(@profile)
   = f.label :screen_name, 'Twitter Handle', class: 'required'
   = f.text_field :screen_name, class: 'required input-primary'
-  = label_tag nil, class: 'checkbox' do
-    = check_box_tag :show_lists
-    Show tweets from my lists
+  / Temporarily disabling checkbox as per JIRA ticket #SRCH-3556
+  / = label_tag nil, class: 'checkbox' do
+  /   = check_box_tag :show_lists
+  /   Show tweets from my lists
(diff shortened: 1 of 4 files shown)


app/views/sites/twitter_profiles/new.html.haml
```diff
@@ -11,6 +11,7 @@
   = render_error_messages(@profile)
   = f.label :screen_name, 'Twitter Handle', class: 'required'
   = f.text_field :screen_name, class: 'required input-primary'
-  = label_tag nil, class: 'checkbox' do
-    = check_box_tag :show_lists
-    Show tweets from my lists
+  / Temporarily disabling checkbox as per JIRA ticket #SRCH-3556
```
lorawoodford, line 14: This is a preference call so I won't block on it, but I'd personally change both these comments and those in `app/views/sites/twitter_profiles/index.html.haml` from an html-style comment (`/`) to a ruby comment (`-#`).  That way the comments won't be visible to users in the html of the page/sent to the client, which they probably don't need to be.

lorawoodford (changes requested):
Request that you add a comment to `app/views/sites/twitter_profiles/index.html.haml`, and then one minor optional suggestion regarding the comment style used in the haml files noted inline in `app/views/sites/twitter_profiles/new.html.haml`.

app/views/sites/twitter_profiles/new.html.haml
krbhavith: Updated the suggested changes above

lorawoodford (approved):
LGTM

PR #1079, opened 2022-12-29
SRCH-3617 Site_limits filtering in Results API

lorawoodford:
## Summary
- Extends the ability to add a `site_limits` filter param to a Results API query;
- Note: site limit filter functionality is already part of the main code via the SERP (documentation/release notes: https://search.gov/about/updates/newsletters/may-2019.html, and some existing test coverage for i14y: https://github.com/GSA/search-gov/blob/main/spec/models/i14y_search_spec.rb#L170-L188)'
- This replicates work done in SRCH-3614 and SRCH-3615 to add additional Results API filter params for facet values
- The second commit is purely rubocop cleanup and contains no functional code changes.
 
### Checklist
 
#### Functionality Checks
 
#### Process Checks


MothOnMars (changes requested):
This looks like slightly more code than we need, as I14y searches already support the `sitelimit` parameter with one or more domains or paths. I don't think we want to use two different parameters for the same thing: `sitelimit` in the browser, vs. `site_limits` for API searches. 

I played around with this a bit, and it appears that all we need to do to support a `sitelimit` parameter in the API is to permit the param and change its key to `site_limits` in the search options (which is somewhat annoying, but mirrors what we do for browser-based searches: https://github.com/GSA/search-gov/blob/main/app/controllers/application_controller.rb#L93. It would be nice, but not required, if we could eventually DRY up the search params code between the app controller and the API controller.)

This is my POC branch with the minimal set of changes required to support the `sitelimit` param in the API (not prescriptive code, just a working example):
https://github.com/GSA/search-gov/compare/main...MothOnMars:SRCH-3617-mct-new?expand=1

It ensures that a search with params such as `query=sitemap&sitelimit=search.gov/indexing+search.gov/about` returns the same results via the browser or the search API. 

It would also help to have an integration spec for this behavior, probably in `spec/requests/api/v2/search_spec.rb`. 

Note that neither approach will support sites using indexed documents instead of I14y. I assume we'll call that out on the updated API documentation. It may also makes sense to use different search options validators for `blended` vs `i14y` api searches:
```
# current
when :blended, :i14y, :video then Api::NonCommercialSearchOptions
```
With the faceted search work, we probably shouldn't continue to use the same options for all those search classes, as they won't all work consistently. @amanifarooque, can you chime in on the expected behavior for sites using the results API with indexed documents? Do we need a follow-up story to either support that or to inform those users that the `sitelimit` param isn't supported for those searches?

To summarize the requested changes at this point in time:
- make the necessary changes to support the `sitelimit` parameter instead of the `site_limits` parameter for API users with i14y content
- add an integration spec for that behavior
- back out any unnecessary changes from the search models

amanifarooque:
Re the below comment:

> With the faceted search work, we probably shouldn't continue to use the same options for all those search classes, as they won't all work consistently. @amanifarooque, can you chime in on the expected behavior for sites using the results API with indexed documents? Do we need a follow-up story to either support that or to inform those users that the sitelimit param isn't supported for those searches?

There is no expectation to support this behavior for blended search / ODIE powered sites. The use case for results API users hitting that endpoint is pretty low. We have some items to update results API documentation, we can make all the changes to the various places wholesale after the bulk of API updates are completed.

lorawoodford:
@MothOnMars This should be ready for re-review.  It may still look like a lot of changes, but that's just because I've not backed out the rubocop-only changes introduced in the earlier draft, though it _does_ back out all the meaningful/impactful code changes initially made in those files.

spec/controllers/api/v2/searches_controller_spec.rb
```diff
@@ -349,6 +349,24 @@
...
+        end
+
+        it { is_expected.to respond_with :success }
+
+        it 'removes the sitelimit filter from its ApiI4ySearch object' do
+          expect(assigns(:search_options).attributes).not_to include({ sitelimit: 'nps.gov' })
+        end
+
+        it 'adds a site_limits search param to the ApiI14ySearch' do
+          expect(ApiI14ySearch).to have_received(:new).with(hash_including(site_limits: 'nps.gov'))
```
MothOnMars, line 367: OK to leave these as-is, but we could also include some line breaks for readability:
```ruby
expect(ApiI14ySearch).to have_received(:new).
  with(hash_including(site_limits: 'nps.gov'))
```

MothOnMars (dismissed):
LGTM

lorawoodford:
@MothOnMars I added the line breaks to the spec, so one more approval needed.

MothOnMars (approved):
Lookin' good!

PR #1084, opened 2023-01-03
SRCH-2327 index RSS sitemaps

MothOnMars:
## Summary
 
### Checklist
 
#### Functionality Checks
 
#### Process Checks


MothOnMars:
@lorawoodford , this is the draft PR I mentioned needing a 2nd 🧠 on. A few items to discuss:

- I had originally planned to do a bit of refactoring to decouple the sitemap parsing logic from the sitemap indexing (because the `SitemapIndexer` shouldn't really "know" how to parse the various types of sitemaps), but the minor amount of code required to index RSS sitemaps has me wondering if I should follow the "done is better than perfect" guideline.
- We could use a full integration test for sitemap indexing, but I'm inclined to add that as a separate story or subtask, as the admin team is waiting on this functionality.
- I'm inclined to disable the Rubocop check re. `allow_any_instance_of`. My feeling is that as long as devs use it with care, it's more useful than harmful.

Thoughts?

MothOnMars:
Update: I spent some time today writing a custom Feedjira parser that would allow us to parse XML sitemaps using the same Feedjira API used for RSS sitemaps. Unfortunately, I ran into the same problem as we did long ago (https://cm-jira.usa.gov/browse/SRCH-1524), which is that the resulting code parsed the full sitemap into memory. The memory usage was 10x that of our existing streaming implementation. This makes me lean a little more heavily toward "done but not perfect".

lorawoodford (commented):
@MothOnMars I wanted to make sure to get my initial thoughts down before meetings, so this is an incomplete review, but a start.

> I had originally planned to do a bit of refactoring to decouple the sitemap parsing logic from the sitemap indexing (because the SitemapIndexer shouldn't really "know" how to parse the various types of sitemaps), but the minor amount of code required to index RSS sitemaps has me wondering if I should follow the "done is better than perfect" guideline.

I 100% agree with this, but also strongly agree that now is not the time.  I'd be on board with creating that backlog item, noting that item in comments left behind, and moving on.  This seems prime to be a chore worthy of its own attention anyway because I suspect it will also allow for some spec cleanup/reorganization/piecing out as well.

> We could use a full integration test for sitemap indexing, but I'm inclined to add that as a separate story or subtask, as the admin team is waiting on this functionality.

Yep!  And doing some of the code separation above would undoubtedly help with this.

> I'm inclined to disable the Rubocop check re. allow_any_instance_of. My feeling is that as long as devs use it with care, it's more useful than harmful.

Could we maybe just do this locally in this one instance (especially since, as noted above, we're filing some backlog work to separate, expand, and cleanup the code and specs here)?  I hesitate to do so globally because it feels like encouraging/allowing [legacy practice](https://relishapp.com/rspec/rspec-mocks/v/3-3/docs/working-with-legacy-code/any-instance) to creep in to new code.  It almost always is a sign that too much is happening in a set of tests.  There will be exceptions to this - and its perfectly fine in the code here, but I'm not sold on disabling the cop which might serve for us to do less introspection about _why_ `any_instance` is needed in a particular case.

MothOnMars:
Thanks, @lorawoodford ! I'm going to close this up and open a fresh PR with some tweaks.

PR #1086, opened 2023-01-04
SRCH-2327 index RSS sitemaps

MothOnMars:
## Summary
This PR enables us to support sitemaps in RSS 2.0 format. Previous discussion re. the approach is in https://github.com/GSA/search-gov/pull/1084. 
 
### Checklist
 
#### Functionality Checks
 
#### Process Checks


spec/vcr_cassettes/sitemap_indexer_index/when_the_sitemap_is_an_rss_2/0_feed.yml
```diff
@@ -0,0 +1,101 @@
+---
+http_interactions:
+- request:
+    method: get
+    uri: http://www.whitehouse.gov/robots.txt
```
MothOnMars, line 5: This new cassette was recorded due to the indexer's final call to `index_urls`, which hits the domain's robots.txt file to extract the `delay`:
```ruby
  def index_urls
    index!
    SearchgovDomainIndexerJob.perform_later(searchgov_domain: self, delay: delay)
    ...
 ```
 I used a new domain in that spec simply because we had an existing RSS fixture that I could use for testing (`/rss/wh_blog.xml`).

File: spec/services/sitemap_indexer_spec.rb [hunk 2]
```diff
@@ -14,13 +14,17 @@
...
+      sitemap_url: sitemap_url,
+      domain: searchgov_domain.domain
+    )
+  end
 
   before do
     stub_request(:get, sitemap_url).
       with(headers: { 'User-Agent' => DEFAULT_USER_AGENT }).
       to_return(body: sitemap_content)
-    allow(searchgov_domain).to receive_message_chain(:reload, :index_urls)
```
MothOnMars, line 23: This line wasn't doing anything as written. The `index_urls` call is tested via:
```ruby
      it 'kicks off indexing' do
        allow(SearchgovDomain).to receive(:find_by).
          with(domain: 'agency.gov').and_return(searchgov_domain)
        expect(searchgov_domain).to receive_message_chain(:reload, :index_urls)
        index
      end
  ```

spec/vcr_cassettes/sitemap_indexer_index/when_the_sitemap_is_an_rss_2/0_feed.yml
lorawoodford: Makes sense.  Thanks for the explanation.

lorawoodford (approved):
LGTM 🎉

PR #1091, opened 2023-01-11
SRCH-3674 Implement support for I14y and blended searches for redesigned SERP

ddzz:
## Summary
This PR modifies the SearchesController#index action to load the redesigned SERP for I14y and blended searches.
 
### Checklist
 
#### Functionality Checks
 
#### Process Checks


app/controllers/searches_controller.rb
```diff
@@ -63,23 +67,17 @@ def advanced
 
   private
 
-  # Temporarily disabling this cop until this method is refactored in
-  # upcoming changes for the SERP redesign.
-  # rubocop:disable Metrics/MethodLength
-  def pick_klass_vertical_template_layout
+  def pick_klass_vertical_template
```
ddzz, line 70: This PR reverts temporary changes made to this method in the first SERP redesign PR.

app/controllers/searches_controller.rb
```diff
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class SearchesController < ApplicationController
+  layout :set_layout, only: [:index]
```
ddzz, line 4: For now, the redesigned layout will be limited to the `#index` action. The other actions will be addressed in future PRs, and eventually the `only: [:index]` will be removed.

danswick: Not a super strong opinion, but I tend to prefer to add comments for temporary code.

ddzz: This is actually going to be removed very soon, possibly in the next redesign PR, so I think we're fine without a comment in this case.

PR #1112, opened 2023-01-27
[TEST] github checks test via branch instead of fork

danswick:
Not a real PR. Testing github checks behavior.


danswick:
This works as expected! Circle CI `build_and_test` workflow was triggered and successfully updated.

danswick:
No longer needed. Closing!

PR #1113, opened 2023-01-30
SRCH 3623 - Investigate onrr.gov JS content not indexing

krbhavith:
## Summary
- Added '--disable-blink-features=AutomationControlled' argument to options to prevent the page from assuming automation tool hitting the page.
- Use `sleep(5)` for the page to load completely. 
 
### Checklist
 
#### Functionality Checks
 
#### Process Checks


lorawoodford (approved):
@krbhavith On first blush, using `sleep(5)` seems pretty heavy handed and something we might not want longer term.  I'm guessing you investigated other approaches here before settling on this - adding an explicit wait with some `ExpectedConditions`? - and those didn't work?

If you haven't tried those things, can you timebox a little time to do so and record your findings?

If you already have, can you briefly document what you tried and what wouldn't work in JIRA just so we don't retrace our steps too much if/when we come back to this (which I think we will).  If we have a lot of adoption of this feature I can see the combo of the implicit wait and now this sleep _really_ slowing things down.

Anyway, that said, I'm going to approve for now since this will help resolve an existing issue for us.

krbhavith:
> 

@lorawoodford I have spent some time on this but could not find the proper functions in Selenium webdriver. There are few `wait` functions on conditional basis, but cannot assume what type of elements on each page we are expecting. This needs the adjustments going forward, I need to do proper testing before confirming that. I will document them on Jira. Thank you

PR #1127, opened 2023-02-13
SRCH-3538 basic header, extended header, slim footer and Identifier

yogesh27:
## Summary
- Added components for basic header and extended header
- Added components for slim footer and Identifier
- Added logo to header
- Added secondary links to the basic header
- Added tests for 100% code coverage which will need to be updated when components are integrated with live data
 
### Checklist
 
#### Functionality Checks
 
#### Process Checks


app/javascript/components/Footer/Footer.tsx
```diff
@@ -1,12 +1,64 @@
-import * as React from 'react';
+import React, { useState, useEffect } from 'react';
```
ddzz, line 1: Can we take out `useState` and `useEffect`? I don't see them being used anywhere in this file.

app/javascript/components/Footer/Footer.tsx
```diff
@@ -1,12 +1,64 @@
...
-
+      <UswdsFooter
+        size="slim"
+        returnToTop={returnToTop}
+        primary={
+          <div className="usa-footer__primary-container grid-row">
+            <div className="mobile-lg:grid-col-8">
+              <FooterNav
+                size="slim"
+                links={Array(4).fill(
```
ddzz, line 26: Should we move this to a `const`?

app/javascript/components/Identifier/Identifier.tsx
```diff
@@ -1,12 +1,57 @@
-import * as React from 'react';
+import React, { useState, useEffect } from 'react';
```
ddzz, line 1: Also not seeing the hooks being used in this file.

app/javascript/components/SearchResultsLayout.tsx
```diff
@@ -14,21 +14,26 @@ interface SearchResultsLayoutProps {
...
-      />
-      <Facets />
-      <SearchBar 
-        results={props.results} 
-      />
-      <Results 
-        results={props.results} 
-        vertical={props.vertical}
+        title="Search.gov"
+        isBasic={true}
```
ddzz, line 20: Should we make this a prop instead of using `true`?

app/javascript/test/SearchResultsLayout.test.tsx
```diff
@@ -1,7 +1,52 @@
 import '@testing-library/jest-dom';
-import { render, screen } from '@testing-library/react';
-import * as React from 'react';
+import { render, screen, fireEvent } from '@testing-library/react';
+import React from 'react';
 import SearchResultsLayout from '../components/SearchResultsLayout';
+import {Header} from '../components/Header';
+
+describe('Header', () => {
```
ddzz, line 7: I think we should move `Header` tests to a new file.

ddzz:
I think we're at a point where we should add `eslint-plugin-react`. I'll create a task for this.

app/javascript/components/Footer/Footer.tsx
yogesh27: @ddzz do you mean to move the links to a const?

app/javascript/components/SearchResultsLayout.tsx
yogesh27: @ddzz So title I can get from the affiliate. Currently, we dont have a way to know if the header is basic or extended, so I will put a function for this which can be updated once data is available. Will this be fine?
```
title={props.affiliate.display_name}
isBasic={isBasic(props.affiliate)} 
```

app/javascript/test/SearchResultsLayout.test.tsx
yogesh27: Sure - sounds good.

app/javascript/components/SearchResultsLayout.tsx
ddzz: Actually, since we don't know which header will be used right now, I think it's fine to keep it as `true` for now.

yogesh27:
@ddzz PR is ready to be reviewed.

app/javascript/components/Footer/Footer.tsx
```diff
@@ -1,12 +1,77 @@
...
+            </div>
+            <div className="tablet:grid-col-4">
+              <Address
+                size="slim"
+                items={addressItems}
+              />
+            </div>
+          </div>
+        }
+        secondary={
```
ddzz, line 61: Instead of hiding this with CSS, we can just use `secondary={<></>}`. That way we don't need the CSS, and we can remove the `Logo` import.

ddzz (changes requested):

app/javascript/components/Footer/Footer.tsx
yogesh27: I tried it but it gives a empty gray section like this:
image

ddzz: Ah, so we need the CSS, but we should still remove the `Logo` and use `<></>`.

yogesh27: This has been updated. Kindly review.

app/javascript/components/Header/index.tsx
```diff
@@ -0,0 +1,62 @@
+import React, { useState, useEffect } from 'react';
```
ddzz, line 1:
```suggestion
import React, { useState } from 'react';
```

yogesh27:
PR ready for review.

ddzz (approved):
LGTM

yogesh27:
Hi @ddzz Kindly review the latest changes to this PR.

app/javascript/components/Header/BasicHeader.tsx
```diff
@@ -0,0 +1,66 @@
+import React from 'react';
+import { Header as UswdsHeader, PrimaryNav, Logo, Title, NavMenuButton } from '@trussworks/react-uswds';
+
+import { HeaderProps } from './../props';
+
+import './ExtendedHeader.css';
```
ddzz, line 6: Looks like this import is wrong.

ddzz (changes requested):
Found one issue but looks good otherwise.

ddzz (approved):

PR #1142, opened 2023-03-01
SRCH-3803 AffiliateIndexedDocumentFetcher rescue

lorawoodford:
## Summary
- Response to the comment here: https://github.com/GSA/search-gov/pull/1124/files#r1102881459
- Adds an additional `rescue` to confirm all documents in a batch that can be fetched, will be fetched, even if other documents have disappeared between when the job was queued and when it was executed.
- I opted to add an additional `rescue` rather than move the existing one because - based on interrogating the test currently in main - I think we're trying to (and should) catch two different scenarios here.  First, an instance where the `Affiliate` is gone (that's currently caught in main), but second, an instance where the IndexedDocument itself is gone (that's what this PR addresses).
 
### Checklist
 
#### Functionality Checks
 
#### Process Checks


MyNameIsMissing (dismissed):
LGTM

Admittedly I'm not qualified to review the spec, but I get what the rescue is doing.  Since you'll have to merge main in again before you can merge this, maybe have someone else double check the spec.

spec/jobs/affiliate_indexed_document_fetcher_job_spec.rb
```diff
@@ -61,9 +61,42 @@
...
+    let(:unfetched2) { IndexedDocument.find_by(unfetched_atts2) }
+    let(:unfetched_atts2) do
+      { url: 'http://nps.gov/foo2.html',
+        title: 'Doc Title 2',
+        description: 'This is a document 2.' }
+    end
+
+    before do
+      affiliate.indexed_documents.build(unfetched_atts2)
+      affiliate.save
```
MothOnMars, line 74: I recommend simplifying these two lines as:
```ruby
affiliate.indexed_documents.create!(unfetched_atts2)
```
I prefer to use the bang methods in specs to ensure that an error will be raised if the spec setup is faulty. However, this is just a suggestion, so I'll go ahead and approve this PR.

MothOnMars (dismissed):
LGTM. Thanks for the comment re. the original `rescue`'s purpose.

spec/jobs/affiliate_indexed_document_fetcher_job_spec.rb
lorawoodford: Good call.  Change made.

MothOnMars (approved):
LGTM

PR #1143, opened 2023-03-01
SRCH-3908 redact PII from analytics log lines

MothOnMars:
## Summary
This PR is intended to redact PII such as email addresses and SSNs from the search and click log lines that are read by Logstash and indexed into Elasticsearch. Follow-up stories will redact that information from additional log lines.
 
### Checklist
 
#### Functionality Checks
 
#### Process Checks


lib/redactor.rb
```diff
@@ -0,0 +1,19 @@
...
+  PATTERNS = {
+    ssn: /\b\d{3}[-\sxX._]?\d{2}[-\sxX._]?\d{4}\b/,
+    email: /[a-zA-Z0-9][-a-zA-Z0-9._]+@[-a-zA-Z0-9]+\.+[a-zA-Z]{2,8}/,
+    cc: /\b(?:\d[ -.]*?){13,17}\b/,
+    phone: /\(?([2-9]\d{2})\)?[- .]?([2-9]\d{2}[- .]?\d{4})/
+  }.freeze
+
+  def self.redact(str)
+    str = str.dup
+    PATTERNS.each { |key, value| str.gsub!(value, "[redacted_#{key}]") }
```
MothOnMars, line 16: I tried hard to find a more elegant way to do this, but it gets the job done. I confirmed that the performance impact is negligible when compared to raw logging:
```ruby
> Benchmark.realtime { 10000.times { Rails.logger.info("[Search Impression] #{Redactor.redact(hash.to_json)}") } }
0.8479499999666587
> Benchmark.realtime { 10000.times { Rails.logger.info("[Search Impression] #{hash.to_json}") } }
0.7438500000862405
```

lorawoodford (dismissed):
@MothOnMars Thanks for the thorough discussion here/documentation in the `Redactor` module.  I agree, done is better than perfect at the moment.

LGTM

MothOnMars:
@lorawoodford , this is ready for re-review. The changes address issues found during testing in staging, including:
- URI-encoded characters in PII (such as the "@" in email addresses)
- a too-hungry approach to log line redaction, which resulted in some information being redacted from non-sensitive log components such as user agents.

I wanted to do some additional DRY-ing up of the code, but held off from refactoring at this point. The `SearchImpression#log` and `Click#log` methods are similar but diverge in ways that make refactoring a bit tricky (`SearchImpression` is a module, while `Click` is a class). I'd like to make those more consistent, but this wasn't the story to address that.

spec/lib/redactor_spec.rb
```diff
@@ -0,0 +1,61 @@
+# frozen_string_literal: true
+
+describe Redactor do
+  describe '.' do
```
lorawoodford, line 4: Perhaps

```suggestion
  describe '.redact' do
```

lorawoodford (approved):
One extremely-minor suggestion in line, otherwise approving/LGTM

spec/lib/redactor_spec.rb
MothOnMars: Ack - that was a typo. Fixing...

lorawoodford (approved):
LGTM

PR #1150, opened 2023-03-08
SRCH-3706 Add thumbnail_url to Results API /i14y response

krbhavith:
## Summary
- Add `thumbnail_url` to i14y facets filter.

### Checklist
 
#### Functionality Checks
 
#### Process Checks


lorawoodford (changes requested):
@krbhavith , @amanifarooque I think some wires got crossed here, and these are not the changes I would expect.  While we're doing this `thumbnail_url` work now alongside facets, it is my understanding that the intention is not to treat `thumbnail_url` as a facet/filterable field, correct?  We're just doing these things together since we're already touching parsing behavior to support facets, right?

I would expect these change to be the same, more or less, as those in the linked branch in the JIRA ticket for SRCH-3706: https://github.com/GSA/search-gov/compare/main...lorawoodford:search-gov:SRCH-3675

But, more importantly, I would expect this scope of work to be proceeding in the order laid out in the breakdown in SRCH-3675: https://cm-jira.usa.gov/browse/SRCH-3675. Namely, SRCH-3690 *has* to happen first over in i14y so that `thumbnail_url` is added to the i14y default return fields.  Once that is done, then the two changes in the [branch linked to in SRCH-3706](https://github.com/GSA/search-gov/compare/main...lorawoodford:search-gov:SRCH-3675) are all that should be needed to get `thumbnail_url` included in all I14y Results API responses.  

To sum up:
- thumbnail_url should *always* be returned (unless nil, of course) for i14y Results API searches no matter whether `include_facets` is true or not
- `thumbnail_url` shouldn't be referred to as a `filter_option` (it's not)
- the provided proof-of-concept branch does need a couple tests added, as I didn't do that in my investigation breakdown for SRCH-3675

Happy to chat more or pair if I'm completely wrong in the assumption that this work has nothing to do with facets.

amanifarooque:
@lorawoodford Confirming all your assumptions above; thumbnail_url should be returned for all results / types of searches, if it exists, regardless of whether or not facets are enabled.

lorawoodford (changes requested):
One more quick change: can you go back and update the checklist?  

- This no longer has codeclimate failures
- The summary should be updated... we aren't adding `thumbnail_url` to any filters or facets... just adding it to the response.

Thanks.

lorawoodford (approved):
LGTM... and just one final reminder that this needs to go out at the same time and/or after SRCH-3690.

lorawoodford (approved):
LGTM

PR #1152, opened 2023-03-09
SRCH-3922 Add date filtering to Results API

lorawoodford:
## Summary
- Adds date filtering functionality to the I14y Results API (has no impact on other Results API searches);
- Mild refactoring of `app/models/filterable_search.rb` to break out various bits of logic in `initialize_date_attributes`
- Basic date validation in `lib/api/i14y_search_options.rb`
- New tests
- Rubocop tidying, especially in spec/controllers/api/v2/searches_controller_spec.rb
- See comments in SRCH-3806 investigation for more on the decisions the influenced this implementation
- Acceptance testing is already spelled out in SRCH-3922 if you want to further interrogate the expected behavior
 
### Checklist
 
#### Functionality Checks
 
#### Process Checks

### DIFF
--- a/app/controllers/api/v2/searches_controller.rb
+++ b/app/controllers/api/v2/searches_controller.rb
@@ -96,6 +96,8 @@ module Api
                                          :api_key,
                                          :audience,
                                          :content_type,
+                                         :created_since,
+                                         :created_until,
                                          :cx, # SRCH-1429 This will be removed as GSS is deprecated
                                          :dc,
                                          :enable_highlighting,
@@ -116,7 +118,9 @@ module Api
                                          :searchgov_custom3,
                                          :sitelimit,
                                          :sort_by,
-                                         :tags).to_h
+                                         :tags,
+                                         :updated_since,
+                                         :updated_until).to_h
 
         # Mirrors param key renaming done for browser-based searches.
         # See: app/controllers/application_controller.rb #search_options_from_params
--- a/app/models/filterable_search.rb
+++ b/app/models/filterable_search.rb
@@ -1,12 +1,14 @@
 class FilterableSearch < Search
   TIME_FILTER_PARAMS = %i[hour day week month year].freeze
-  TIME_BASED_SEARCH_OPTIONS = Hash[TIME_FILTER_PARAMS.collect { |p| [p.to_s[0], p] }]
+  TIME_BASED_SEARCH_OPTIONS = TIME_FILTER_PARAMS.index_by { |p| p.to_s[0] }
 
   class_attribute :default_sort_by
   self.default_sort_by = 'r'.freeze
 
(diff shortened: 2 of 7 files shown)


app/controllers/api/v2/searches_controller.rb
```diff
@@ -96,6 +96,8 @@ def search_params
                                          :api_key,
                                          :audience,
                                          :content_type,
+                                         :created_since,
```
ddzz, line 99: I'm wondering if `:created_after` and `:created_before` are better names here?

lorawoodford: @ddzz I'll be the first to admit the naming conventions here need some work (see: SRCH-3924 in the backlog).  I was trying to find a balance between not heavily reworking what is already in the codebase (`since`, `until`, `since_date`, and `until_date`), and still providing parameter names that make it clear what field is being searched.

Ultimately these params are become `created_since_date`, `created_until_date`, `since_date`, and `until_date` (in `lib/api/i14y_search_options.rb`) where they are then passed on to `app/models/filterable_search.rb`.  The FilterableSearch class is all through the models and controllers and keeping all the various logic there for the time being helps ensure that web searching and api searching functionality is the same (at least to the extent that various search options are allowed by the results api and/or made apparent in the web search ui for the various search types).

It's a tangle.  The other consideration is - at least with the minimal validation I've implemented to this point - if an API user hits a validation error in their query at the time being it's going to throw the name of that attribute as defined in `lib/api/i14y_search_options.rb` and passed to `app/models/filterable_search.rb` (so `created_since_date` and `created_until_date` for `created_since` and `created_until` or `since_date` and `until_date` for `updated_since` and `updated_until`).

ddzz: Thanks for the explanation. Since it does look like a tangle, I'm fine with keeping it as-is for now -- hopefully SRCH-3924 will improve the situation.

app/models/filterable_search.rb
```diff
@@ -42,19 +44,35 @@ def default_sort?
...
+    @created_since = parse_since_ts(options[:created_since_date], @created_until)
 
-    if @since && @until && @since > @until
-      @since, @until = @until.beginning_of_day, @since.end_of_day
-    end
+    flip_reversed_dates(@since, @until, @created_since, @created_until)
+
+    return unless (@since.nil? && @until.nil?) || (@created_since.nil? && @created_until.nil?)
 
+    time_based_search(options)
```
ddzz, line 55: Since `#time_based_search` only needs `options[:tbs]`, we could pass only that option into the method instead of the full `options`.

lorawoodford: Good call.

ddzz (approved):
LGTM

PR #1163, opened 2023-03-23
SRCH-3918 Redact cc numbers in searches

lorawoodford:
## Summary
- Redacts portions of search strings that match common credit card patterns, while accounting for the possibility that spaces in the numbers may be URI-encoded (+).
 
### Checklist
 
#### Functionality Checks
 
#### Process Checks


ddzz (approved):
LGTM

lorawoodford:
@ddzz Apologies, but requesting a re-review.  Changes/justifications as follows:
- AT found 1 issue - failure to account for 17-digit Discover cards... this has been addressed in that regex;
- Merging in main (now with SRCH-3919 in place) caused the unintended consequence of having some CC numbers identified as partial phone numbers. In SRCH-3919 I opted to leave out a starting word boundary in the phone number regex (because sometimes they start `+` or `(` which `\b` misses), but the error in that decision was clear as soon as I ran the CC specs.  I've updated the phone number regex to catch an optional `+` or `(`, then word boundary.

Thanks!

ddzz (approved):

PR #1165, opened 2023-03-24
SRCH-4003 bug: invalid thumbnail_url should not block indexing page c…

krbhavith:
## Summary
- Set value to `nil` for invalid `thumbnail_url`, so that it will not block indexing URL.
 
### Checklist
 
#### Functionality Checks
 
#### Process Checks


MothOnMars (changes requested):
@krbhavith , this looks like quite a bit more code than we need. I was thinking we could fix this in the `HtmlDocument#thumbnail_url` method:
```ruby
  def thumbnail_url
    url = metadata['og:image']&.first
    url&.match?(URI::DEFAULT_PARSER.make_regexp) ? url : nil
  end
```
Per our discussion with Amani, that code ensures `thumbnail_url` will return full, valid URLs and return `nil` otherwise.

krbhavith:
I have added code to build absolute url before, now I reverted that and updated as per your suggesstion.

spec/models/html_document_spec.rb
```diff
@@ -213,6 +213,20 @@
 
       it { is_expected.to eq 'http://www.foo.gov/og_image.jpg' }
     end
+
+    context 'when an og:image is not available' do
```
MothOnMars, line 217: Please reword this to clarify what is being tested:
```ruby
context 'when og:image is not a valid URL' do
```

There should also be a spec to test that we return `nil` when there is no `og:image`.

MothOnMars (changes requested):

spec/models/html_document_spec.rb
```diff
@@ -213,6 +213,34 @@
 
       it { is_expected.to eq 'http://www.foo.gov/og_image.jpg' }
     end
+
+    context 'when og:image is not available' do
+      let(:raw_document) do
+        <<~HTML
+          <html lang="en">
+            <head>
+              <meta name="og:image" content="">
```
MothOnMars, line 222: @krbhavith , I was actually hoping for a spec where the `og:image` meta tag isn't present at all. That possibility is the reason for the safe navigation operator, because without the `og:image` tag, this would fail:
```ruby
url.match?(URI::DEFAULT_PARSER.make_regexp)
```
while this would not:
```ruby
url&.match?(URI::DEFAULT_PARSER.make_regexp)
```

MothOnMars (approved):

PR #1175, opened 2023-04-07
SRCH-4075 Update OdieSearch default per_page

lorawoodford:
## Summary
- Adds a `default_per_page` to OdieSearch and ensures that that per_page value is used in odie backfill scenarios;
- Broke out several new methods from `WebSearch.handle_response` as the original was very complex and hard to follow;
- Large amount of restructuring and updates to `web_search_spec.rb`, including confirming these tests fail against current main with the existing bug;
- All rubocop-only changes are in the second commit (quite a few in these legacy files).
 
### Checklist
 
#### Functionality Checks
 
#### Process Checks


app/models/web_search.rb
```diff
@@ -48,46 +48,52 @@ def domains_scope_options
...
+    odie_search = initialize_odie_search(available_search_engine_pages)
+    odie_response = run_odie_search_and_handle_response(odie_search, available_search_engine_pages)
+    return unless odie_response&.total&.zero? && (suggestion = odie_response.spelling_suggestion)
+
+    odie_search_with_spelling_suggestions(available_search_engine_pages, suggestion)
+  end
+
   def initialize_odie_search(available_search_engine_pages, query = nil)
-    odie_search_params = @options.merge(per_page: self.default_per_page,
+    @per_page = OdieSearch.default_per_page
```
lorawoodford, line 70: Primary code change to ensure everything from here on out uses the OdieSearch pagination (20).

spec/models/web_search_spec.rb
```diff
@@ -334,44 +338,77 @@
...
+        let(:page) { 1 }
 
         before do
           ElasticIndexedDocument.recreate_index
 
           bing_api_url = "#{BingV6WebSearch::API_HOST}#{BingV6WebSearch::API_ENDPOINT}"
-          page2_6results = Rails.root.join('spec/fixtures/json/bing_v6/web_search/page2_6results.json').read
-          stub_request(:get, /#{bing_api_url}.*odie backfill page 2/).
-            to_return(status: 200, body: page2_6results)
+          page1_6results = Rails.root.join('spec/fixtures/json/bing_v6/web_search/page1_6results.json').read
```
ddzz, line 349: I noticed this is using Bing V6. Do we need to update these specs to use V7?

spec/models/web_search_spec.rb
```diff
@@ -334,44 +338,77 @@
...
 
-          it 'indicates that there is another page of results' do
-            search.run
-            expect(search.total).to be >= 20
-            expect(search.results.size).to be == 6
-            expect(search.startrecord).to be == 11
-            expect(search.endrecord).to be == 16
+          context 'when returning the first page with commercial results' do
+            it 'indicates via the search.total that there is another page of results' do
+              search.run
```
ddzz, line 370: I think `search.run` could be moved to a `before` block in this `context`.

lorawoodford: Done.

spec/models/web_search_spec.rb
lorawoodford: I was attempting to avoid having to re-stub a BingV7-style response, but you're right... I've updated to V7, deleted the old stubbed json response, and added a new v7 response.

app/models/web_search.rb
ddzz: To keep this consistent with L74, this could be written as `@per_page = odie_search_class.default_per_page`, or we could strip out the `odie_search_class` method entirely. What do you think?

lorawoodford: I like dumping the odie_search_class entirely.  I'm sure there was probably a reason it was like that in the first place, but that reason is probably long gone now.  Updated.

PR #1188, opened 2023-04-18
SRCH-4110 Paginate document collections

lorawoodford:
## Summary
- The bug fix in SRCH-4075 introduced a regression in how pagination is handled within document collections on blended sites that get commercial results and are seeded with supplemental urls;
- The cause of the bug was from being too aggressive in attempting to reduce the complexity of existing odie pagination code in `app/models/web_search.rb`;
- This change reverts re-introduces some code that was present in the pre-4075 main;
- And also adds some cucumber tests to more fully cover this edge case functionality in tests;
- And, finally, adds a step that makes it easier to bulk setup IndexedDocuments for testing with cucumber (necessary here to get enough IndexedDocument results loaded that we can test pagination, but may be helpful in future scenarios as well).
 
### Checklist
 
#### Functionality Checks
 
#### Process Checks


app/models/web_search.rb
```diff
@@ -79,15 +79,18 @@ def run_odie_search_and_handle_response(odie_search, available_search_engine_pag
...
+        paginate_odie_response(odie_search, odie_response, adjusted_total)
       end
       @total = adjusted_total
     end
     odie_response
   end
 
-  def paginate_odie_response(odie_search, odie_response)
+  def paginate_odie_response(odie_search, odie_response, adjusted_total)
+    temp_total = @total
```
lorawoodford, line 90: The primary code change in this PR is that lns 90-96 have been reverted to what they were prior to SRCH-4075 (though this might not be obvious at first since SRCH-4075 broke this code out into its own `paginate_odie_response` method to reduce the complexity of the original `run_odie_search_and_handle_response` method).  See: https://github.com/GSA/search-gov/blob/6f7530171bfd96c65cd682a9dad0fbcfc1bb1842/app/models/web_search.rb#L78-L84

ddzz (approved):
LGTM

PR #1190, opened 2023-04-18
SRCH-3230 use "https" by default for SearchgovDomains

MothOnMars:
## Summary
This PR ensures that `SearchgovDomain`s will use `https` by default. Follow-up tickets will enforce "https" for sitemaps and SearchgovUrl records. 
 
### Checklist
 
#### Functionality Checks
 
#### Process Checks


ddzz:
@MothOnMars I grepped around and found other uses of `http` and not `https` -- I'm assuming those other files don't need to be updated? One that stuck out is `search-gov/app/models/affiliate.rb:332`.

MothOnMars:
@ddzz , this story is deliberately limited in scope to the SearchgovDomain records. Follow-up stories will enforce `https` for other use cases. The line you mentioned is one example, but there are more, such as any model with a `url` or `link` column.

spec/models/searchgov_domain_spec.rb
```diff
@@ -431,79 +404,79 @@
 
     context 'when there is no robots.txt' do
```
ddzz, line 405: Would it make sense to add a `context` for `#sitemap_urls` here for when `http` is used?

MothOnMars: No, because the `sitemap_urls` method will always return `https` URLs:
https://github.com/GSA/search-gov/pull/1190/files#diff-3b12b8e6b938e033ac47da6f7612a003044090621d665af314b50ca29cbcdc57R72

ddzz (approved):
LGTM

PR #1197, opened 2023-04-26
SRCH-4083, SRCH-3962 Pass in total result page count to redesigned SERP and integrate pagination into the UI

ddzz:
## Summary
This PR modifies the results data sent to the redesigned SERP to include the total number of pages of results and whether or not to use the "unbounded" pagination UI component [as shown here](https://designsystem.digital.gov/components/pagination/). Unbounded pagination is used only for Bing searches. These new attributes have been integrated into the new frontend, allowing pagination to use data from the Rails backend.
 
### Checklist
 
#### Functionality Checks
 
#### Process Checks


app/models/web_results_post_processor.rb
```diff
@@ -1,9 +1,10 @@
-class WebResultsPostProcessor
+class WebResultsPostProcessor < PostProcessor
```
MothOnMars, line 1: Creating a `PostProcessor` class isn't necessarily a bad idea. (However, I would give it a less generic name such as `SearchResultsPostProcessor`. It actually seems like `WebResultsPostProcessor` would be a logical parent class for any of the search types that returns standard url/description/title results, but I haven't looked deeply into the logical division between those classes.) There's a faction of Ruby developers who argue that there should never be "abstract" super classes that will never be instantiated directly (I'm not in that faction).

The other way to pet that cat is to create, say, a `SearchResultsPostProcessor` module that you can then mix into the other processor classes. A perfect example of this pattern is the `ResultsRejector` module, which is `include`d in a couple of the post processor classes:
```
git grep ResultsRejector
app/models/i14y_post_processor.rb:  include ResultsRejector
app/models/web_results_post_processor.rb:  include ResultsRejector
lib/results_rejector.rb:module ResultsRejector
```

app/models/web_search.rb
```diff
@@ -138,10 +138,15 @@ def module_tag_for_search_engine
...
   def post_process_results(results)
     post_processor = WebResultsPostProcessor.new(@query, @affiliate, results)
-    @normalized_results = post_processor.normalized_results
+    process_data_for_redesign(post_processor)
     post_processor.post_processed_results
   end
 
+  def process_data_for_redesign(post_processor)
+    @normalized_results = post_processor.normalized_results
+    @total_pages = post_processor.total_pages(@total)
```
MothOnMars, line 147: Again - I would make `total_pages` part of the `normalized_results` to allow the front-end code to rely on a single source of data.

MothOnMars (commented):
added preliminary comments

app/models/search.rb
```diff
@@ -23,7 +23,8 @@ class SearchError < RuntimeError;
               :queried_at_seconds,
               :module_tag,
               :modules,
-              :normalized_results
+              :normalized_results,
+              :total_pages
```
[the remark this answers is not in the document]
ddzz: First, thanks so much for reviewing. I agree that this should move out of the `*Search` classes, and I've taken it out. Moving `total_pages` to `normalized_results` also simplifies the code. Thanks for the idea!

app/models/web_results_post_processor.rb
ddzz: I considered something like this, but it seems not strictly needed at the moment. But that may change as we develop more of the new SERP, so I'll keep this in mind.

app/javascript/components/Pagination/Pagination.tsx
```diff
@@ -8,6 +8,7 @@ import './Pagination.css';
 interface PaginationProps {
   pathname: string
   totalPages: number
+  unboundedResults: boolean
```
ddzz, line 11: `unboundedResults` will be `true` only for Bing searches. [The component is documented here](https://designsystem.digital.gov/components/pagination/).

app/javascript/components/Results/Results.tsx
```diff
@@ -1,27 +1,31 @@
+/* eslint-disable complexity */
```
ddzz, line 1: Due to the significant amount of complexity introduced by Bing pagination being different than pagination for all other types of searches, I've disabled the `complexity` rule for this file. And this may end up being removed as we develop the new SERP further.

app/javascript/components/Results/Results.tsx
```diff
@@ -1,27 +1,31 @@
...
-    thumbnail: {
+    thumbnail?: {
       url: string
     },
     description: string,
-    updatedDate: string,
-    publishedDate: string,
-    thumbnailUrl: string
-  }[];
+    updatedDate: string | null,
```
ddzz, line 19: There are several new `null` types introduced in PR. Some of them should've been added in previous PRs, and some are needed due to how Bing pagination behaves.

app/javascript/components/Results/Results.tsx
```diff
@@ -128,15 +132,27 @@ export const Results = (props: ResultsProps) => {
                 </Grid>
               </GridContainer>
             );
-          })}
+          })) : (
+            <GridContainer className='result search-result-item'>
+              <Grid row>
+                <Grid tablet={{ col: true }}>
+                  <h4>Sorry, no results found for &#39;{props.query}&#39;. Try entering fewer or more general search terms.</h4>
```
ddzz, line 139: This is temporary hard-coded language that will be removed once we introduce internationalization.

app/javascript/components/Results/Results.tsx
```diff
@@ -128,15 +132,27 @@ export const Results = (props: ResultsProps) => {
...
+                </Grid>
+              </Grid>
+            </GridContainer>)}
         </div>
       </div>
-      {totalPages > 0 && 
+      {props.totalPages > 1 ? (
```
ddzz, line 145: This is where the logic gets complicated, all because of Bing. There's no obvious way to tell whether or not there's another page of results for Bing searches. The logic is:
```
if totalPages > 1, show pagination
else if totalPages = 0 or there's only one page of results, do not show pagination
else show pagination (this is a special case for Bing when you reach a page with no results)
```

app/javascript/components/SearchResultsLayout.tsx
```diff
@@ -48,12 +55,21 @@ const SearchResultsLayout = (props: SearchResultsLayoutProps) => {
...
         <SearchBar 
           query={props.params.query}
-          results={props.results} 
-        />
-        <Results 
-          results={props.results} 
-          vertical={props.vertical}
+          results={props.resultsData ? props.resultsData.results : null} 
         />
+        {props.resultsData ? (
```
ddzz, line 60: More complicated logic here due to Bing's complexity.

app/javascript/components/UswdsOverrides/UswdsPagination.tsx
ddzz, line 1: Some context here: we're using the [react-uswds npm package](https://github.com/trussworks/react-uswds) for most of the UI, but we had to extract out the `UswdsPagination` component out of the package since we had to make multiple internal modifications to it to work with our SERP.

app/javascript/components/UswdsOverrides/UswdsPagination.tsx
```diff
@@ -92,7 +94,13 @@ export const UswdsPagination = ({
   const showOverflow = totalPages > maxSlots; // If more pages than slots, use overflow indicator(s)
 
   const middleSlot = Math.round(maxSlots / 2); // 4 if maxSlots is 7
-  const showPrevOverflow = showOverflow && currentPage > middleSlot;
+  let showPrevOverflow; 
+  if (!unboundedResults) {
```
ddzz, line 98: The pagination component from react-uswds doesn't support unbounded pagination, and we're using a modified version of unbounded pagination for Bing searches, so all changes in this file are for Bing pagination.

app/javascript/utils/index.tsx
```diff
@@ -1,6 +1,9 @@
 export const getUriWithParam = (baseUrl: string, urlParam: string, urlParamQuery: string): string => {
   const Url = new URL(baseUrl);
   const urlParams: URLSearchParams = new URLSearchParams(Url.search);
+  if (urlParam === 'query' && urlParams.has('page')) {
```
ddzz, line 4: This was added to handle a bug where you would not return to page 1 of results after entering a new search query.

app/models/blended_search.rb
```diff
@@ -44,40 +44,46 @@ def first_page?
...
-    @modules << "LOVER" << "SPEL" unless self.spelling_suggestion.nil?
-    @modules |= (@govbox_set.modules - %w(NEWS)) if @govbox_set
+    @modules << 'LOVER' << 'SPEL' unless spelling_suggestion.nil?
+    @modules |= (@govbox_set.modules - %w[NEWS]) if @govbox_set
     @modules << modules_in_results if @total > 0
   end
 
   private
 
+  def process_data_for_redesign(post_processor)
```
ddzz, line 76: This new method (called in L53) is the only change in this file. The rest is Rubocop.

app/models/i14y_post_processor.rb
```diff
@@ -14,17 +14,12 @@ def post_process_results
...
-        url: result['link'],
-        description: result['body'],
-        updatedDate: parse_result_date(result['changed']),
-        publishedDate: parse_result_date(result['published_at']),
-        thumbnailUrl: result['thumbnail_url'] || nil
-      }
-    end
+  def normalized_results(total_results)
+    {
+      totalPages: total_pages(total_results),
```
ddzz, line 19: The `#total_pages` method is defined in the new `ResultsPostProcessor` base class.

app/models/news_search.rb
```diff
@@ -63,7 +63,11 @@ def results_to_hash
 
   # TO-DO: Normalize news results in https://cm-jira.usa.gov/browse/SRCH-4071
   def normalized_results
-    @results
+    {
```
ddzz, line 66: This is all temporary/placeholder code that will be changed or moved elsewhere once we integrate news searches in SRCH-4071, so this can be ignored.

package.json
```diff
@@ -14,6 +14,7 @@
     "@babel/preset-typescript": "^7.18.6",
     "@babel/runtime": "^7.20.1",
     "@trussworks/react-uswds": "^3.2.0",
+    "@types/jest": "^29.5.1",
```
ddzz, line 17: This package was added to fix some warnings from TypeScript and is unrelated to the pagination work.

tsconfig.json
```diff
@@ -13,6 +13,7 @@
     "noEmit": true,
     "baseUrl": ".",
     "allowSyntheticDefaultImports": true,
+    "types": ["node", "jest", "@testing-library/jest-dom"],
```
ddzz, line 16: These are more changes to fix some warnings from TypeScript and are unrelated to the pagination work.

app/javascript/components/SearchResultsLayout.tsx
```diff
@@ -48,12 +55,22 @@ const SearchResultsLayout = (props: SearchResultsLayoutProps) => {
...
+          <Results 
+            results={props.resultsData.results}
+            unboundedResults={props.resultsData.unboundedResults}
+            vertical={props.vertical}
+            totalPages={props.resultsData.totalPages}
+            query={props.params.query}
+          />) : (props.params.query || getCurrentPage() > 1) ? (
```
ddzz, line 67: This complicated ternary had to be added due to there being no obvious way to tell if Bing has another page of results.

ddzz: There may be a way to simplify this actually. Will do some testing soon.

ddzz: Good point, I think a comment is needed here to explain the logic. And I looked over the PR again this morning, and there may be a way to simplify this ternary. I'll test it out soon.

I'll get the props ordered.

app/javascript/components/UswdsOverrides/UswdsPagination.tsx
ddzz: We haven't discussed this, but you bring up a good point. In its present form, `UsdsPagination` does what we need without any apparent bugs. I'll bring it up with Yogesh once he's back.

app/models/blended_search.rb
```diff
@@ -44,40 +44,46 @@ def first_page?
...
-      @endrecord = @startrecord + @results.size - 1
-      assign_spelling_suggestion_if_eligible(response.suggestion.text) if response.suggestion.present?
-    end
+    return unless response
+
+    @total = response.total
+    post_processor = ResultsWithBodyAndDescriptionPostProcessor.new(response.results)
+    post_processor.post_process_results
+    @results = paginate(response.results)
+    process_data_for_redesign(post_processor)
```
lorawoodford, line 53: Unless you have a strong preference for this, I mildly prefer:

```suggestion
    @normalized_results = process_data_for_redesign(post_processor)
```

and removing the `@normalized_results =` from the method itself.  Makes it easier to glance down `handle_response` and see all the instance variables.  

But that's just a minor quibble.  If you do like the idea of updating this, do it also in `i14y_search` and `web_search`.

spec/models/results_post_processor_spec.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe ResultsPostProcessor do
+  describe '#total_pages' do
```
lorawoodford, line 6: These tests violate the general 'one expectation per test' principle we try to follow, but any other way of doing this is likely going to overcomplicate (running through an array of possible page count/result combos in one expectation) or bloat the file (setting total pages in `let`s and doing one liner `it`s for the expectations).  It's fine as is, but just highlighting other possible approaches.  I'd probably have a slight preference for something like this, but your call:

```suggestion
  describe '#total_pages' do
    subject(:total_pages) { described_class.new.total_pages(total_results) }

    let(:total_results) { 0 }

    it { is_expected.to eq(0) }

    context 'when there is 1 result' do
      let(:total_results) { 1 }

      it { is_expected.to eq(1) }
    end

    context 'when there are 10 results' do
      let(:total_results) { 10 }

      it { is_expected.to eq(1) }
    end
    
    etc. etc.
```

lorawoodford (commented):
@ddzz Alright, first pass complete.  Other than the few changes I've left in line (probably the most substantive of which is the refactoring of/commenting on that ternary I know you're already working on), this is looking close.

app/models/results_post_processor.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class ResultsPostProcessor
```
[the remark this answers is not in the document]
ddzz: This is an initial implementation, and it may end up changing as we develop more of the SERP. This may be moved into a module. See Martha's comment [here](https://github.com/GSA/search-gov/pull/1197#discussion_r1179410819).

app/models/blended_search.rb
ddzz: I like this idea and will use it. Thanks for the idea!

spec/models/results_post_processor_spec.rb
ddzz: I'll clean this up.

app/javascript/components/SearchResultsLayout.tsx
lorawoodford: 👍 on comment and reordering.

ddzz:
@lorawoodford requesting another review. Comments have been addressed.

lorawoodford (dismissed):
LGTM 🎉

ddzz:
@lorawoodford requesting yet another review, hopefully the last. We've decided not to show pagination when there are no results, and this required only minor changes. See this Slack thread for more details: https://gsa-tts.slack.com/archives/C02JZQYR24U/p1683314232124419

lorawoodford (approved):
LGTM

PR #1203, opened 2023-05-01
SRCH-4170 add hashed_url column to searchgov_urls

MothOnMars:
## Summary
This PR is the first step required to properly de-duplicate `SearchgovUrl` records. This adds a `document_id` column to the `searchgov_urls` table and ensures that the `document_id` is set on record creation. Once this column is in the production table, I'll update the data for the existing records and purge any duplicates. A final PR will set set the `document_id` column as non-nullable and add a unique index on that column.
 
### Checklist
 
#### Functionality Checks
 
#### Process Checks

### DIFF
--- a/app/models/searchgov_url.rb
+++ b/app/models/searchgov_url.rb
@@ -31,6 +31,7 @@ class SearchgovUrl < ApplicationRecord
     allow_blank: true
 
   before_validation :set_searchgov_domain, on: :create
+  before_create :set_document_id
   before_destroy :delete_document
 
   belongs_to :searchgov_domain
@@ -82,10 +83,6 @@ class SearchgovUrl < ApplicationRecord
       end
     end
     save!
-  end
-
-  def document_id
-    Digest::SHA256.hexdigest(url_without_protocol)
   end
 
   private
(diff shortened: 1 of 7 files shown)


lorawoodford (dismissed):
LGTM.

My only question was going to be about the choice of `document_id` for this.  It could be easily confused with the - as-of-yet unused - searchgov_documents table which itself has an id as well as all the other types and ways we reference "documents."  Then I realized this isn't net new, just the implementation is changing.  So, makes sense to stick with `document_id`.

MothOnMars:
@lorawoodford , that's a good point. While it's true that `document_id` isn't new, it's *also* true that our code is getting more and more crowded with various `document`-y things. It may avoid confusion if I use a more distinctive, descriptive name such as `hashed_url`. I'm going to noodle this a bit before moving forward on this story and the follow-up stories that depend on this change.

MothOnMars:
@lorawoodford , this is ready for a re-review. I've renamed the `document_id` column to `hashed_url`. That actually simplifies the deployment of this story, because the existing `document_id` method will continue to work for indexing until I've done the necessary follow-up work referenced in the comments.

lorawoodford (approved):
LGTM

PR #1217, opened 2023-05-15
SRCH-4141 Add fonts and colors section

lorawoodford:
## Summary
 
### Checklist
 
#### Functionality Checks
 
#### Process Checks

### DIFF
--- a/config/locales/dynamic_form.en.yml
+++ b/config/locales/dynamic_form.en.yml
@@ -1,9 +1,10 @@
+---
 en:
   dynamic_form:
+    error_messages:
+      body: 'There were problems with the following fields:'
+      header:
+        one: 1 error prohibited this %{record_name} from being saved
+        other: "%{count} errors prohibited this %{record_name} from being saved"
     errors:
       dynamic_format: "%{message}"
-    error_messages:
-      header:
-        one: "1 error prohibited this %{record_name} from being saved"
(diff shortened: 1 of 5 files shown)


config/locales/dynamic_form.en.yml
```diff
@@ -1,9 +1,10 @@
+---
 en:
   dynamic_form:
-    errors:
-      dynamic_format: "%{message}"
     error_messages:
+      body: 'There were problems with the following fields:'
       header:
-        one: "1 error prohibited this %{record_name} from being saved"
+        one: 1 error prohibited this %{record_name} from being saved
```
MyNameIsMissing, line 7: So, what is the norm for no-quotes, single quotes, and double quotes?

All the other lines shown here have either single quotes or double quotes.  Since this has a variable in it, I would guess it needs double quotes like it originally had, but I don't know for sure.  I just know it seems inconsistent with the other message values shown.

lorawoodford: @MyNameIsMissing This was auto-updated by our internationalization gem `i18n-tasks` and is alright as is.  In these .yml files quotes (single or double) are used for escaping purposes only.  So, the single quotes on line 5, for example, are being used to escape the `:`. The message on ln10 is double quoted because that interpolated string *could* contain special characters (including `'` which can't be escaped inside of a single quote).

MyNameIsMissing (approved):
LGTM

PR #1226, opened 2023-05-24
SRCH-4243 notify Datadog when the search client fails

MothOnMars:
## Summary
This PR ensures that error information is sent to Datadog when the search client is unable to reach the API host. The errors will be tagged with the namespace of the search API (`I14y`, `bing_v6`, etc.). It also ensures that search connection failures (such as 4xx, 5xx reponses from the various search APIs) will no longer be silent failures.

### Checklist
 
#### Functionality Checks
 
- ⚠️  Automated checks pass. If Code Climate checks do not pass, explain reason for failures:
The failure is for too many lines of code in `i14y_search_spec.rb`.  The spec I added isn't absolutely necessary in that class, but I wanted at least one integration spec to cover the full stack. We might also disable the "too many lines" check for specs, as that is more appropriate for app code.
 
#### Process Checks

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -96,16 +96,6 @@ Rails.application.configure do
   # Do not dump schema after migrations.
   config.active_record.dump_schema_after_migration = false
 
-  # Datadog API client used to send exception data to Datadog.
-  datadog_api_config = Rails.application.secrets.datadog
-  if datadog_api_config[:api_enabled]
-    datadog_api_client = Dogapi::Client.new(datadog_api_config[:api_key], datadog_api_config[:application_key])
-    config.middleware.use ExceptionNotification::Rack,
-    datadog: {
-      client: datadog_api_client
-    }
-  end
-
   # Inserts middleware to perform automatic connection switching.
   # The `database_selector` hash is used to pass options to the DatabaseSelector
   # middleware. The `delay` is used to determine how long to wait after a write
--- a/config/secrets.yml.dev
+++ b/config/secrets.yml.dev
@@ -21,7 +21,7 @@ secret_keys: &SECRET_KEYS
     web_subscription_id: FILL_THIS_IN_ONCE_WE_HAVE_A_WEB_SUBSCRIPTION_ID
     image_subscription_id: FILL_THIS_IN_ONCE_WE_HAVE_AN_IMAGE_SUBSCRIPTION_ID
   datadog:
-    api_enabled: false
+    api_enabled: true
     api_key: datadogapikey
     application_key: datadogapplicationkey
   email:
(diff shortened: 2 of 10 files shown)


config/secrets.yml.dev
```diff
@@ -21,7 +21,7 @@ secret_keys: &SECRET_KEYS
     web_subscription_id: FILL_THIS_IN_ONCE_WE_HAVE_A_WEB_SUBSCRIPTION_ID
     image_subscription_id: FILL_THIS_IN_ONCE_WE_HAVE_AN_IMAGE_SUBSCRIPTION_ID
   datadog:
-    api_enabled: false
+    api_enabled: true
```
ddzz, line 24: Wouldn't we want to keep this as `false` for dev secrets?

config/environments/production.rb
```diff
@@ -96,16 +96,6 @@
...
 
-  # Datadog API client used to send exception data to Datadog.
-  datadog_api_config = Rails.application.secrets.datadog
-  if datadog_api_config[:api_enabled]
-    datadog_api_client = Dogapi::Client.new(datadog_api_config[:api_key], datadog_api_config[:application_key])
-    config.middleware.use ExceptionNotification::Rack,
-    datadog: {
-      client: datadog_api_client
-    }
-  end
```
MothOnMars, line 107: I've moved this to its own initializer and enabled it in dev & test environments. This ensures that if there are any issues with the configuration, we'll find out about it during CI runs. (No data will be sent to Datadog unless the developer adds legitimate keys to `config/secrets.yml`.)

ddzz:
Should we increase the `file-lines` limit in the Code Climate config to get Code Climate to pass?

config/secrets.yml.dev
MothOnMars: Nope - see my comments

spec/models/i14y_search_spec.rb
```diff
@@ -215,14 +215,24 @@
...
       i14y_search.run
       expect(Rails.logger).to have_received(:error).with(/I14y search problem/)
     end
+
+    # semi-integration spec to confirm that we send i14y search client errors to Datadog
+    it 'sends the information to datadog' do
+      i14y_search.run
+      expect(a_request(
+        :post,
+        'https://api.datadoghq.com/api/v1/events?api_key=datadogapikey'
```
ddzz, line 233: Wouldn't this test fail locally if the dev has an actual API key set?

MothOnMars: Yes -  that's an example of why devs should copy `config/secrets.yml.dev` into `secrets.yml` before running the specs: https://github.com/GSA/search-gov#service-credentials-how-we-protect-secrets

Devs should only be using production API keys if they need to interact with a production service or if they need to record external interactions in new VCR cassettes. Otherwise, they should be using the dummy values from `config/secrets.yml.dev` (which is what will be used in CircleCI).

PR #1244, opened 2023-06-29
SRCH-4337 update API path

jimkreft:
## Summary
- fix the example urls to include the new gsa API url path
 
### Checklist
 
#### Functionality Checks
 
- [X] You have merged the latest changes from the target branch (usually `main`) into your branch.
  
- [X] Your primary commit message is of the format **SRCH-#### \<description\>** matching the associated Jira ticket.

- [X] PR title is either of the format **SRCH-#### \<description\>** matching the associated Jira ticket (i.e. "SRCH-123 implement feature X"), or **Release - SRCH-####, SRCH-####, SRCH-####** matching the Jira ticket numbers in the release.
 
- [X] Automated checks pass. If Code Climate checks do not pass, explain reason for failures:
 
#### Process Checks

- [X] You have specified at least one "Reviewer".


lorawoodford (approved):
@jimkreft This removes our ability to be able to use/rely on these links in development and staging environments so I was initially going to request changes to support those use cases, but in digging deeper I'm seeing that we've already introduced all kinds of hard-coded inconsistencies like this throughout these md files so I don't want to obstruct getting accurate information out to customers just because this will introduce another inconvenience to devs.

That said, I'm going to put in a backlog ticket to do some additional cleanup/refinement of this (and other) files to hopefully solve these things holistically.  At the very lest - with these hardcoded - we can now remove several instances of the `api_scheme_and_host` variable, but ideally, we can add some logic so that these links are correct in all envs.

jimkreft:
yeah, I agree that there should be another task to clean up these hard-coded things in the content

PR #1281, opened 2023-08-22
SRCH-4387: hiding the no search result text

yogesh27:
## Summary
- hiding the no search result text as we are not using jquery-ui.css which hides this text by default
 
### Checklist
 
#### Functionality Checks
 
#### Process Checks

### DIFF
--- a/app/assets/stylesheets/legacy_sass/sayt/custom.sass
+++ b/app/assets/stylesheets/legacy_sass/sayt/custom.sass
@@ -52,6 +52,14 @@
     opacity: 0
     filter: Alpha(Opacity = 0)
 
+/* Hiding 'No search results.' text 
+ *----------------------------------
+ 
+#usasearch_sayt ~ .ui-helper-hidden-accessible
+  position: absolute !important
+  clip: rect(1px 1px 1px 1px)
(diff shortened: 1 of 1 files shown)


app/assets/stylesheets/legacy_sass/sayt/custom.sass
```diff
@@ -52,6 +52,14 @@
     opacity: 0
     filter: Alpha(Opacity = 0)
 
+/* Hiding 'No search results.' text 
+ *----------------------------------
+ 
+#usasearch_sayt ~ .ui-helper-hidden-accessible
+  position: absolute !important
```
stevenbarragan, line 59: what's the difference with something with `display: none`?

stevenbarragan (approved):
I have a small question but it LGTM

app/assets/stylesheets/legacy_sass/sayt/custom.sass
yogesh27: it's used for accessibility so people with screen readers can read it. By using display: none; you hide it from screen readers as well. So, it's better to move it off the screen with position: absolute;

PR #1269, opened 2023-08-09
SRCH-4142 Set redesigned SERP font in admin center

lorawoodford:
## Summary
- Adds two new font selection fields to the "Font and Colors" section of the admin center redesigned display settings page when "Show Redesign Display Settings" is true (set in Super Admin center);
- Adds new json type column to the affiliates table to store these settings;
- Adds tests and helpers as relevant;
- The final commit stems from running rubocop across existing code in `affiliates_spec.rb` to knock out 185 preexisting issues and includes no real changes relevant to this scope of work.
 
### Checklist
Please ensure you have addressed all concerns below before marking a PR "ready for review" or before requesting a re-review. If you cannot complete an item below, replace the checkbox with the ⚠️ `:warning:` emoji and explain why the step was not completed.
 
#### Functionality Checks
 
- [x] You have merged the latest changes from the target branch (usually `main`) into your branch.
  
- [x] Your primary commit message is of the format **SRCH-#### \<description\>** matching the associated Jira ticket.

- [x] PR title is either of the format **SRCH-#### \<description\>** matching the associated Jira ticket (i.e. "SRCH-123 implement feature X"), or **Release - SRCH-####, SRCH-####, SRCH-####** matching the Jira ticket numbers in the release.
 
- [x] Automated checks pass. If Code Climate checks do not pass, explain reason for failures:
 
#### Process Checks

- [x] You have specified at least one "Reviewer".

PR #1284, opened 2023-08-23
SRCH-4143 Set redesigned SERP colors in admin center

lorawoodford:
## Summary
- Adds new SERP color settings (as defined in mini-PRD) to the "Font and Colors" section of the admin center "Visual design (new)" display settings page when "Show Redesign Display Settings" is true (set in Super Admin center);
- Adds/updates `affiliates.visual_design_json` column as needed;
- Adds tests and test helpers as relevant; and,
- Tightens up some existing fonts testing (originally from #1269).
 
### Checklist
 
#### Functionality Checks
 
#### Process Checks


app/views/sites/visual_designs/_fonts_and_colors.html.haml
```diff
@@ -10,4 +10,12 @@
     options_for_select(Affiliate::USWDS_FONTS.index_by { |font| t(".fonts.#{font}", scope: 'admin_center') },
     render_affiliate_visual_design_value(@site.visual_design_json, :footer_and_results_font_family))
 
+  %fieldset
+    %legend= t('.color_scheme', scope: 'admin_center')
```
MyNameIsMissing, line 14: Should this have a space on both side of the = sign?

MyNameIsMissing (dismissed):
LGTM

One comment.  It obviously passes robocop, but looks odd to me.

app/views/sites/visual_designs/_fonts_and_colors.html.haml
lorawoodford: @MyNameIsMissing We don't do any sort of haml-linting (I may create a story for that, but suspect we have tons of violations at this point so not sure it is worth it), but this is valid [haml](https://github.com/haml/haml/blob/main/REFERENCE.md).  The percent sign denotes the html tag `<legend>`, followed immediately by an `=` which denotes that what comes next should be interpreted then printed (in this case the translation for the section heading).

This results in: `<legend>Color scheme</legend>`

lorawoodford:
@MyNameIsMissing This will need a re-review as I had to make one minor change (see last commit).  

Prepping my staging stack for Acceptance Testing revealed an issue when flagging on "display_redesign_display_settings" on an existing site.  Since validation was running before the `set_visual_design_json` in the `before_save` action, existing sites (that wouldn't have new SERP colors assigned yet) were being flagged as invalid because colors are required.  Moving the default settings assignment from a `before_save` to a `before_validation` hook resolves this issue.

MyNameIsMissing (approved):
LGTM

PR #1298, opened 2023-09-11
SRCH-3990 Prepare Backend Data for Search Page Alert

krbhavith:
## Summary
- Add alert for search page component.
 
### Checklist
 
#### Functionality Checks
 
#### Process Checks

### DIFF
--- a/app/helpers/react_helper.rb
+++ b/app/helpers/react_helper.rb
@@ -10,7 +10,8 @@ module ReactHelper
       relatedSites: related_sites(affiliate.connections, search.query),
       resultsData: search.normalized_results,
       translations: translations(affiliate.locale),
-      vertical: vertical
+      vertical: vertical,
+      alert: affiliate.alert
     }
 
     react_component('SearchResultsLayout', data.compact_blank)
--- a/app/javascript/components/SearchResultsLayout.tsx
+++ b/app/javascript/components/SearchResultsLayout.tsx
@@ -112,6 +112,10 @@ interface SearchResultsLayoutProps {
     label: string;
     link: string;
   }[];
+  alert?: {
+    title: string;
+    text: string;
+  } | null;
 }
 
 // To be updated
@@ -124,7 +128,7 @@ const isBasicHeader = (): boolean => {
   return true;
 };
 
-const SearchResultsLayout = ({ resultsData, additionalResults, vertical, params = {}, translations, currentLocale = 'en', relatedSites = [] }: SearchResultsLayoutProps) => {
+const SearchResultsLayout = ({ resultsData, additionalResults, vertical, params = {}, translations, currentLocale = 'en', relatedSites = [], alert }: SearchResultsLayoutProps) => {
   const i18n = new I18n(translations);
   i18n.defaultLocale = 'en';
   i18n.enableFallback = true;


app/helpers/react_helper.rb
```diff
@@ -29,6 +30,12 @@ def translations(locale)
     I18n.backend.translations.slice(:en, locale.to_sym)
   end
 
+  def search_page_alert(alert)
+    return if alert.text.blank? && alert.title.blank?
```
stevenbarragan, line 34: What do you think about moving  this code from Line 15 `if affiliate.alert.present?` in here like
```ruby
return if !alert || alert.text.blank? && alert.title.blank?
```

spec/helpers/react_helper_spec.rb
```diff
@@ -38,6 +39,54 @@
...
     end
 
+    context 'when alert is present and has text and title' do
+      let(:alert) { instance_double(Alert, text: 'alert_title', title: 'alert_title') }
+
+      it 'includes alert and has text and title' do
+        helper.search_results_layout(search, {}, vertical, affiliate)
+        expect(helper).to have_received(:react_component).with(
+          'SearchResultsLayout',
+          hash_excluding(:alert)
```
stevenbarragan, line 49: Did you mean to use `hash_including` instead? Because right now I don't think the test is testing what is saying it does.

spec/helpers/react_helper_spec.rb
```diff
@@ -38,6 +39,54 @@
...
+        )
+      end
+    end
+
+    context 'when alert is present but text is blank' do
+      let(:alert) { instance_double(Alert, text: '', title: 'alert_title') }
```
stevenbarragan, line 55: Rather than an instance double it's better to use the actual object. In this case I think something like will do the trick.

```ruby
before do
   affiliate.alert.text = ''
end
```

spec/helpers/react_helper_spec.rb
```diff
@@ -38,6 +39,54 @@
...
+        helper.search_results_layout(search, {}, vertical, affiliate)
+        expect(helper).to have_received(:react_component).with(
+          'SearchResultsLayout',
+          hash_excluding(:alert)
+        )
+      end
+    end
+
+    context 'when alert is present but title is blank' do
+      let(:alert) { instance_double(Alert, text: 'alert_title', title: '') }
```
stevenbarragan, line 67: Same comment about using `instace_double`'s, can we use the actual alert from the affiliate?

spec/helpers/react_helper_spec.rb
```diff
@@ -38,6 +39,54 @@
...
+        helper.search_results_layout(search, {}, vertical, affiliate)
+        expect(helper).to have_received(:react_component).with(
+          'SearchResultsLayout',
+          hash_excluding(:alert)
+        )
+      end
+    end
+
+    context 'when alert is not present' do
+      let(:alert) { nil }
```
stevenbarragan, line 79: This doesn't ensure an affiliate doesn't has an alert. This should be done with something like

```ruby
  before do
    affiliate.alert = nil
  end
```

stevenbarragan (changes requested):
I have requested a few changes and I have some comments. Let me know what you think.

spec/helpers/react_helper_spec.rb
```diff
@@ -38,6 +39,56 @@
...
     end
 
+    context 'when affiliate has alert with text and title' do
+      let(:alert) { Alert.new(text: 'Alert text', title: 'Alert title', affiliate_id: affiliate.id) }
+
+      it 'sets alert to contain both text and title' do
+        helper.search_results_layout(search, {}, vertical, alert.affiliate)
+        expect(helper).to have_received(:react_component).with(
+          'SearchResultsLayout',
+          hash_including(:alert)
```
stevenbarragan, line 49: I missed this last time but what do you think about testing that it includes the alert fields as well. e.g.
```ruby
 hash_including(alert: {text: ''Alert text',  title: 'Alert title'})
```

or putting it all together without the `let` statement 

```ruby
  alert_data = {text: ''Alert text',  title: 'Alert title'}
  affiliate.alert.update(alert_data)
  hash_including(alert: alert_data)
```

Same thing for the following test cases

stevenbarragan (approved):
LGTM

PR #1300, opened 2023-09-12
SRCH-4145 Header and footer links in admin center

lorawoodford:
## Summary
- Adds identifier subsection text fields for Parent Agency Name, Parent Agency Link, Identifier Domain Name;
- Adds (repeating) link title and url fields for primary header links, secondary header links, footer links, and identifier links to the visual design settings page if `show_redesign_display_settings` is true;
- Limits secondary header links to 3 link sets;
- Tweaks to anchor links and anchor link styling to confirm links word on render as well as redirect and links account for fixed header heights of admin center;
- Adds db migrations;
- Adds tests and test helpers as necessary;
- Reuses existing table drag-and-drop implementation since other parts of the codebase will continue to rely on it - if we wish to refactor this at some future point this can/should be done holistically across all DnD features.
 
### Checklist
 
#### Functionality Checks
 
#### Process Checks


app/controllers/admin/affiliates_controller.rb
```diff
@@ -7,8 +7,27 @@ class Admin::AffiliatesController < Admin::AdminController
     config.actions.add :field_search
     config.field_search.columns = :id, :name, :display_name, :website
 
+    hidden_columns_regexp = /
```
stevenbarragan, line 10: What do you think about making this a constant `HIDDEN_COLUMS_REGEXP`

app/controllers/admin/affiliates_controller.rb
```diff
@@ -7,8 +7,27 @@ class Admin::AffiliatesController < Admin::AdminController
...
+      _logo|
+      _mappings|
+      parent_agency_link|
+      parent_agency_name|
+      scope_ids|
+      size|
+      use_extended_header)\z/x
+
     attribute_columns = config.columns.reject do |column|
-      column.association or column.name =~ /(_created_at|_updated_at|agency_id|css_properties|content_type|file_name|_image|json|label|_logo|_mappings|scope_ids|size|use_extended_header|visual_design_json)\z/
```
stevenbarragan, line 11: Regex may be too much for this. I suggest to use an array instead and then use something like`HIDDEN_COLUMS.include?(column.name)`

app/controllers/sites/visual_designs_controller.rb
```diff
@@ -14,14 +16,30 @@ def update
...
+  def new_link
+    @index = params[:index].to_i
+    @type = params[:type]
+    respond_to { |format| format.js }
+  end
+
   private
 
+  def build_links
+    @site.primary_header_links = [{}] if @site.primary_header_links.blank?
```
stevenbarragan, line 28: Unless somehow `primary_header_links` ends as an empty string we may be able to do this in here 
```ruby
@site.primary_header_links ||= [{}]
```

app/controllers/sites/visual_designs_controller.rb
```diff
@@ -30,6 +48,14 @@ def site_params
         :header_links_font_family,
         :footer_and_results_font_family,
         color_params
+      ],
+      links_json: [
+        {
+          primary_header_links: [links_attributes],
```
stevenbarragan, line 54: We end up with double array here right? is that expected?
```ruby
...
primary_header_links: [[:position, :title, :url]],
...
```

spec/models/affiliate_spec.rb
```diff
@@ -224,6 +224,109 @@
...
+    describe 'links json' do
+      subject(:affiliate) { described_class.create!(valid_create_attributes.merge(links_attributes)) }
+
+      context 'when links_json is valid' do
+        before do
+          affiliate.save!
+        end
+
+        context 'when no links are provided' do
+          let(:links_attributes) { {} }
```
stevenbarragan, line 236: In general is better to use `Hash.new` rather than `{}` to make the code a little more readable.

app/controllers/sites/visual_designs_controller.rb
stevenbarragan, line 28: Why do we need an empty hash inside an empty array?

db/migrate/20230901144301_add_links_json_to_affiliates.rb
```diff
@@ -0,0 +1,5 @@
+class AddLinksJsonToAffiliates < ActiveRecord::Migration[7.0]
+  def change
+    add_column :affiliates, :links_json, :json
```
stevenbarragan, line 3: Why are we opting for using `json` to store this data instead of a table? Is there an advantage or something we need to be compatible with driving this decision?

My suggestion would be to do it more "the rails" way. We could have a `Link` model with all the related validation logic in there, It'd need to be a polymorphic relationship. I'll be happy to help modeling it out.

stevenbarragan (changes requested):
I added some suggestion but I'm requesting changes to use a table instead of a json field to store then links data. Let me know what you think of if you need help with that.

lorawoodford (commented):
@stevenbarragan Completed my unofficial review and I'm going to push updates that will get tests passing and resolve a few little things:

- avoiding having duplicate tr ids like `new_footer_link`;
- sentence casing the add new links like "Add new identifier link";
- ensuring label text isn't mushed together like "Primaryheaderlink link title 0," but rather "Primary header link link title 0"

Other than these changes, this LGTM, so consider this me passing it back to you for review of my additional commit, and approval.

stevenbarragan (dismissed):
LGTM

lorawoodford:
@stevenbarragan One more round.  In setting up a testing stack this weekend I realized that we weren't handling the use case of: 1. when a link already exists, 2. a user deletes the title and url of that link, then 3. that link should be deleted.  (FYI: I debated going with the more standard approach of adding delete checkboxes for each nested link, but since we were asked to preserve existing functionality and this is the existing functionality I abandoned that idea).

Small update and related tests in the final commit.  Thanks again.

app/models/affiliate.rb
```diff
@@ -584,6 +592,14 @@ def set_managed_links(managed_links_attributes, managed_links)
     end
   end
 
+  def empty_link?(link)
+    empty_link = link['title'].blank? && link['url'].blank?
+    existing_link = link['id'].present?
+    # delete newly empty existing link
+    link[:_destroy] = 1 if empty_link && existing_link
```
stevenbarragan, line 599: The issue with this is that the "destroying a link" logic gets buried into something completely unrelated,`#empty_link?` is now doing too much.

I think we could either have a hidden delete check box that gets checked when the link inputs get emptied. 

Or, in the backend, we can call `mark_for_destruction` on a link if title and link are blank, something like.

```ruby
class Link
  ...
  before_validate :mark_for_destruction, if: has_id_but_empty?

  def has_id_but_empty?
    !new_record? && (!title.present? && !url.present?)
  end
end
```

What do you think?

stevenbarragan (changes requested):
I believe there is still room for improvement in our code quality and readability. Let me know what you think.

app/models/affiliate.rb
```diff
@@ -69,7 +69,6 @@ class Affiliate < ApplicationRecord
 
   accepts_nested_attributes_for :header_logo_attachment, :identifier_logo_attachment, allow_destroy: true
   accepts_nested_attributes_for :header_logo_blob, :identifier_logo_blob
-  accepts_nested_attributes_for :primary_header_links, :secondary_header_links, :footer_links, :identifier_links
```
lorawoodford: Missed earlier, but this was nearly duplicating ln 195.

app/models/affiliate.rb
lorawoodford: Thanks @stevenbarragan .  Much cleaner.  Updates made.

app/models/affiliate.rb
```diff
@@ -584,6 +591,12 @@ def set_managed_links(managed_links_attributes, managed_links)
     end
   end
 
+  def empty_link?(link)
+    return if link['id'].present?
+
+    link['title'].blank? && link['url'].blank?
```
stevenbarragan, line 597: This could be

```ruby
 link['id'].blank? && link['title'].blank? && link['url'].blank?
```

stevenbarragan (dismissed):
LGTM

Thanks for bearing with me and going through my nitpicky comments and opting for improvement of our code quality

stevenbarragan (approved):
LGTM

PR #1312, opened 2023-10-04
SRCH-3667 Include third party tracking codes in the new serp

krbhavith:
## Summary
- add third party tracking code to hosted SERP page
 
### Checklist
 
#### Functionality Checks
 
#### Process Checks

### DIFF
--- a/app/helpers/react_helper.rb
+++ b/app/helpers/react_helper.rb
@@ -72,8 +72,6 @@ module ReactHelper
   end
 
   def external_tracking_code(affiliate)
-    return if affiliate.external_tracking_code.blank?
-
     affiliate.external_tracking_code
   end
 
--- a/app/helpers/react_helper.rb
+++ b/app/helpers/react_helper.rb
@@ -13,6 +13,7 @@ module ReactHelper
       relatedSearches: related_searches(search),
       relatedSites: related_sites(affiliate.connections, search.query),
       resultsData: search.normalized_results,
+      externalTrackingCode: external_tracking_code(affiliate),
       translations: translations(affiliate.locale),
       vertical: vertical
     }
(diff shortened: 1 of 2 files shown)


app/helpers/react_helper.rb
```diff
@@ -70,6 +71,12 @@ def related_sites(connections, query)
     end
   end
 
+  def external_tracking_code(affiliate)
+    return if affiliate.external_tracking_code.blank?
```
stevenbarragan, line 75: Can we just return `affiliate.external_tracking_code` in here? I don't think there is a need for the guard clause because we get the same result without it

spec/helpers/react_helper_spec.rb
```diff
@@ -124,5 +124,20 @@
           with('SearchResultsLayout', hash_including(relatedSearches: [related_search]))
       end
     end
+
+    context 'when affiliate has a non-blank external_tracking_code' do
+      let(:external_tracking_code) { '<script>var foo;</script>' }
+
+      before do
+        allow(affiliate).to receive(:external_tracking_code).and_return(external_tracking_code)
```
stevenbarragan, line 132: There is not need to use a mock in here. This can be done with a regular assignment

stevenbarragan (changes requested):
I have a couple of comments. Let me know what you think.

stevenbarragan (approved):

stevenbarragan (changes requested):
After reading the story description I've realized that the tracking code hasn't been included in the new SERP yet and that we actually don't need React for this.

My suggestion is to include this code directly in the `searches_redesign.html.haml` layout page, at the bottom inside the body tag.

krbhavith:
@stevenbarragan your comments makes sense, so I am using the PR https://github.com/GSA/search-gov/pull/1314 for this change as the code is related. I am closing this PR.

PR #1322, opened 2023-10-13
SRCH-4530 Legacy sidebar links will disappear when Use Redesigned Display is true

krbhavith:
## Summary
- Hide Fonts and Colors, Image Assets, or Header & Footer when `use_redesigned_results_page` is true.

### Checklist
 
#### Functionality Checks
 
#### Process Checks

### DIFF
--- a/app/views/sites/shared/_manage_display_sub_nav.html.haml
+++ b/app/views/sites/shared/_manage_display_sub_nav.html.haml
@@ -2,8 +2,7 @@
   .title Manage Display
 %ul.nav.nav-list
   %li{ site_nav_css_class_hash('displays') }= link_to 'Display Overview', edit_site_display_path(@site)
-  - if @site.use_redesigned_results_page
-    %li{ site_nav_css_class_hash('visual_designs') }= link_to t('sites.visual_designs.edit.visual_design',
+  %li{ site_nav_css_class_hash('visual_designs') }= link_to t('sites.visual_designs.edit.visual_design',
   scope: 'admin_center'), edit_site_visual_design_path(@site)
   - unless @site.use_redesigned_results_page
     %li{ site_nav_css_class_hash('font_and_colors') }= link_to 'Font & Colors', edit_site_font_and_colors_path(@site)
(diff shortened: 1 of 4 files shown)


app/views/sites/shared/_manage_display_sub_nav.html.haml
```diff
@@ -2,10 +2,12 @@
   .title Manage Display
 %ul.nav.nav-list
   %li{ site_nav_css_class_hash('displays') }= link_to 'Display Overview', edit_site_display_path(@site)
-  %li{ site_nav_css_class_hash('visual_designs') }= link_to t('sites.visual_designs.edit.visual_design',
+  - if @site.use_redesigned_results_page
```
lorawoodford, line 5:
```suggestion

```

The `if` here is incorrect.  We need affiliate admins to be able to see and update their visual design settings all the time, not just after they've been opted in to using the redesign.

features/admin_center_dashboard.feature
```diff
@@ -229,6 +229,7 @@ Feature: Dashboard
     Then the "Discover and add the RSS feeds and social media accounts listed on the following page:" field should contain "https://www.nih.gov"
 
     When I follow "Display"
+    And the use redesigned results page is disabled for the site "agencygov"
```
lorawoodford, line 232: Rather than checking that this behavior hasn't changed (if `use_redesigned_results_page` is false that you can still see this one sidebar link), we should be confirming that the sidebar links have actually disappeared for users that *are* opted in to the redesign ( `use_redesigned_results_page=true`).

There's already a setup block in place for this scenario here that can be used for these confirmations: https://github.com/GSA/search-gov/blob/main/features/admin_center_manage_display.feature#L655

features/step_definitions/affiliate_steps.rb
```diff
@@ -88,6 +88,11 @@
   site.update!(is_rss_govbox_enabled: true)
 end
 
+When /^the use redesigned results page is disabled for the site "(.*?)"$/ do |name|
```
lorawoodford, line 91: This seems unnecessary - `use_redesigned_results_page` defaults to false on affiliate creation, and can be easily/explicitly set in an affiliate creation given block like it's done here: https://github.com/GSA/search-gov/blob/391152e26c35d56b48d45c2ffe2fba42121aad99/features/admin_center_manage_display.feature#L657-L658.

lorawoodford (changes requested):
@krbhavith Comments left inline -- main thing is we need to remove the if that would effectively hide the redesign settings from affiliate admins.  They need to be able to see these regardless of whether or not they're using the redesign.

stevenbarragan (approved):
LGTM

lorawoodford (approved):
LGTM.

PR #1331, opened 2023-10-24
SRCH-4451 Use redesigned results page flag

krbhavith:
## Summary
 
### Checklist
 
#### Functionality Checks
 
#### Process Checks


app/controllers/searches_controller.rb
```diff
@@ -140,7 +140,7 @@ def docs_search_klass
   end
 
   def redesign?
```
stevenbarragan, line 142: For better code quality I suggest to move this method to `ApplicationController` and remove it from the controllers.

spec/controllers/searches_controller_spec.rb
```diff
@@ -52,7 +52,8 @@
 
     context 'when searching in English (redesign)' do
       before do
-        get :index, params: { query: 'social security', affiliate: 'usagov', redesign: 'true' }
+        affiliate.update(use_redesigned_results_page: true)
```
stevenbarragan, line 55: It'd be better to have an affiliate with use_redesigned_results_page set to true that we could reuse so we don't have to update it on every test. What do you think about adding a new fixture, something like:

```yml
# spec/fixtures/affiliates.yml

redesigned_usagov:
  <<: *DEFAULTS
  display_name: USA.gov
  name: new-usagov
  website: https://www.usa.gov
  search_engine: BingV7
  use_redesigned_results_page: true
```

and use it instead:

```rb
  let(:affiliate) { affiliates(:redesigned_usagov) }
```

stevenbarragan (changes requested):

stevenbarragan (approved):
LGTM

PR #1335, opened 2023-10-30
SRCH-4442 The existing sidebar links will prepend 'Legacy' to the front when 'show redesign settings' is true

krbhavith:
## Summary
- Display `Legacy Font & Colors`, `Legacy Image Assets` & `Legacy Header & Footer` when `show redesign settings` is **true**.
- Hide sidebar links when `use redesign display` is set to **true**
 
### Checklist
 
#### Functionality Checks
 
#### Process Checks


features/admin_center_manage_display.feature
```diff
@@ -652,14 +652,36 @@ Feature: Manage Display
     Then I should be able to access 3 "new_secondary_header_link" rows
     And I should not see "Add new secondary header link"
 
-  Scenario: Editing the Visual Design Settings when "Show Redesign Display Settings" and "Use Redesigned Results Page" are true
+  Scenario: Display sub navigation links when "Show Redesign Display Settings" and "Use Redesigned Results Page" are true/false
```
lorawoodford, line 655: @krbhavith The changes here are removing some existing/needed test cases.  Can you take a closer look at this and make sure you aren't removing tests for things like the check for the "preview purposes" blurb and confirmation that the "Visual design (new)" page is reached.

lorawoodford (changes requested):
See comment in spec.

features/admin_center_manage_display.feature
```diff
@@ -652,14 +652,51 @@ Feature: Manage Display
...
     Given the following Affiliates exist:
       | display_name | name       | contact_email   | first_name | last_name | show_redesign_display_settings | use_redesigned_results_page |
       | agency site  | agency.gov | john@agency.gov | John       | Bar       | true                           | true                        |
     And I am logged in with email "john@agency.gov"
+    And I go to the agency.gov's Manage Display page
```
lorawoodford, line 660: Is this needed?

features/admin_center_manage_display.feature
```diff
@@ -652,14 +652,51 @@ Feature: Manage Display
...
     When I go to the agency.gov's Visual Design page
     Then I should see "Visual design (new)"
     And the page body should not contain "These settings are for preview purposes only."
-    And I should not see a link to "Fonts & Colors" in the active site sub navigation
+    And I should not see a link to "Font & Colors" in the active site sub navigation
```
lorawoodford, line 664: Typo?

features/admin_center_manage_display.feature
```diff
@@ -652,14 +652,51 @@ Feature: Manage Display
...
     And I should not see a link to "Image Assets" in the active site sub navigation
     And I should not see a link to "Header & Footer" in the active site sub navigation
+    
+    Given the following Affiliates exist:
```
lorawoodford, line 668: Sorry to be nit-picky, but can we make each of these setup blocks separate stand-alone Scenarios?  It is very hard to follow what type of site we're looking at as this stands since you've got to carefully look at the true/false setup data for each affiliate under this Scenario.  The one Scenario that is spelled out for all of these tests says: 

>"Show Redesign Display Settings" and "Use Redesigned Results Page" are true/false

But that's not actually the case for each of these blocks.

lorawoodford (changes requested):

features/admin_center_manage_display.feature
krbhavith: It is `Font and Colors` right now in production.

features/admin_center_manage_display.feature
krbhavith: For other scenarios it is needed as user need to go to 'Display page' to see other sub navigation links. But for this scenario Visual Design page is enough to view the sub navigation links.

lorawoodford (approved):
You'll need to update to main, but otherwise LGTM.

PR #1340, opened 2023-11-01
SRCH-4015 mobile tabs navigation

stevenbarragan:
## Summary
- The tabs navigation is made to behave very different on mobile so I had to move some styles into here
- Rearrange tabs on window resize


yogesh27 (approved):
LGTM

PR #1352, opened 2023-11-13
SRCH-3672 Update referrer-policy on hosted SERP pages

krbhavith:
## Summary
- Set referrer policy to `no-referrer-when-downgrade` on SERP pages.
 
### Checklist
 
#### Functionality Checks
 
#### Process Checks


lorawoodford (changes requested):
@krbhavith We should have test coverage for this to ensure we'll catch it if/when this inadvertently gets removed/broken.  Probably the simplest is to throw something like the following in the [redesign search context of searches_controller_spec](https://github.com/GSA/search-gov/blob/main/spec/controllers/searches_controller_spec.rb#L53-L62):

```
it 'renders the referrer policy' do
    expect(response.body).to match(/meta content='no-referrer-when-downgrade' name='referrer'/)
end
```

You'll want to confirm this works as I'm just spitballing here based one what I'm seeing in the page locally.

lorawoodford (approved):
LGTM

PR #1353, opened 2023-11-14
SRCH-4631 Re-enable redesign=true URL parameter

ddzz:
## Summary
SRCH-4631 Re-enable redesign=true URL parameter
 
### Checklist
 
#### Functionality Checks
 
#### Process Checks


app/controllers/application_controller.rb
```diff
@@ -127,15 +127,24 @@ def set_search_params
     @search_params = ActiveSupport::HashWithIndifferentAccess.new(query: @search.query, affiliate: @affiliate.name)
     @search_params[:sitelimit] = permitted_params[:sitelimit] if permitted_params[:sitelimit].present?
     @search_params[:dc] = permitted_params[:dc] if permitted_params[:dc].present?
+    @search_params[:redesign] = permitted_params[:redesign] if permitted_params[:redesign].present?
```
ddzz, line 130: This line and L165 are the only changes in this PR. The rest are Rubocop fixes.

app/controllers/application_controller.rb
```diff
@@ -153,6 +162,6 @@ def append_info_to_payload(payload)
   end
 
   def redesign?
-    @affiliate.use_redesigned_results_page
+    @affiliate.use_redesigned_results_page or permitted_params[:redesign] == 'true'
```
stevenbarragan, line 165: small improvement but what do you think about using `permitted_params[:redesign].present?` here as well, that way we can set the variable to anything to enable it.

stevenbarragan (approved):
I caught an small improvement but it LGTM

app/controllers/application_controller.rb
ddzz: We have some tests using the hard-coded `true`. This PR would need more changes to use `#present?`.

PR #1366, opened 2023-11-20
SRCH-4577 Update backend for Jobs module for the missing data needed for UI

krbhavith:
## Summary
- Update agency name & abbreviation.
- Send `no_job_results` text when there no jobs in the search results.
 
### Checklist
 
#### Functionality Checks
 
#### Process Checks


spec/models/govbox_set_spec.rb
```diff
@@ -126,6 +126,20 @@
         end
       end
 
+      context 'when there are no job results' do
+        subject(:govbox_set_json) do
+          described_class.new('jobs', affiliate, geoip_info, highlighting_options).as_json
+        end
+
+        before do
+          allow(affiliate).to receive(:jobs_enabled?).and_return(true)
```
stevenbarragan, line 135: There is no need to create a mock in here. Plain assignment should work
```ruby
affiliate.jobs_enabled = true
```

stevenbarragan (changes requested):

app/javascript/components/Results/Jobs/Jobs.tsx
```diff
@@ -59,7 +59,7 @@ export const Jobs = ({ jobs=[] }: JobsProps) => {
 
   return (
     <>
-      {jobs?.length > 0 && (
+      {typeof jobs !== 'string' && jobs?.length > 0 && (
```
stevenbarragan, line 62: Why do we need to validate that `typeof jobs !== 'string'`? when can `jobs` be a string?

app/models/govbox_set.rb
```diff
@@ -138,6 +140,7 @@ def format_graphics_best_bet
 
   def format_jobs
     return unless @affiliate.jobs_enabled?
+    return I18n.t(:no_job_results) if @jobs.blank?
```
stevenbarragan, line 143: I don't think there is not need to return this. The frontend already has all of them and it can display it if jobs is empty.

spec/models/govbox_set_spec.rb
```diff
@@ -126,6 +126,20 @@
...
+        subject(:govbox_set_json) do
+          described_class.new('jobs', affiliate, geoip_info, highlighting_options).as_json
+        end
+
+        before do
+          affiliate.jobs_enabled = true
+        end
+
+        it 'returns no_job_results international text' do
+          expect(govbox_set_json[:jobs]).to eq('No job openings in your region match your search')
```
stevenbarragan, line 139: `jobs` has enough with dealing with two different types already `nil` and `Array`. Let's not add `String` to it. Also the `React` already knows this. I don't think we need to return it here.

stevenbarragan (changes requested):
It looks good but I think we can handle the translation better. Let me know what you think. Sorry I didn't got this before.

app/models/govbox_set.rb
krbhavith: As per the requirement Yogesh wanted me to return something to know when there are no jobs. The reason I am validating jobs above as string is because I am returning the string here which is failing without validation.

stevenbarragan: Do you think returning an empty array would be good enough to know there are no jobs? what about `nil`?

ddzz: If an empty array or `nil` are returned they'll be removed from the JSON due to the `#compact_blank` in `#as_json`.

ddzz: If we go this route (using `I18n.t()` we'll need to update the types/props in the relevant component(s).

stevenbarragan (dismissed):
LGTM

PR #1387, opened 2023-12-19
SRCH-4642 jobs missing data integration

yogesh27:
## Summary
- Jobs header integration
- No Jobs found message integration
 
### Checklist
 
#### Functionality Checks
 
#### Process Checks


app/javascript/components/Results/Jobs/Jobs.tsx
```diff
@@ -115,7 +124,7 @@ export const Jobs = ({ jobs=[] }: JobsProps) => {
               );
             })}
 
-            {moreJobs?.length > 0 && (
+            {moreJobs?.length > 0 && (<>
```
stevenbarragan, line 127: Do we need this empty element?

app/javascript/components/Results/Jobs/Jobs.tsx
```diff
@@ -68,26 +70,33 @@ export const Jobs = ({ jobs=[] }: JobsProps) => {
...
-
+              <Grid col={true} className='jobs-logo-wrapper'>
+                <a className="usajobs-logo" href="https://www.usajobs.gov/">
+                  <img alt="USAJobs.gov" />
+                </a>
+              </Grid>
+            </Grid>
+          </GridContainer>
+          
+          {jobs?.length > 0 ? <>
```
stevenbarragan, line 99: Do we need this empty element?

stevenbarragan (changes requested):
I think might be able to get rid of the "empty tags" but let me know otherwise.

app/javascript/components/Results/Jobs/Jobs.tsx
yogesh27: We did it because multiple child elements need a parent. Let me know if there is a way to get rid of them. I didn't want an unnecessary div tag as a parent wrapper.

stevenbarragan (dismissed):

app/javascript/components/Results/Jobs/Jobs.tsx
stevenbarragan: I didn't see the multiple child

stevenbarragan (approved):

PR #1388, opened 2023-12-20
SRCH-4687 Upgrade node version to 20.10.1

stevenbarragan:
## Summary
- Update circleci configuration to use node 20.10.1
- uswds libraries had to be updated to be compatible with the new version
- New cucumber step to click through navigation. The regular `I follow` was getting confused with other page links.

# NOTE
- It needs to be deployed along with this changes https://github.com/GSA/usasearch-cookbooks/pull/809


ddzz:
@stevenbarragan We will need to upgrade the Node version in the cookbooks as well since staging and production do not use the version specified in this repo.

features/vcr_cassettes/Search_-_redesign/News_search.yml
ddzz, line 1: Is there a reason this cassette is so large? Do we need it?

ddzz:
@stevenbarragan Is there an associated PR for the cookbooks to upgrade Node? This PR alone will not upgrade Node in staging and production; a cookbooks change is needed to do that.

features/vcr_cassettes/Search_-_redesign/News_search.yml
stevenbarragan: I've seen this cassettes being too big but I don't know why is that.

We do need it so test doesn't do actual http requests every time they run.

stevenbarragan:
@ddzz this is the PR with the corresponding change on cookbooks https://github.com/GSA/search-gov/pull/1388

ddzz (dismissed):
LGTM

ddzz (approved):

PR #1390, opened 2023-12-22
SRCH-4707 Fallback to StyleContext defaults for SERP

lorawoodford:
## Summary
- Resolves bug wherein affiliates that have not had any visual design settings saved to the db yet were returning a "white page" redesigned SERP when appending the `redesign=true` parameter to their query;
- Ensures all affiliate styles fallback to the defaults set in `app/javascript/contexts/StyleContext.ts` if nothing is in the db

Note, this also includes an unrelated update to a newly failing test: `spec/models/results_with_body_and_description_post_processor_spec.rb`. Since we're now more than 1 yr, 3 months from the hardcoded date set in the test data, the matcher for "about 12 years ago" would now become "over 12 years ago" (see: https://apidock.com/rails/ActionView/Helpers/DateHelper/distance_of_time_in_words).  Rather than to have to keep changing this matcher regularly, I've now set up the test so that it's consistently 12 years from the date the test is run, and updated the matchers accordingly.
 
### Checklist
 
#### Functionality Checks
 
#### Process Checks


stevenbarragan (approved):
LGTM

PR #1393, opened 2023-12-28
SRCH-4710 Setup images on react

stevenbarragan:
## Summary
- Setup webpacker to load assets from the rails app and use them in react directly

story: https://cm-jira.usa.gov/browse/SRCH-4710


yogesh27 (approved):
LGTM

PR #1399, opened 2024-01-04
SRCH-4714 Set crossorigin to anonymous

stevenbarragan:
## Summary
- Set crossorigin tag on assets to avoid CORS errors.

## NOTE
- The [Docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link#crossorigin) indicate that this will fix the error but since the error is not happening either locally nor on staging it's hard to know if it will actually fix it but regardless we need this for security reasons anyway.


ddzz (approved):
LGTM

PR #1413, opened 2024-01-11
SRCH-4772 Remove Azure

lorawoodford:
## Summary
- Full removal of the long-deprecated Azure search engine, and related code and tests;
- Remaining instances of the Azure-related module codes (`AIMAG` `AWEB` `AZCI` `AZCW`) will come in a subsequent PR;
- BingV5 will be similarly removed in an additional PR;
- Note: A large portion of this PR includes re-recorded or permanently deleted VCR cassettes for both rspec and feature specs, because:
     - Since we're removing a fair chunk of code, I wanted to make sure we caught and removed all unnecessary cassettes at the same time; 
     - During the course of the work on this ticket, I discovered that most of our cassettes were recording with the old, deprecated, and no longer functioning Bing cognitive API, so wanted to re-record them (for why I didn't/couldn't record BingV7 Image searches successfully, see Slack);
     - I also noticed while working through these cassettes that I had improperly set us to globally record exact bytes, which is definitely not necessary and makes our cassettes difficult to deal with.  I've [modified that](https://github.com/GSA/search-gov/pull/1413/files#diff-d0368ac40b9e2814bd2678f957b9b9bbee394be126985b5cf2780b06ee2899f0), so the re-recorded cassettes now contain human-readable responses;
     - Finally, I just wanted to fully confirm old cassettes weren't masking new problems with any existing search behaviors introduced by this removal.
 
### Checklist
 
#### Functionality Checks
 
       ⚠️ 3 pre-existing CodeClimate failures popped stemming from files unchanged as part of this PR.  I suspect this is because I rerecorded some cassettes as part of this work.  Long story short, these are not new findings, all 3 stem from long-existing views that are not modified in this PR, and all 3 will be themselves deleted when the "old" SERP is removed in a few months.  Seeing as these would be nontrivial to fix, I'm requesting we ignore these failures (CC passing is an optional branch protection), and proceed as normal.
 
#### Process Checks


lib/api/v2/search_as_json.rb
```diff
@@ -33,17 +33,6 @@ def as_json_results_to_hash
     @results.collect { |result| as_json_result_hash result }
```
krbhavith, line 33: Is this `as_json_result_hash` calling the method in `Api::V2::NonCommercialSearch`?

lorawoodford: @krbhavith Yes, good question.  I've moved the long remaining `as_json_result_hash` method from `NonCommercialSearch` to this module to make that fact even more obvious.  `CommercialSearch` itself is essentially deprecated as well, but still pops up in places related to the deprecated `ApiDocs` search functionality.  This was already a huge PR so I didn't want to take it any further, but I'm going to confirm there are still investigation tickets around covering the future removal of all that code as well (at which point this should become even simpler).  Thanks.

lorawoodford:
@krbhavith Just another ping on this.  I know it's a lot of changes, so definitely feel free to reassign/ping slack for help if you're swamped.

krbhavith (approved):
LGTM

PR #1414, opened 2024-01-12
SRCH-4771 Remove deprecated search modules

lorawoodford:
## Summary
 
### Checklist
 
#### Functionality Checks
 
#### Process Checks


ddzz (approved):
LGTM

lorawoodford:
@ddzz Acceptance testing revealed some lingering Google-related search modules that we decided to yank now as well.  This then led to new codeclimate violations, so please review the final 2 commits.  Thanks!

lorawoodford:
I know you were out when I updated this, but another (gentle) ping @ddzz

db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.0].define(version: 2024_01_05_215009) do
+ActiveRecord::Schema[7.0].define(version: 2024_01_11_183907) do
```
ddzz, line 13: Is the schema missing any changes? Or is there no table for `SearchModule`?

lorawoodford: There is a search_modules table, however since this migration is removing data (and not a column), the only thing changed as part of this migration is the schema version.  Yes, modifying data via migration is icky and something I generally want to avoid, but seeing as this is how we populated the data in the first place (see for example: https://github.com/GSA/search-gov/blob/main/db/migrate/20151023145023_add_azure_composite_search_modules.rb) and seeing as this is pruning dead module codes from the db I thought this would be the cleanest way to address this.  Future grepping of the codebase for these module codes will show the migration that created them, so there's a certain value add to having it documented that they were subsequently removed with an additional migration.

ddzz (approved):
LGTM

PR #1468, opened 2024-02-16
SRCH-4868 Fix YouTube video timestamp on main search results page

ddzz:
## Summary
SRCH-4868 Fix YouTube video timestamp on main search results page
 
### Checklist
 
#### Functionality Checks
 
#### Process Checks


app/helpers/react_helper.rb
```diff
@@ -216,7 +216,6 @@ def govbox_set_data(search)
     if show_results_format?(affiliate) && !affiliate.display_created_date_on_search_results?
       reject_keys_from_hash(govbox_set_json[:federalRegisterDocuments], 'publication_date') if govbox_set_json[:federalRegisterDocuments].present?
       reject_keys_from_hash(govbox_set_json[:newsItems], 'published_at') if govbox_set_json[:newsItems].present?
-      reject_keys_from_hash(govbox_set_json[:youtubeNewsItems], 'published_at') if govbox_set_json[:youtubeNewsItems].present?
```
ddzz, line 219: I'm not sure why this line was added in the first place, but it's stripping off the `published_at` date for YouTube results on the main search page, causing it to default to `a few seconds ago`.

lorawoodford (dismissed):
@ddzz I'm approving because I agree with your assessment and quick fix here, but if you've got the time, can you throw in a test for this?  In a perfect world, this should've been caught by having adequate test coverage confirming that time stamps are present and accurate when we expect them to be.

features/vcr_cassettes/Search_-_redesign/Search_with_video_results.yml
```diff
@@ -0,0 +1,387 @@
+---
+http_interactions:
+- request:
```
lorawoodford, line 3: @ddzz Sorry, nit-picky, but the first 100ish lines in this recording are unnecessary noise likely recorded as you iteratively worked on this test.  Can you delete this recording then rerun the test to get a clean initial recording for this?  (I just did this here locally without capturing these extraneous dd calls and error messages.)

That said, I'm also going to go ahead an "approve" because this is in no way a blocker, just a preference.  If something is up with your environment that a delete -> rerecord doesn't remedy this, there's no reason to spend more time on it.  We can troubleshoot another day.

lorawoodford (approved):
@ddzz Approving, but recommending a change inline below.   Proceed as you'd like.

features/vcr_cassettes/Search_-_redesign/Search_with_video_results.yml
ddzz: I did do a delete and re-record to remedy a very large cassette. I just tried again and it didn't affect the cassette size, so I don't think it's worth extra time perfecting this. It's a small cassette anyway.

PR #1472, opened 2024-02-21
SRCH-4861 Set all search sites to show_redesign_display_settings=TRUE

krbhavith:
## Summary
- Remove column from DB and code for Show Redesign Display Settings.
 
### Checklist
 
#### Functionality Checks
 
#### Process Checks


features/admin_center_manage_display.feature
```diff
@@ -452,33 +452,23 @@ Feature: Manage Display
...
+    And the page body should not contain "These settings are for preview purposes only."
     And I should see "Fonts & Colors" within the navigation tabs
     And I should not see "Results Format" within the navigation tabs
     And I should see "Image Assets" within the navigation tabs
     And I should see "Header & Footer" within the navigation tabs
 
     Given the following SearchGov Affiliates exist:
-      | display_name    | name       | contact_email   | first_name | last_name | show_redesign_display_settings |
-      | searchgov site  | agency.gov | john@agency.gov | John       | Bar       | true                           |
+      | display_name    | name       | contact_email   | first_name | last_name | use_redesigned_results_page |
```
lorawoodford, line 470: This test setup doesn't match the conditions given in the Scenario description.  I'd break this into two scenarios - one where use_redesigned_results_page is true, and the other where it is false.  For the false condition, you'll only really need to test for the presence of the preview warning message, and the bulk of the tests should be kept under a true condition scenario since, in the future, all sites will be true and that preview message will be permanently removed.

db/migrate/20240221024839_remove_show_redesign_display_settings_from_affiliates.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveShowRedesignDisplaySettingsFromAffiliates < ActiveRecord::Migration[7.0]
+  def change
+    remove_column :affiliates, :show_redesign_display_settings, :boolean
```
lorawoodford, line 3: Can we add in the `default: false` here so that, on rollback, we'd have the same state we started with.

lorawoodford (changes requested):
@krbhavith Two very small changes requested inline.

PR #1473, opened 2024-02-21
SRCH-103: Super Admins should be able to destroy SearchGov domains

luisgmetzger:
## Summary
- This PR addresses the ability for super admins to be able to destroy their searchgov_domains and associated urls as outlined in this ticket: https://cm-jira.usa.gov/browse/SRCH-103
 
### Checklist
 
#### Functionality Checks
 
#### Process Checks


app/controllers/admin/searchgov_domains_controller.rb
```diff
@@ -21,10 +23,34 @@ class Admin::SearchgovDomainsController < Admin::AdminController
...
   end
 
+  def confirm_delete
+    render :delete_domain
+  end
+
+  def delete_domain
+    if destroy_domain_confirmation_valid?
+      enqueue_deletion_job
+      flash_message(:success)
```
stevenbarragan, line 46: nitpick and I know there will be repetition but in this case I'd prefer setting the flash message explicitly. It'd add less lines of code and we'd not have to ignore the translation check on the i18 task.

```
flash[:succes] = I18n.t(:'.flash_delete_success', domain: @searchgov_domain.domain)
```

config/routes.rb
```diff
@@ -228,6 +228,8 @@
     resources :searchgov_domains, concerns: :active_scaffold do
       member do
         post 'reindex'
+        get 'confirm_delete'
+        post 'delete_domain'
```
stevenbarragan, line 232: Shouldn't we able to delete the domain doing a `DELETE` request to the searchgov_domains url from the resources setting above?

spec/jobs/searchgov_domain_destroyer_job_spec.rb
```diff
@@ -1,38 +1,90 @@
...
+        expect(SearchgovUrl.exists?(searchgov_url2.id)).to be false
+      end
+    end
+
+    context "with an acceptable number of URLs" do
+      before do
+        create_batch_of_urls(10_000)
+      end
+
+      it "handles large domains without timing out", :aggregate_failures do
```
stevenbarragan: I do agree that we need to test this but my main concern is that it will drastically increase our testing time. Also do we know if a domain with 10,000 urls is consider a large domain?

I've usually done this kind of performance test separately. We could load a db dump of the urls instead of creating one by one. My succession is to leave this testing and optimization as part of another task and just focus on the deleting feature for now. It seems to me that it requires its own chunk of work.

stevenbarragan (changes requested):
I have some suggestions let me know what you think

app/controllers/admin/searchgov_domains_controller.rb
```diff
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class Admin::SearchgovDomainsController < Admin::AdminController
+  before_action :find_searchgov_domain, only: %i[confirm_delete delete_domain reindex]
```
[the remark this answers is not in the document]
luisgmetzger: @stevenbarragan  - I usually use `set` but code climate yelled at me for prefixing a method with this naming convention.

app/controllers/admin/searchgov_domains_controller.rb
luisgmetzger: @stevenbarragan  - I would agree ... Another change from code climate on this to use locales. If that is fine to ignore, I can do.

spec/controllers/admin/searchgov_domains_controller_spec.rb
```diff
@@ -1,21 +1,57 @@
...
+    activate_authlogic
+    UserSession.create(users(:affiliate_admin))
+  end
+
+  describe 'after_create_save' do
+    it 'sets flash[:info] message correctly' do
+      expect { post :create, params: { record: { domain: 'www.usa.gov' } } }.
+        to change { SearchgovDomain.count }.by(1)
+
+      # Validate flash message
```
stevenbarragan, line 18: This comment seems redundant to me as well and I'm not sure if it'll be helpful in the future. What do you think about removing it?

spec/controllers/admin/searchgov_domains_controller_spec.rb
```diff
@@ -1,21 +1,57 @@
...
+    end
+  end
+
+  describe '#delete_domain' do
+    context 'with correct confirmation text' do
+      it 'enqueues the deletion job and redirects to index' do
+        delete :delete_domain, params: { id: basic_domain.id, confirmation: 'DESTROY DOMAIN' }
+
+        expect(response).to redirect_to(action: :index)
+        expect(flash[:success]).to eq("Deletion has been enqueued for #{basic_domain.domain}")
```
stevenbarragan, line 43: In general is better to validate against the `i18n` translation just because if the text in the translation changes we won't need to update the tests as well. e.g.

```ruby
 expect(flash[:success]).to eq(i18n.t('key.', domain: domain))
```

spec/jobs/searchgov_domain_destroyer_job_spec.rb
```diff
@@ -1,38 +1,67 @@
...
-      it 'destroys the searchgov_urls' do
-        expect { perform }.to change{ SearchgovUrl.count }.by(-2)
+      it 'destroys all associated searchgov_urls' do
+        perform
+        expect(SearchgovUrl.exists?(searchgov_url1.id)).to be false
+        expect(SearchgovUrl.exists?(searchgov_url2.id)).to be false
+      end
+    end
+
+    context 'when there are errors' do
```
stevenbarragan, line 34: We haven't have "dangling" undeleted urls. This is not an issue and not something we need to worry about right now. Just let delete do it's thing and trust the framework in reporting back any error

spec/jobs/searchgov_domain_destroyer_job_spec.rb
```diff
@@ -1,38 +1,67 @@
...
+        end
+      end
+
+      context 'when SearchgovDomain destruction fails' do
+        before do
+          allow(searchgov_domain).to receive(:destroy).and_return(false)
+          allow(searchgov_domain).to receive(:id).and_return(999)
+        end
+
+        it 'logs an error and continues without raising an exception' do
```
stevenbarragan, line 54: I know this is different of what's been done before but we want the exception to be raised and we need to see them. There is no need to add logic to catch and log a failure

stevenbarragan (changes requested):

config/routes.rb
```diff
@@ -228,6 +228,8 @@
     resources :searchgov_domains, concerns: :active_scaffold do
       member do
         post 'reindex'
+        get 'confirm_delete'
+        delete 'delete_domain'
```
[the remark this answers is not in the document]
stevenbarragan: Thinking about how using the delete method from active scaffold directly will look like. We'd need to run this check `destroy_domain_confirmation_valid` in a before delete callback and then run `enqueue_deletion_job` in an after delete callback. Which seems more complicated than having our own delete. You can ignore my comment.

app/jobs/searchgov_domain_destroyer_job.rb
```diff
@@ -1,11 +1,34 @@
...
+  private
+
+  def destroy_associated_records(records, record_type)
+    records.find_each(batch_size: 100) { |record| return false unless destroy_record(record, record_type) }
+
+    true
+  end
+
+  def destroy_record(record, record_type)
+    unless record.destroy
```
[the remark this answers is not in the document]
luisgmetzger: @stevenbarragan  - I see your point, however, the current approach has a specific rationale behind it.

Rails calls destroy on an ActiveRecord object will attempt to destroy the object and its associated dependent records, this will handle hooks (like before_destroy or after_destroy) that throw an :abort symbol, then the destruction process will be halted, no exceptions are raised, and destroy will return false.

In the current implementation, checking the return value of destroy allows it to gracefully handle situations where the record could not be destroyed due to one of these hooks. If the destroy operation were to fail, it would catch it and log an appropriate error message, which is preferable over having an unhandled exception.

spec/jobs/searchgov_domain_destroyer_job_spec.rb
```diff
@@ -1,38 +1,39 @@
...
-  let(:args) {
-    { searchgov_domain: searchgov_domain }
-  }
+  let(:searchgov_domain) { SearchgovDomain.create(domain: 'www.archive.gov', status: '200 OK') }
 
   describe '#perform' do
     it 'requires a searchgov_domain as an argument' do
-      expect{ described_class.perform_now }.
-        to raise_error(ArgumentError, /missing keyword: :?searchgov_domain/)
+      expect { described_class.perform_now }.to raise_error(ArgumentError)
```
stevenbarragan, line 10: Why did you decided to leave the `/missing keyword: :?searchgov_domain/` out?

stevenbarragan (approved):
LGTM 🙏🏽

app/jobs/searchgov_domain_destroyer_job.rb
```diff
@@ -1,11 +1,12 @@
...
   queue_as :searchgov
 
-  def perform(searchgov_domain:)
-    searchgov_domain.searchgov_urls.find_each do |url|
-      url.destroy!
-    end
+  def perform(searchgov_domain)
+    searchgov_domain.searchgov_urls.find_each(&:destroy!)
     searchgov_domain.destroy!
+    Resque::Job.destroy('searchgov',
```
stevenbarragan, line 7: Can you add a test for this change?. Something along the lines of:
* Enqueue indexing with a SearchgovDomainIndexerJob job
* Execute SearchgovDomainDestroyerJob
* Validate queue is empty

stevenbarragan (changes requested):
We need some testing for the last change and we should be good to go

stevenbarragan (approved):
LGTM

app/jobs/searchgov_domain_destroyer_job.rb
stevenbarragan, line 7: After helping with acceptance testing we realized that if we're currently indexing the domain, after  triggering the `SearchgovDomainDestroyerJob` , it stars deleting all the urls form the database but since the indexing hasn't been stopped yet, the SearchgovDomainIndexerJob is adding urls while the SearchgovDomainDestroyerJob is trying to delete them.

Can you please dequeue the job first thing then start deleting the urls?

stevenbarragan (changes requested):
Something came up after acceptance testing. It should be a quick fix.

stevenbarragan (approved):
🙏🏽

stevenbarragan (approved):

PR #1477, opened 2024-02-26
SRCH-589 Stop Indexing

stevenbarragan:
## Summary
- Adds link to stop indexing
- When the link is clicked the domain's activity goes to idle
- SearchgovDomainIndexerJob won't index if job is in idle

### DIFF
--- a/app/models/searchgov_domain.rb
+++ b/app/models/searchgov_domain.rb
@@ -71,6 +71,16 @@ class SearchgovDomain < ApplicationRecord
     urls.presence || ["#{url}sitemap.xml"]
   end
 
+  def stop_indexing!
+    Resque::Job.destroy(:searchgov, 'SearchgovDomainIndexerJob', searchgov_domain: self, delay: delay)
+
+    self.status = 'indexing stopped'
+
+    done_indexing
+
+    save
+  end
+
   private
 
   def robotex
--- a/app/models/searchgov_domain.rb
+++ b/app/models/searchgov_domain.rb
@@ -76,8 +76,6 @@ class SearchgovDomain < ApplicationRecord
   end
 
   def stop_indexing!
-    Resque::Job.destroy(:searchgov, 'SearchgovDomainIndexerJob', searchgov_domain: self, delay: delay)
-
     done_indexing
     self.status = INDEXING_STOPPED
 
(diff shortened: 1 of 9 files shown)


app/models/searchgov_domain.rb
```diff
@@ -22,8 +26,8 @@ class DomainError < StandardError; end
   attr_readonly :domain
   attr_reader :response
 
-  scope :ok, -> { where(status: OK_STATUS) }
-  scope :not_ok, -> { where.not(status: OK_STATUS).or(where(status: nil)) }
+  scope :ok, -> { where(status: GOOD_STATUS) }
+  scope :not_ok, -> { where.not(status: GOOD_STATUS).or(where(status: nil)) }
```
luisgmetzger, line 30: I am not in love with the naming of these scopes. This is not a blocker but since the scope is referring to the status, maybe something like: 

:status_ok
:status_not_ok

These scopes are a bit more clear to me.

app/models/searchgov_domain.rb
```diff
@@ -71,6 +75,13 @@ def sitemap_urls
     urls.presence || ["#{url}sitemap.xml"]
   end
 
+  def stop_indexing!
+    done_indexing
+    self.status = INDEXING_STOPPED
```
luisgmetzger, line 80: do we need to explicitly declare self here?

spec/controllers/admin/searchgov_domains_controller_spec.rb
```diff
@@ -19,4 +21,16 @@
       end
     end
   end
+
+  # rubocop:disable RSpec/AnyInstance
+  describe 'stop_indexing' do
+    let(:domain) { searchgov_domains(:agency_gov) }
+
+    it 'calls stop_indexing! on domain' do
+      expect_any_instance_of(SearchgovDomain).to receive(:stop_indexing!)
```
luisgmetzger, line 30: I usually try to find alternatives to the `expect_any_instance_of` method because of brittleness. Is there an opportunity here to use a mock or stub for this specific instance? 

Not a block either.

luisgmetzger (approved):

spec/controllers/admin/searchgov_domains_controller_spec.rb
stevenbarragan: Me too but in this particular case I couldn't picture how to build the test without it.

app/models/searchgov_domain.rb
stevenbarragan: Yes!, we do. If not it only creates a `status` variable for the method

luisgmetzger (dismissed):

luisgmetzger (approved):

PR #1481, opened 2024-02-28
SRCH-1919 use rel=canonical to avoid indexing duplicate content

krbhavith:
## Summary
- Create canonical url when available on searchgov_url and avoid indexing duplicate content.
 
### Checklist
 
#### Functionality Checks
 
#### Process Checks


lorawoodford (changes requested):
@krbhavith Inline comments left below.  Additionally, please correct the new whitespace error raised by codeclimate.

spec/models/searchgov_url_spec.rb
```diff
@@ -890,6 +890,75 @@
       end
     end
 
+    context 'when url has different canonical_url' do
+      let(:canonical_url) { 'https://www.agency.gov/new.html' }
+
+      before do
+        expect(I14yDocument).not_to receive(:create).
```
lorawoodford, line 897: We should avoid using `expect` in before blocks, and this expectation actually seems unneeded/unrelated.

Also, let's set up a spy here so that we can use the `have_received` syntax in the expectation below.

So, here in the before:

`allow(described_class).to receive(:create)`

spec/models/searchgov_url_spec.rb
```diff
@@ -890,6 +890,75 @@
...
+                    body: html,
+                    headers: { location: canonical_url })
+        stub_request(:get, canonical_url).
+          to_return(status: 200,
+                    body: html,
+                    headers: { content_type: 'text/html' })
+      end
+
+      it 'creates a new url' do
+        expect(described_class).to receive(:create).with(url: canonical_url)
```
lorawoodford, line 910: With the spy above, this then becomes:

```
     fetch
     expect(described_class).to have_received(:create).with(url: canonical_url)
```

spec/models/searchgov_url_spec.rb
```diff
@@ -890,6 +890,75 @@
...
+        fetch
+      end
```
lorawoodford, line 912: We could also add in an additional test to make sure the original/non-canonical url is not fetched:

```
it 'does not fetch the original url' do
     fetch
     expect(described_class).not_to have_received(:create).with(url: url)
end
```

spec/models/searchgov_url_spec.rb
```diff
@@ -890,6 +890,75 @@
...
+      let(:searchgov_document) { searchgov_url.searchgov_document }
+
+      before do
+        allow(searchgov_url).to receive(:document).and_return(searchgov_document)
+        allow(searchgov_document).to receive(:canonical_url).and_return(canonical_url)
+        allow(described_class).to receive(:persisted?).and_return(true)
+      end
+
+      it 'updates last_crawl_status with canonical URL info' do
+        expect { searchgov_url.send(:use_canonical_url) }.
```
lorawoodford, line 926: Let's break this into two separate tests rather than having one test with two expectation.

spec/models/searchgov_url_spec.rb
```diff
@@ -890,6 +890,75 @@
...
+          to raise_error(SearchgovUrl::SearchgovUrlError, "Created canonical url #{canonical_url}")
+        expect(searchgov_url.last_crawl_status).to include("Canonical URL: #{canonical_url}")
+      end
+    end
+
+    describe '#check_canonical_url' do
```
lorawoodford, line 932: We shouldn't be testing private methods like this, and I'm not sure why `searchgov_document` is relevant here.  The remaining tests here will need to be rethought, but I don't want to go into too much detail until we see what the refactor of the `use_canonical_url` looks like.

lorawoodford (changes requested):
@krbhavith Getting there.  I've left feedback below.

app/models/searchgov_url.rb
```diff
@@ -144,6 +142,27 @@ def validate_document
...
 
+  def check_canonical_url
+    document.canonical_url.present? ? use_canonical_url : index_and_update_status
+  end
+
+  def use_canonical_url
+    canonical_url = document.canonical_url
+    raise SearchgovUrlError, "Record creation forbidden to #{canonical_url}" if redirected_outside_domain?(canonical_url)
+
+    searchgov_url = SearchgovUrl.create(url: canonical_url)
```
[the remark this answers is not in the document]
krbhavith: @lorawoodford Do you think its better to call `fetch` again on new canonical_url to save the new parsed document or just index and update status should be enough?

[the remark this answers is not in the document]
lorawoodford: @krbhavith Yes, that makes sense to throw the new/canonical url into the same fetch process as all the rest of the urls.

Also, while we're at it, it would also be great not to have the actual indexing itself buried in the misleadingly named `check_canonical_url` method.  Can we have that method look more like:

```
def check_canonical_url
     return unless document.canonical_url.present?
      
     use_canonical_url
end
```

and then move the `index_and_update_status` back into fetch itself so it is less hidden?  I know this fetch method was a mess before you got here, but we should try to not make it more confusing with these additions.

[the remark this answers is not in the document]
krbhavith: But when I move `index_and_update_status` back to fetch, it is going be called on url when `canonical_url` is created, which we want to avoid that right.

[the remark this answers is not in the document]
lorawoodford: I think that depends on what the rewritten/split out `use_canonical_url` ends up looking like.

app/models/searchgov_url.rb
```diff
@@ -144,6 +152,30 @@ def validate_document
...
+  def use_canonical_url
+    canonical_url = document.canonical_url
+    return if canonical_url.nil?
+
+    check_for_redirection(canonical_url)
+    create_canonical_url(canonical_url)
+  end
+
+  def create_canonical_url(canonical_url)
+    searchgov_url = SearchgovUrl.create!(url: canonical_url)
```
lorawoodford, line 167: What if a `SearchgovUrl` already exists for this canonical url?  (Either because we've been through this fetch/redirection before, or because the canonical url appears elsewhere in the sitemap?).  If we attempt to create this `SearchgovUrl` here we're going to get a Validation error because of the duplicate url, right?

lorawoodford (changes requested):
@krbhavith Comments left in line.

app/models/searchgov_url.rb
krbhavith: Yes, we are gonna get a Validation error here and no duplicate `SearchgovUrl` is created here

lorawoodford: In that case, we will only ever see the `last_crawl_status` pointing to the canonical URL one the very first fetch of that original URL.  Each time reindexing happens (when triggered manually or ever 30(?) days) after that initial run the `last_crawl_status` will only report the (confusing/not helpful) generic validation error.

spec/models/searchgov_url_spec.rb
```diff
@@ -890,6 +890,65 @@
...
+                    headers: { content_type: 'text/html' })
+      end
+
+      it 'does not fetch the original url' do
+        fetch
+        expect(described_class).not_to have_received(:create!).with(url: url)
+      end
+
+      context 'when canonical url already exists' do
+        before do
```
lorawoodford, line 925: The setup for this and what's on line 941 are the same.  They aren't actually different conditions.

spec/models/searchgov_url_spec.rb
```diff
@@ -890,6 +890,65 @@
...
+          allow(canon).to receive(:persisted?).and_return(false)
+        end
+
+        it 'updates last_crawl_status with canonical URL info' do
+          fetch
+          expect(searchgov_url.last_crawl_status).to include("Canonical URL: #{canonical_url}")
+        end
+
+        it 'fetches the canonical url' do
+          expect(searchgov_url).to receive(:fetch)
```
lorawoodford, line 935: Stylistically, this should do the fetch first then expect `to have_received`.  You'll likely have to set up a spy as well.  But before doing any of that, I suggest you take a step back and consider what actually needs to be tested here since currently the two contexts are the same.

lorawoodford (changes requested):
@krbhavith Comments in line.  One small code change, but also a request to really review the purpose of each of the tests.

lorawoodford (approved):
@krbhavith LGTM

PR #1496, opened 2024-03-11
SRCH-4874 Data Migration for Launch Support - ALL sites - Content settings

krbhavith:
## Summary
- Update new SERP content settings data with old content settings

### Checklist
 
#### Functionality Checks
 
#### Process Checks


app/models/content_settings_updater.rb
```diff
@@ -0,0 +1,72 @@
...
+    update_header_links(affiliate)
+    update_footer_links(affiliate)
+    affiliate.save
+  end
+
+  def migrate_paperclip_image_to_active_storage(affiliate)
+    return if affiliate.mobile_logo_file_name.blank?
+
+    logo_url = affiliate.send(:mobile_logo).url.sub(Rails.env, 'production').gsub(' ', '%20')
+    affiliate.header_logo.attach(io: Tempfile.open(logo_url),
```
krbhavith, line 37: @lorawoodford is this the right way to migrate the Paperclip format file to ActiveStorage type. All the date is transfer perfectly, except byte_size (`affiliate.header_logo_blob.byte_size`) which may be reason for the broken logo image on the page.

lorawoodford: @krbhavith Take a look at the official recommendations from the paperclip maintainers on how to migrate existing data from paperclip to ActiveStorage: https://github.com/thoughtbot/paperclip/blob/main/MIGRATING.md. We'll want to follow that lead and not roll our own approach here.

Additionally, we should probably create a separate rake task that only has responsibility for this image data migration as it has the potential to be a rather long-running task.  We will be duplicating all our existing logo images in S3 as part of this task.  We should also pull in @GSA/search-gov-devops in testing/planning for this in case they have any concerns about our ability to handle (temporarily) doubling our S3 storage usage for production logos.  The alt text data update should also be part of this separate task since it relies on the activestorage record being present first.

So:

- One task for header/footer links
- One task for image-related content

I think it would be totally appropriate to split this out into two stories as well (so as not to hold up your links work from going out), but leave that to you since you did the prior investigation and leg work.

krbhavith: We already have ActiveStorage configuration required for this migration before. I was following this implementation https://github.com/thoughtbot/paperclip/blob/main/MIGRATING.md#moving-files-on-a-remote-host-s3-azure-storage-gcs-etc. I will try other approaches for migrating in this documentation.

krbhavith: I have separated out the image content settings from this PR. Image content settings migration will be handled in. [SRCH-4961](https://cm-jira.usa.gov/browse/SRCH-4961)

lorawoodford (commented):
@krbhavith This code LGTM, but I'm not yet hitting "approve" because I'm leaving a question over in JIRA about what the intended functionality is here.  Once that's answered, happy to approve.

lorawoodford (approved):
@krbhavith LGTM

PR #1504, opened 2024-03-18
SRCH-4961 Data Migration for Launch Support - ALL sites - Image Content settings

krbhavith:
## Summary
- Update new SERP image content settings data with old content settings data
 
### Checklist
 
#### Functionality Checks
 
#### Process Checks

### DIFF
--- a/app/models/image_content_updater.rb
+++ b/app/models/image_content_updater.rb
@@ -1,50 +1,57 @@
 # frozen_string_literal: true
 
 class ImageContentUpdater
-    def update(args)
-      ids = args == 'all' ? Affiliate.all.ids : args.split
-      update_image_content(ids)
-    end
-  
-    private
-  
-    def update_image_content(ids)
-      updated_list = []
-      failed_list = []
-  
-      Affiliate.where(id: ids).find_each do |affiliate|
+  def update(args)
+    ids = args == 'all' ? Affiliate.ids : args.split
+    process_images(ids)
+  end
+
+  private
+
+  def process_images(ids)
+    updated_list = []
+    failed_list = []
+    update_image_content(ids, updated_list, failed_list)
+    log_completion(updated_list, failed_list)
+    [updated_list, failed_list]
+  end
+
+  def update_image_content(ids, updated_list, failed_list)
+    time = Benchmark.measure do
+      get_affiliates(ids).find_each do |affiliate|
         update_affiliate_image_content(affiliate)
         updated_list << affiliate.id
       rescue StandardError => e
         failed_list << { affiliate_id: affiliate.id, reason: e.inspect }
       end
-  
-      log_completion(updated_list, failed_list)
-      [updated_list, failed_list]
     end
-  
-    def update_affiliate_image_content(affiliate)
-      migrate_paperclip_image_to_active_storage(affiliate)
-      update_alt_text(affiliate)
-      affiliate.save
-    end
-  
-    def migrate_paperclip_image_to_active_storage(affiliate)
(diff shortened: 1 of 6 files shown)


app/models/image_content_updater.rb
```diff
@@ -0,0 +1,50 @@
...
+      migrate_paperclip_image_to_active_storage(affiliate)
+      update_alt_text(affiliate)
+      affiliate.save
+    end
+  
+    def migrate_paperclip_image_to_active_storage(affiliate)
+      return if affiliate.mobile_logo_file_name.blank?
+  
+      logo_url = affiliate.send(:mobile_logo).url.sub(Rails.env, 'production').gsub(' ', '%20')
+      affiliate.header_logo.attach(io: Tempfile.open(logo_url),
```
krbhavith, line 36: @lorawoodford here I am following the same process what is mentioned in [documentation](https://github.com/thoughtbot/paperclip/blob/main/MIGRATING.md#moving-files-on-a-remote-host-s3-azure-storage-gcs-etc) to migrate the images.

app/models/image_content_updater.rb
```diff
@@ -0,0 +1,50 @@
...
+    def update_affiliate_image_content(affiliate)
+      migrate_paperclip_image_to_active_storage(affiliate)
+      update_alt_text(affiliate)
+      affiliate.save
+    end
+  
+    def migrate_paperclip_image_to_active_storage(affiliate)
+      return if affiliate.mobile_logo_file_name.blank?
+  
+      logo_url = affiliate.send(:mobile_logo).url.sub(Rails.env, 'production').gsub(' ', '%20')
```
lorawoodford, line 35: @krbhavith This seems unnecessarily complex.  What stops us from doing `affiliate.mobile_logo.url`?

krbhavith: Right we can use `affiliate.mobile_logo_url`

app/models/image_content_updater.rb
krbhavith: Still `active_record_attachment` and `active_record_blob` is created properly with all the values except the `byte_size`, size is always 0.

lorawoodford (commented):
@krbhavith This is definitely on the right track.  I may be paranoid, but I also think it would be wise to add some additional feedback/confirmation steps and logging, particularly I'm wondering about:

- Should we printing to the console what Paperclip bucket we're working against before continuing the task?  We want to be very careful when we're testing that we're working against the `dgsearch-images-dev` bucket, but want to be equally assured that we're working against `dgsearch-images` when we run it for real in production.
- Can we anticipate any additional logging needs beyond just success/fail?  Do we want to know how long this takes?
- How many production affiliates have a logo?  If there's a sizable number lacking a logo, we'd likely be better off filtering them out in the initial query before getting into the `migrate_paperclip_image_to_active_storage` method.

Again, your call to make here.  Once the codeclimate issues and other comment is addressed, I'm happy to do a formal review.  Ping me with questions.

app/models/image_content_updater.rb
lorawoodford: @krbhavith Looks like all this need is changing

```
affiliate.header_logo.attach(io: Tempfile.open(logo_url),
```

to:
```
affiliate.header_logo.attach(io: URI.open(logo_url),
```

lorawoodford (approved):
@krbhavith This LGTM.  As a nitpick, it should be fairly straightforward to stub a response from S3 so we don't get those meaningless/error-filled VCR cassettes, but I don't see that as a reason to delay this any further.  If you have time, though, it'd be a nice to do.

PR #1505, opened 2024-03-18
SRCH-4865 Good bye yui-compressor and legacy scss

stevenbarragan:
## Summary
- Transpile all legacy scss files into regular css
- Migrate sites.css to sites.less to be processed by the `less` compiler directly
  - Copy some css files from gems into the vendor folder to make them available for the `less` compiler
  - Remove legacy scss files
- Remove the yui-compressor gem

## Notes

There is a ton of codeclimate issues but there is no point on going to fix them all since it's still legacy code


luisgmetzger (dismissed):
@stevenbarragan  - This looks to be good. I am not sure why we are using both CSS and LESS but you might have more context on this.

ddzz (approved):

PR #1524, opened 2024-04-02
SRCH-5032 send client_id instead of id_token

stevenbarragan:
## Summary
- Stop sending id_token_hint to logout of login.gov
- Send client_id instead


lorawoodford (approved):
@stevenbarragan LGTM.

PR #1528, opened 2024-04-02
SRCH-4957 Setup search-gov production's Dockerfile

stevenbarragan:
## Summary
- Create Dockerfile for production
- Rename current Dockerfile to Dockerfile.dev

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -68,7 +68,7 @@ USER 1000:1000
 
 ENTRYPOINT ["/rails/bin/docker-entrypoint"]
 
-ENV PORT=${PORT:-80}
+ENV PORT=${PORT:-3000}
 EXPOSE ${PORT}
 
 # Start the server by default, this can be overwritten at runtime
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 ARG RUBY_VERSION=3.0.6
-FROM registry.docker.com/library/ruby:$RUBY_VERSION-slim as base
+FROM public.ecr.aws/docker/library/ruby:$RUBY_VERSION-slim as base
 
 # Rails app lives here
 WORKDIR /rails
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,40 +1,75 @@
-FROM --platform=linux/amd64 ruby:3.0.6
-RUN rm /bin/sh && ln -s /bin/bash /bin/sh
-WORKDIR /usr/src/app
-EXPOSE 3100
+ARG RUBY_VERSION=3.0.6
+FROM registry.docker.com/library/ruby:$RUBY_VERSION-slim as base
 
-RUN apt-get update -y && apt-get install -y --no-install-recommends \
-  default-jre \
-  default-mysql-client \
-  imagemagick \
-  libprotobuf-dev \
-  protobuf-compiler 
+# Rails app lives here
+WORKDIR /rails
 
-# Workaround for PhantomJS: https://github.com/DMOJ/online-judge/pull/1270
-ENV OPENSSL_CONF /etc/ssl/
+# Install base packages
+RUN apt-get update -qq && \
+    apt-get install --no-install-recommends -y \
+    curl libjemalloc2 libcurl4-openssl-dev default-libmysqlclient-dev
 
-ENV NODE_VERSION 16.18.1
-ENV NVM_DIR=/root/.nvm
+# Set production environment
+ENV RAILS_ENV="production" \
+    BUNDLE_DEPLOYMENT="1" \
+    BUNDLE_PATH="/usr/local/bundle" \
+    BUNDLE_WITHOUT="development"
 
+# Install JavaScript dependencies | Needed in base due coffee-rails gem
+ARG NODE_VERSION=16.20.2
+ARG YARN_VERSION=latest
+ENV PATH=/usr/local/node/bin:$PATH
+RUN curl -sL https://github.com/nodenv/node-build/archive/master.tar.gz | tar xz -C /tmp/ && \
+    /tmp/node-build-master/bin/node-build "${NODE_VERSION}" /usr/local/node
+
+# Throw-away build stage to reduce size of final image
+FROM base as build
+
+# Install packages needed to build gems and node modules
+RUN apt-get install --no-install-recommends -y build-essential git libvips pkg-config
+
+# Install JavaScript dependencies
+RUN npm install -g yarn@$YARN_VERSION && rm -rf /tmp/node-build-master
+
+# Install application gems
+COPY Gemfile Gemfile.lock ./
+RUN bundle install && \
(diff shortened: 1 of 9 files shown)


Dockerfile
```diff
@@ -1,40 +1,75 @@
-FROM --platform=linux/amd64 ruby:3.0.6
-RUN rm /bin/sh && ln -s /bin/bash /bin/sh
-WORKDIR /usr/src/app
-EXPOSE 3100
+ARG RUBY_VERSION=3.0.6
+FROM registry.docker.com/library/ruby:$RUBY_VERSION-slim as base
```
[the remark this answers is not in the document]
stevenbarragan: @ddzz it's not necessary. Maybe we can switch in the future if we find it beneficial but for now we're following rails conventions.

Dockerfile
```diff
@@ -1,40 +1,78 @@
...
-ENV OPENSSL_CONF /etc/ssl/
+# Install base packages
+RUN apt-get update -qq && \
+    apt-get install --no-install-recommends -y \
+    curl libjemalloc2 libcurl4-openssl-dev default-libmysqlclient-dev
 
-ENV NODE_VERSION 16.18.1
-ENV NVM_DIR=/root/.nvm
+# Set production environment
+ENV RAILS_ENV="production" \
```
luisgmetzger, line 13: I would recommend grouping all of the environment variables for the Dockerfile here: 

# Set all environment variables before hand to help reduce Docker layers and configure bundler.
ENV RAILS_ENV=production \
    BUNDLE_WITHOUT=development:test \
    BUNDLE_PATH=/usr/local/bundle \
    BUNDLE_DEPLOYMENT=1 \
    NODE_VERSION=16.20.2 \
    YARN_VERSION=latest \
    PATH=/usr/local/node/bin:$PATH

Dockerfile
```diff
@@ -1,40 +1,78 @@
...
+FROM registry.docker.com/library/ruby:$RUBY_VERSION-slim as base
 
-RUN apt-get update -y && apt-get install -y --no-install-recommends \
-  default-jre \
-  default-mysql-client \
-  imagemagick \
-  libprotobuf-dev \
-  protobuf-compiler 
+# Rails app lives here
+WORKDIR /rails
```
luisgmetzger, line 5: I would recommend that we install all server packages together at one time rather than separate times. Probably after creating the working directories we can do this: 

# Install system dependencies
RUN apt-get update -qq && apt-get install --no-install-recommends -y \
    curl \
    libjemalloc2 \
    libcurl4-openssl-dev \
    default-libmysqlclient-dev \
    # Clean up in the same layer to reduce image size
    && apt-get clean \
    && rm -rf /var/lib/apt/lists/*
    
    
We can also add apt-get clean and removed the apt cache (/var/lib/apt/lists/*) in the same RUN command where packages are installed to further reduce image size.

Dockerfile
```diff
@@ -1,40 +1,78 @@
...
+# Install JavaScript dependencies
+RUN npm install -g yarn@$YARN_VERSION && rm -rf /tmp/node-build-master
+
+# Install application gems
+COPY Gemfile Gemfile.lock ./
+RUN bundle install && \
+    rm -rf ~/.bundle/ "${BUNDLE_PATH}"/ruby/*/cache "${BUNDLE_PATH}"/ruby/*/bundler/gems/*/.git && \
+    bundle exec bootsnap precompile --gemfile
+
+# Install node modules
```
luisgmetzger, line 40: Maybe we can modify this a bit more, like this: 

# Install Node.js and Yarn
RUN curl -sL https://github.com/nodenv/node-build/archive/master.tar.gz | tar xz -C /tmp/ && \
    /tmp/node-build-master/bin/node-build "${NODE_VERSION}" /usr/local/node && \
    npm install -g yarn@"${YARN_VERSION}" && \
    rm -rf /tmp/* /usr/local/share/.cache
    
    
 This would remove unnecessary caching by cleaning up after Node.js and Yarn installations.

Dockerfile
```diff
@@ -1,40 +1,78 @@
...
+RUN yarn install --frozen-lockfile
+
+# Copy application code
+COPY . .
+
+# Precompile bootsnap code for faster boot times
+RUN bundle exec bootsnap precompile app/ lib/
+
+# Precompiling assets for production without requiring secret RAILS_MASTER_KEY
+RUN SECRET_KEY_BASE=1 ./bin/rails assets:precompile
```
luisgmetzger, line 51: We may want to explicitly state production environment here. Maybe something like this: 

RUN SECRET_KEY_BASE=1 RAILS_ENV=production bundle exec rails assets:precompile && \
    bundle exec bootsnap precompile --gemfile

Dockerfile
stevenbarragan: Are you suggesting to install this packages together in here https://github.com/GSA/search-gov/pull/1528/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R29 ?

If so, the idea is that the second packages are only installed in the `build` stage and not in the `base`. That's why they have to be installed separately.

Dockerfile
stevenbarragan: The `RAILS_ENV` variable is already se in [line 13](https://github.com/GSA/search-gov/pull/1528/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R13). No need to duplicated it in here.

Dockerfile
stevenbarragan: I'd like to keep them separate since they belong to separate things (Rails and Node). Ideally we should't have to have Node in the `base` and I'd like to keep all Node related setup together so its easier to identify in the future when we can move all environment variables and node related installation to the `build` stage.

What do you think?

Dockerfile
stevenbarragan: I had to split up node installation and installing the node packages since we need Node in the `base` image. I've added the line for removing the `/tmp/* and `/usr/local/share/.cache` directories in here https://github.com/GSA/search-gov/pull/1528/commits/bfba5428e3583aafe6298c26ab229737a5231a99

Dockerfile
luisgmetzger: @stevenbarragan  - Sure, that works ... I recommended keeping them together because of readability, setup, and also because that is how I have configured in the past. No further action needed here.

Dockerfile
luisgmetzger: @stevenbarragan  - No, what I am saying is that after the build stage, we add the following two lines to the build dependencies: 

```
  && apt-get clean \
    && rm -rf /var/lib/apt/lists/*
```

This will clean and remove the apt cache within the same RUN command

stevenbarragan: I wasn't able to add the second part `&& rm -rf /var/lib/apt/lists/*` because this line https://github.com/GSA/search-gov/pull/1528/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R32 breaks but `apt-get clean` worked just fine

luisgmetzger (approved):

PR #1540, opened 2024-04-17
SRCH-4982 Super Admin controls for Search Filters feature

ddzz:
## Summary
- SRCH-4982 Super Admin controls for Search Filters feature
- https://cm-jira.usa.gov/browse/SRCH-4982
 
### Checklist
 
#### Functionality Checks
 
#### Process Checks


app/models/affiliate.rb
```diff
@@ -726,4 +730,10 @@ def validate_managed_no_results_pages_guidance_text
 
     errors.add(:base, 'Additional guidance text is required when links are present.')
   end
+
+  def set_show_search_filter_settings
```
ddzz, line 734: This method ensures that search filter settings are disabled when the search engine is set to Bing.

yogesh27 (commented):
LGTM.

@stevenbarragan can you co-review this as well?

app/models/affiliate.rb
stevenbarragan: If we want to avoid accidentally setting it `true` on affiliates that are using bing maybe we can add a validation that ensures show_search_filter_settings can't be true if they're using bing

stevenbarragan (changes requested):
I have a few suggestions. Let me know what you think.

app/models/affiliate.rb
```diff
@@ -491,6 +491,10 @@ def no_results_error
     }.compact_blank
   end
 
+  def show_search_filter_settings_authorized?
```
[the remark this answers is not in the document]
ddzz: This is a specific Active Scaffold method that needs to be in the model according to Active Scaffold documentation. It determines whether the new column should appear in Super Admin and not the Admin Center, which uses the view you mentioned.

app/models/affiliate.rb
ddzz: This method is meant to ensure that the new parameter is always set to `false` when switching from SearchGov to Bing multiple times.

ddzz:
@stevenbarragan can you re-review when you have a chance? I responded to your comments. Thanks.

app/models/affiliate.rb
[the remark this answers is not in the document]
ddzz: Here's the Active Scaffold documentation for this method: https://github.com/activescaffold/active_scaffold/wiki/Security#model-methods-restricting-anything-else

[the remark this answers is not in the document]
stevenbarragan: Got it. Thank you for the explanation.

app/models/affiliate.rb
stevenbarragan: I didn't fully explained my self. This can be better accomplished through a validation as it provides user feedback making clients aware of show_search_filter_settings needing to be turned off if there is a switch. What do you think?

Something like:

```ruby
  validates : show_search_filter_settings, absence: { message: "should be false when search_engine is bing7" }, if: :bing7_search_engine?

  def bing7_search_engine?
    search_engine == 'bing7'
  end
```

stevenbarragan (changes requested):
One more comment left. The validation should be the last thing.

app/models/affiliate.rb
ddzz: Thanks, I like this approach because it’s more efficient than mine. I’ll update the PR shortly.

stevenbarragan (approved):
LGTM

PR #1549, opened 2024-04-25
SRCH-5129 Use rails cache for http external calls

stevenbarragan:
## Summary
- Use standard rails cache and Remove file-based api caching 
 
### Checklist
 
#### Functionality Checks
 
* This [change](https://github.com/GSA/search-gov/pull/1549/commits/91f03710baf2266e395ec19ac4c949aaafafb1da) is fixing a rubocop test for ruby +3.1 but makes the app crash on ruby 3.0.6.
* Failing test will be irrelevant after merging this changes https://github.com/GSA/search-gov/pull/1561
 
#### Process Checks


spec/engines/cached_search_api_connection_spec.rb
```diff
@@ -35,29 +30,20 @@
...
-      before do
-        allow(cache).to receive(:read).with(endpoint, params).and_return(nil)
-      end
-
-      it 'sends outbound request and cache response' do
-        allow(cached_connection.connection).to receive(:get).with(endpoint, params).and_return(response)
-        allow(cache).to receive(:write).with(endpoint, params, response)
+    context 'when response has been cached' do
+      it 'calls the api' do
+        allow(Rails.cache).to receive(:fetch).and_return(response)
```
lorawoodford, line 44: Can we move this `allow` into a before block to clearly illustrate the context that it is setting and how this context is different from that on ln 34.

spec/engines/cached_search_api_connection_spec.rb
```diff
@@ -35,29 +30,20 @@
...
 
-  describe '#get', vcr: { record: :skip } do
-    context 'on cache hit' do
-      before do
-        expect(cache).to receive(:read).with(endpoint, params).and_return(response)
-      end
+  describe '#get' do
+    context 'when response has not been cached' do
+      it 'calls the api' do
+        expect(cached_connection.connection).to receive(:get).with(endpoint, params)
```
lorawoodford, line 36: Nit-picky since it seems codeclimate isn't complaining about it, but I've generally seen us prefer `have_received` and setting a spy in a before block, over this style of placing `to receive` before the action in the body of the test.  Not a blocking issue at all.

lorawoodford (changes requested):
@stevenbarragan One minor bit of feedback regarding the tests, but this looks close to me.  We'll need to also get the flaky cucumber test to pass under 3.1.4 so that we can get all the required tests to pass (3.1.4 cucumber and 3.1.4 report coverage).  I'm rerunning tests for you to at least get all required tests green.  I understand about the failing 3.0.6 tests.  It's not blocking since they're no longer required.  (That said, @linuxcomputergeek it would be great to get that tiny ci fix in sooner rather than later so we can avoid these issues.)

What is the deployment plan for this?  I know you mentioned having to think it through in standup yesterday.  If this were merged prematurely it would break existing caching functionality in prod, yes?  What does the cutover plan look like?

I know at the 11th hour we probably don't care much/at all about cleaning up old cookbooks code, but I do know there are a number of recipes over there related to these existing api_cache processes - crons for cleaning up files, etc.  We'll want to make sure that @sb-ebukaanene isn't porting over any of that unnecessarily.

spec/engines/cached_search_api_connection_spec.rb
stevenbarragan: I keep getting other errors. I'll leave it as is for now
```
     Failure/Error: expect(connection).to have_receive(:get).with(endpoint, params)

     NoMethodError:
       undefined method `with' for #<RSpec::Matchers::BuiltIn::Has:0x000000011b0a43a0 @method_name=:have_receive, @args=[:get], @block=nil>

               expect(connection).to have_receive(:get).with(endpoint, params)
                                                       ^^^^^
```

stevenbarragan:
@lorawoodford we're the following regular deployment since now we have a cache server in legacy infrastructure. I've added a comment on the ticket tagging @sb-ebukaanene about the removal of the cache cleaning task.

lorawoodford (approved):
Ok, thanks @stevenbarragan.  Confirming that we will be acceptance testing this on current/existing infrastructure, right?  Otherwise, LGTM.

stevenbarragan:
@lorawoodford yes! Cache servers are in place in legacy infrastructure and this is been reviewed in staging.

PR #1568, opened 2024-05-10
SRCH-5042 Ensure paused domains stay paused

lorawoodford:
## Summary
- #1477 delivered the ability to stop indexing a domain, but since these paused domains were included in SearchgovDomain's ok scope that means that the `SitemapMonitorJob` that runs on cron every 4 hours resumed indexing these domains
- This also addresses some silent test failures that were being hidden by missing test steps: https://app.circleci.com/pipelines/github/GSA/search-gov/2047/workflows/7746b503-fc91-4ad6-bd83-d0a4fef19cef/jobs/17943?invite=true#step-113-1522_75
- Also ensures that when a domain is set to resume indexing via "reindex" that it's status gets updated (this seems to have been accidentally dropped here: https://github.com/GSA/search-gov/pull/1473/files#diff-3c3f47243cefd3130bddaed07d13516b796b15c32e1b5f31ccced09fcc31881cL56)
 
### Checklist
 
#### Functionality Checks
 
#### Process Checks


app/models/searchgov_domain.rb
```diff
@@ -26,7 +27,7 @@ class DomainError < StandardError; end
   attr_readonly :domain
   attr_reader :response
 
-  scope :ok, -> { where(status: GOOD_STATUS) }
+  scope :ok, -> { where(status: INDEXABLE_STATUS) }
```
stevenbarragan, line 30: It took me some time to figure this out. I think `ok` and `good` doesn't say much. What do you think about renaming this scope to `indexable` and the next one `no_indexable` instead?

stevenbarragan (approved):
Not a blocker but I leave a comment about improving the naming of the scopes in SearchgovDomain.

stevenbarragan (approved):
🙏🏽

PR #1608, opened 2024-06-27
SRCH 5324 Allow for uploading a CSV to mass update styles

krbhavith:
## Summary
- Upload affiliate styles through CSV from super admin 
 
### Checklist
 
#### Functionality Checks
 
- [⚠️] Automated checks pass. If Code Climate checks do not pass, explain reason for failures: [These failure can be ignored for now. 1. **branch condition size too high** which cannot be resolved for visual design settings, 2. **Similar blocks of code** is in unrelated files](https://codeclimate.com/github/GSA/search-gov/pull/1608)
 
#### Process Checks


app/services/bulk_affiliate_styles_uploader.rb
```diff
@@ -0,0 +1,205 @@
...
+  MAXIMUM_FILE_SIZE = 4.megabytes
+  VALID_CONTENT_TYPES = %w[text/csv].freeze
+  URL_ALREADY_TAKEN_MESSAGE = 'Validation failed: Url has already been taken'
+
+  attr_reader :results
+
+  class Error < StandardError
+  end
+
+  class Results
```
luismetzger, line 13: Would we be able to extract this rather than have a nested class that is this big?

app/jobs/.DS_Store
luisgmetzger, line 1: We may want to uncheck these type of files so they don’t end up on the repo.

app/jobs/bulk_affiliate_styles_uploader_job.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  def perform(user, filename, filepath)
+    @user = user
+    @filename = filename
+    @uploader = BulkAffiliateStylesUploader.new(filename, filepath)
+    @uploader.upload
+    report_results
+  end
+
+  def report_results
```
luisgmetzger, line 14: If the parent method and/or children methods of this class  doesn’t have to be publicly accessed by the rest of the app, let’s make them private.

app/services/.DS_Store
luisgmetzger, line 1: Same as comment earlier

luisgmetzger (approved):

PR #1612, opened 2024-07-09
SRCH-5270 updated aria labels for pagination to come from locales

yogesh27:
## Summary
- updated aria labels for pagination to come from locales
 
### Checklist
 
#### Functionality Checks
 
#### Process Checks

### DIFF
--- a/app/javascript/components/UswdsOverrides/UswdsPagination.tsx
+++ b/app/javascript/components/UswdsOverrides/UswdsPagination.tsx
@@ -25,10 +25,12 @@ type PaginationProps = {
 const PaginationPage = ({
   page,
   isCurrent,
+  label,
   onClickPageNumber
 }: {
   page: number
   isCurrent?: boolean
+  label: string
   onClickPageNumber?: (
     event: React.MouseEvent<HTMLButtonElement>,
     page: number
@@ -48,7 +50,7 @@ const PaginationPage = ({
           unstyled
           data-testid="pagination-page-number"
           className={linkClasses}
-          aria-label={`Page ${page}`}
+          aria-label={label}
           aria-current={isCurrent ? 'page' : undefined}
           onClick={(event) => {
             onClickPageNumber(event, page);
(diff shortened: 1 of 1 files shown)


app/javascript/components/UswdsOverrides/UswdsPagination.tsx
```diff
@@ -48,7 +50,7 @@ const PaginationPage = ({
           unstyled
           data-testid="pagination-page-number"
           className={linkClasses}
-          aria-label={`Page ${page}`}
+          aria-label={label}
```
stevenbarragan, line 53: What do you think about pulling the label directly in here? I know it'd be duplicated but It will reduce the dependencies of the `PaginationPage` component.
```tsx
aria-label={`${i18n.t("page")} ${pageNum}`}
```

stevenbarragan: It could be also calculated in here like
```
const label = `${i18n.t("page")} ${pageNum}`
...
aria-label={label}
```

stevenbarragan (changes requested):
I have a suggestion but overall it looks good.

app/javascript/components/UswdsOverrides/UswdsPagination.tsx
yogesh27: That was my initial implementation but we will have to use `i18n` in `PaginationPage` 
`const i18n = useContext(LanguageContext); ` component also. 

It is already being used at `UswdsPagination` component so it will be duplicated.

stevenbarragan: That's fine since contexts are mean to be shared this way. Context's are already shared among all components.

yogesh27: sure, let me update.

stevenbarragan (approved):

PR #1625, opened 2024-07-31
SRCH-5312 Improve error logging for search-gov

krbhavith:
## Summary
- Send whole exception instead of just the error message in the logger.
 
### Checklist
 
#### Functionality Checks
 
#### Process Checks

### DIFF
--- a/app/jobs/affiliate_indexed_document_fetcher_job.rb
+++ b/app/jobs/affiliate_indexed_document_fetcher_job.rb
@@ -9,9 +9,9 @@ class AffiliateIndexedDocumentFetcherJob < ApplicationJob
       where(conditions).order(:last_crawled_at).each do |indexed_document|
       IndexedDocument.find(indexed_document.id).fetch
     rescue ActiveRecord::RecordNotFound => e
-      Rails.logger.warn "Cannot find IndexedDocument to fetch: #{e}"
+      Rails.logger.warn "Cannot find IndexedDocument to fetch:", e
     end
   rescue ActiveRecord::RecordNotFound => e
-    Rails.logger.warn "Ignoring race condition in AffiliateIndexedDocumentFetcherJob: #{e}"
+    Rails.logger.warn "Ignoring race condition in AffiliateIndexedDocumentFetcherJob:", e
   end
 end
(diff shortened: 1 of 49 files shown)


stevenbarragan (changes requested):
We need to send the exception directly to the logger not just in a string.

app/jobs/affiliate_indexed_document_fetcher_job.rb
```diff
@@ -9,9 +9,9 @@ def perform(affiliate_id, start_id, end_id, scope)
       where(conditions).order(:last_crawled_at).each do |indexed_document|
       IndexedDocument.find(indexed_document.id).fetch
     rescue ActiveRecord::RecordNotFound => e
-      Rails.logger.warn "Cannot find IndexedDocument to fetch: #{e}"
+      Rails.logger.warn 'Cannot find IndexedDocument to fetch:', e
     end
   rescue ActiveRecord::RecordNotFound => e
-    Rails.logger.warn "Ignoring race condition in AffiliateIndexedDocumentFetcherJob: #{e}"
+    Rails.logger.warn 'Ignoring race condition in AffiliateIndexedDocumentFetcherJob:', e
```
stevenbarragan, line 15: Can we change this to log the exception as `error` instead?
```
 Rails.logger.error 'Ignoring race condition in AffiliateIndexedDocumentFetcherJob:', e
 ```

File: lib/geoip_lookup.rb [hunk 2]
```diff
@@ -6,7 +6,7 @@ def self.establish_connection!
   def self.lookup(ip)
     @geoip_db.city(ip)
   rescue SocketError => e
-    Rails.logger.warn("GeoipLookup failed lookup on #{ip}: #{e.message}")
+    Rails.logger.warn "GeoipLookup failed lookup on #{ip}", e
```
stevenbarragan, line 9: Can we use `Rails.logger.warn` here instead?

stevenbarragan (changes requested):
One last change.

lib/geoip_lookup.rb
krbhavith: You mean `Rails.logger.error`?

stevenbarragan (approved):
LGTM

PR #1653, opened 2024-09-19
SRCH-5524: LoadError: cannot load such file -- i18n/tasks/scanners/file_scanner

luisgmetzger:
## Summary
- This PR makes a change to the JSX file scanner file. This class depends on this development and test gem: `i18n-tasks`. The class does not need to be loaded in `production` environment.
 
### Checklist
 
#### Functionality Checks
 
#### Process Checks


config/application.rb
```diff
@@ -29,6 +28,7 @@ class Application < Rails::Application
     config.autoload_paths += Dir[config.root.join('lib', 'middlewares').to_s]
     config.autoload_paths += Dir[config.root.join('lib', 'parsers').to_s]
     config.autoload_paths += Dir[config.root.join('lib', 'renderers').to_s]
+    config.autoload_paths += Dir[config.root.join('lib', 'i18n_scanners').to_s]
```
stevenbarragan, line 31: Should we move this line to `config/environment/development.rb` and `config/environment/testing.rb` instead so it only loads in production?

stevenbarragan (changes requested):

config/environments/development.rb
```diff
@@ -1,4 +1,7 @@
 require "active_support/core_ext/integer/time"
+if Rails.env.development?
```
stevenbarragan, line 2: This file only gets loaded in development. No need to put a conditional around it.

config/environments/test.rb
```diff
@@ -1,4 +1,7 @@
 require "active_support/core_ext/integer/time"
+if Rails.env.test?
```
stevenbarragan, line 2: Same here. Conditional is no needed since file is loaded only in the test environment.

spec/spec_helper.rb
```diff
@@ -11,6 +11,10 @@
 require 'active_storage_validations/matchers'
 require 'webmock/rspec'
 
+if Rails.env.test?
```
stevenbarragan, line 14: No need to check this inside the `spec_helper`. This file only gets loaded in testing already.

stevenbarragan (changes requested):

stevenbarragan (approved):

PR #1737, opened 2024-11-18
SRCH-4984 Allow regular search to return faceted results

stevenbarragan:
## Summary
- Move faceted search functionality from i14y_search_api to  i14y_search so regular search can return faceted results as well.
 
### Checklist
 
#### Functionality Checks
 
#### Process Checks


luisgmetzger:
@stevenbarragan - Is there any tests that need to added or modified for this?

stevenbarragan:
@luisgmetzger I added a test to make sure facet fields are making it to the react app.

app/models/api_search.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+module ApiSearch
+  def initialize(options = {})
+    super
+
+    @highlight_options[:pre_tags]  ||= Api::V2::HighlightOptions::DEFAULT[:pre_tags]
+    @highlight_options[:post_tags] ||= Api::V2::HighlightOptions::DEFAULT[:post_tags]
```
luisgmetzger, line 8: Not sure if we will be using it but could it be beneficial to set `attrs` here for the instantiated variables.

config/application.rb
```diff
@@ -68,10 +68,6 @@ class Application < Rails::Application
     # Rails 5.0 had false.
     config.active_record.belongs_to_required_by_default = false
 
-    # Temporary workaround for:
-    # https://discuss.rubyonrails.org/t/cve-2022-32224-possible-rce-escalation-bug-with-serialized-columns-in-active-record/81017
```
luisgmetzger, line 72: This doesn't look to be part of the ticket. Are we just good to remove from our earlier discussion this morning?

app/models/api_search.rb
stevenbarragan: I not sure why we would need to set attrs. It everything works without them I rather keep it like it is now.

config/application.rb
stevenbarragan: Yes. Removing this config since it's not doing anything anymore.

stevenbarragan: Actually it still exists in rails. I'm going to put it back for now.

luisgmetzger (approved):

PR #1743, opened 2024-11-26
SRCH-5154 Bulk delete zombie records from Elastic Search

krbhavith:
## Summary
- Upload a list of url's(zombie) through Super Admin page to delete from `SearchgovUrl` table and from Kibana.
- Added concern `BulkUploadHandler` to be used by all bulk upload controllers.
 
### Checklist
 
#### Functionality Checks
 
#### Process Checks

### DIFF
--- a/app/controllers/admin/bulk_zombie_url_upload_controller.rb
+++ b/app/controllers/admin/bulk_zombie_url_upload_controller.rb
@@ -1,41 +1,34 @@
 # frozen_string_literal: true
 
-module Admin
-  class BulkZombieUrlUploadController < AdminController
-    include BulkUploadHandler
-    before_action :set_page_title
+class Admin::BulkZombieUrlUploadController < Admin::BulkUploadController
+  def upload
+    handle_bulk_upload(
+      params_key: :bulk_upload_zombie_urls,
+      validator_class: BulkZombieUrls::FileValidator,
+      error_class: BulkZombieUrlUploader::Error,
+      success_path: admin_bulk_zombie_url_upload_index_path,
+      logger_message: 'Zombie Url upload failed'
+    )
+  end
 
-    def index; end
+  private
 
-    def upload
-      handle_bulk_upload(
-        params_key: :bulk_upload_zombie_urls,
-        validator_class: BulkZombieUrls::FileValidator,
-        error_class: BulkZombieUrlUploader::Error,
-        success_path: admin_bulk_zombie_url_upload_index_path,
-        logger_message: 'Zombie Url upload failed'
-      )
-    end
+  def set_page_title
+    @page_title = 'Bulk Zombie Url Upload'
+  end
 
-    private
+  def success_message(filename)
+    <<~SUCCESS_MESSAGE
+      Successfully uploaded #{filename} for processing.
+      The results will be emailed to you.
+    SUCCESS_MESSAGE
+  end
 
-    def set_page_title
-      @page_title = 'Bulk Affiliate Styles Upload'
-    end
-
-    def success_message(filename)
-      <<~SUCCESS_MESSAGE
-        Successfully uploaded #{filename} for processing.
-        The results will be emailed to you.
-      SUCCESS_MESSAGE
-    end
-
-    def enqueue_job
-      BulkZombieUrlUploaderJob.perform_later(
-        current_user,
-        @file.original_filename,
-        @file.tempfile.path
-      )
-    end
+  def enqueue_job
+    BulkZombieUrlUploaderJob.perform_later(
+      current_user,
+      @file.original_filename,
+      @file.tempfile.path
+    )
   end
 end
--- a/app/controllers/admin/bulk_zombie_url_upload_controller.rb
+++ b/app/controllers/admin/bulk_zombie_url_upload_controller.rb
@@ -3,17 +3,17 @@
 module Admin
   class BulkZombieUrlUploadController < AdminController
     def index
-      @page_title = 'Bulk Zombie URL Upload'
+      @page_title = 'Bulk Zombie Url Upload'
     end
 
     def upload
       begin
-        @file = params[:bulk_upload_urls]
-        BulkZombieUrlUploader::UrlFileValidator.new(@file).validate!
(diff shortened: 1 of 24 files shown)


stevenbarragan (changes requested):
Plase make sure rspec, cucumber, and codeclimate checks pass.

krbhavith:
> Plase make sure rspec, cucumber, and codeclimate checks pass.

I have to make one more change to delete the record with `document_id` in I14yDocument when url is not available in searchgov_urls table.

stevenbarragan (changes requested):
@krbhavith don't forget to fix the test coverage

app/controllers/concerns/bulk_upload_handler.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+module BulkUploadHandler
+  def handle_bulk_upload(options)
```
stevenbarragan, line 4: Is there a reason for using a hash instead regular parameters? Can we just use regular parameters for this. e.g.
```
def handle_bulk_upload(:file, :error_class, :validator_class)
  ...
end
```

app/controllers/concerns/bulk_upload_handler.rb
```diff
@@ -0,0 +1,29 @@
...
+  def validate_file(file, validator_class)
+    validator_class.new(file).validate!
+  end
+
+  def perform_upload(file)
+    enqueue_job
+    flash[:success] = success_message(file.original_filename)
+  end
+
+  def handle_upload_error(error, options)
```
stevenbarragan, line 25: Same comment as before. Please prefer using regular parameters to pass data between functions.

app/controllers/admin/bulk_zombie_url_upload_controller.rb
```diff
@@ -0,0 +1,34 @@
...
+      Successfully uploaded #{filename} for processing.
+      The results will be emailed to you.
+    SUCCESS_MESSAGE
+  end
+
+  def enqueue_job
+    BulkZombieUrlUploaderJob.perform_later(
+      current_user,
+      @file.original_filename,
+      @file.tempfile.path
```
stevenbarragan, line 31: This approach won't work on staging or production. When the request comes in the tempfile gets created in the `app` servers but the workers will try to perform the upload from the `crawl` server. 

This file needs to be send to S3 and the job needs to read it from there.

app/jobs/bulk_zombie_url_uploader_job.rb
```diff
@@ -0,0 +1,33 @@
...
+  private
+
+  def report_results
+    log_results
+    send_results_email
+  end
+
+  def log_results
+    results = @uploader.results
+    Rails.logger.info "BulkZombieUrlUploaderJob: #{results.file_name}"
```
stevenbarragan, line 23: This is not how we send extra data with our logger.  This extra spaces will just look weird in the logs. Just send extra data as a hash. Our logger will formate it accordingly.
```
 Rails.logger.info(BulkZombieUrlUploaderJob: results.file_name, total_urls: results.total_count, errors:  results.error_count)
```

app/services/bulk_zombie_url_uploader.rb
```diff
@@ -0,0 +1,110 @@
...
+  private
+
+  def initialize_results
+    @results = BulkZombieUrls::Results.new(@file_name)
+    raise Error, 'Results object not initialized' unless @results
+  end
+
+  def process_upload
+    parse_csv.each { |row| process_row(row) }
+  rescue CSV::MalformedCSVError => e
```
stevenbarragan, line 32: We need to stop using `rescue` everywhere.

app/services/bulk_zombie_url_uploader.rb
```diff
@@ -0,0 +1,110 @@
...
+    handle_csv_error(e)
+  end
+
+  def parse_csv
+    csv = CSV.parse(File.read(@file_path), headers: true)
+    raise CSV::MalformedCSVError, "Missing required headers" unless %w[URL DOC_ID].all? { |col| csv.headers.include?(col) }
+    csv
+  rescue CSV::MalformedCSVError, ArgumentError => e
```
stevenbarragan, line 40: On line 38 we're raising an CSV::MalformedCSVError but rescuing the same error in the same method and creating a new CSV::MalformedCSVError. Please improve the workflow and rescue only when it's complete necessary and there is not other way.

stevenbarragan (changes requested):
Please make sure file exists where the job will read it. And avoid using rescue as much.

app/controllers/concerns/bulk_upload_handler.rb
krbhavith: Using a hash would be cleaner and maintainable when calling `handle_bulk_upload` from different controllers,  and makes it easier to extend the method without breaking existing calls.

app/controllers/admin/bulk_zombie_url_upload_controller.rb
krbhavith: Existing similar functionality at https://github.com/GSA/search-gov/blob/main/app/controllers/admin/bulk_url_upload_controller.rb#L33 working fine now in production

app/services/bulk_zombie_urls/file_validator.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+class BulkZombieUrls::FileValidator
+  MAXIMUM_FILE_SIZE = 4.megabytes
```
stevenbarragan, line 4: Where this this restriction comes from? I don't see it as part of the ticket.

stevenbarragan (changes requested):
We should not need to use`.send` to call methods plase change that. There is a few places where logging can be improved.

app/services/bulk_zombie_urls/file_validator.rb
krbhavith: It's not the requirement from the ticket, but I am following this as per the previous other upload processes.

selfdanielj:
Feature is no longer necessary

PR #1781, opened 2025-03-11
Staging

stevenbarragan:
## Summary
- SRCH-5549 Remove hardcoded ruby version from deploy.rb.
- SRCH-5849 Implement a new search engine SearchElastic.
- SRCH-5876 Parameterize shards and replicas of the SearchElastic index.


jimkreft (approved):
Approved, let's ship this!

PR #1800, opened 2025-04-11
SRCH-4985: Customer controls for Search Filters component

luisgmetzger:
## Summary
- This PR adds search filter controls to display overview.
- This PR adds filter subclasses for STI pattern.
- This PR modifies the existing `initialize_default_filters` method to use subclass mapping.
- This PR adds additional test coverage on changes.
 
### Checklist
 
#### Functionality Checks
 
#### Process Checks


Gemfile
```diff
@@ -16,6 +16,7 @@ gem 'nokogiri', '~> 1.16'
 gem 'authlogic', '~> 6.4', '>= 6.4.3'
 gem 'omniauth_login_dot_gov', git: 'https://github.com/18f/omniauth_login_dot_gov', ref: '6e117a9c68b19a1fbc70533613b74b0d8affd641'
 gem 'rack', '~> 2.2'
+gem 'acts_as_list', '~> 1.2', '>= 1.2.4'
```
stevenbarragan, line 19: What is this gem needed for?

app/models/file_type_filter.rb
```diff
@@ -0,0 +1,3 @@
+class FileTypeFilter < Filter
+  validates :label, presence: true, if: :enabled?
```
stevenbarragan, line 2: Can you move this validation to the `Filter`?

app/views/sites/displays/edit.html.haml
```diff
@@ -12,8 +12,8 @@
 
   = render partial: 'sidebar', locals: { f: f }
   = render partial: 'govboxes', locals: { f: f }
+  - if @site.search_engine == 'SearchGov' && @site.show_search_filter_settings
```
stevenbarragan, line 15: If the affiliate is using the new `SearchElastic` engine we should display this settings too.

spec/models/custom_filter_spec.rb
```diff
@@ -14,10 +14,5 @@
       expect(custom_filter).not_to be_valid
       expect(custom_filter.errors[:label]).to include("can't be blank")
     end
-
-    it 'is valid without a label when not enabled' do
```
stevenbarragan, line 18: Did this changed? CustomFilter can't be valid without a label?

spec/models/file_type_filter_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+require 'spec_helper'
+
+describe FileTypeFilter, type: :model do
+  describe 'Validations' do
+    it 'is valid with a label when enabled' do
+      file_type_filter = described_class.new(label: 'Some label', enabled: true)
+      expect(file_type_filter).to be_valid
+    end
+
+    it 'is invalid without a label when enabled' do
```
stevenbarragan, line 12: I though the ticket said that when enabled, if the label is not set it should use the default value, so the filter can be enabled without a label, but I'll show up the default value in the frontend.

The issue I see now is not having default values for filters. Label should not be need to be set in order to use the filter AFIK.

stevenbarragan (changes requested):

Gemfile
luisgmetzger: @stevenbarragan  - Originally, I added this to handle positioning as the gem acts as a positioning helper for ActiveRecord. This implementation was before I knew what the front-end would look like in a previous ticket. I will remove.

app/models/file_type_filter.rb
luisgmetzger: @stevenbarragan  - Sure, I can remove from the subclasses and have this only in the parent class.

app/views/sites/displays/edit.html.haml
luisgmetzger: @stevenbarragan - I will change the condition to check for `SearchElastic`.

spec/models/custom_filter_spec.rb
luisgmetzger: @stevenbarragan  - I have added a new set of tests to replace this.

stevenbarragan (approved):
:shipit:

PR #1827, opened 2025-05-21
SRCH-6000 SRCH-5978 SRCH-6015  Release

stevenbarragan:
## Summary
- SRCH-6000 Stop displaying result's images
- SRCH-5978 Bulk set to active = false a csv of affiliates
- SRCH-6015 Add index for links query


jimkreft (approved):
approved!

PR #1832, opened 2025-06-02
SRCH-5993 SRCH-6018 Release

stevenbarragan:
## Summary
- SRCH-5993 Add bulk operation for search engine change for affiliates.
- SRCH-6018 Create SearchModule for search elastic and report to it when using Search Elastic engine.


jimkreft (approved):
Approved

PR #1833, opened 2025-06-02
SRCH-6022 Set SearchElastic as default search engine

stevenbarragan:
## Summary
- Set SearchElastic as default search engine
- Update test to account for this change
 
### Checklist
 
#### Functionality Checks
 
#### Process Checks

### DIFF
--- a/db/migrate/20250602213821_change_search_engine_from_affiliates.rb
+++ b/db/migrate/20250602213821_change_search_engine_from_affiliates.rb
@@ -1,0 +1,5 @@
+class ChangeSearchEngineFromAffiliates < ActiveRecord::Migration[7.1]
+  def change
+    change_column :affiliates, :search_engine, :string, default: 'SearchElastic'
+  end
+end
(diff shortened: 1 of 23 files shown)


db/migrate/20250602213821_change_search_engine_from_affiliates.rb
```diff
@@ -0,0 +1,5 @@
+class ChangeSearchEngineFromAffiliates < ActiveRecord::Migration[7.1]
+  def change
+    change_column :affiliates, :search_engine, :string, default: 'SearchElastic'
```
luisgmetzger, line 3: Could you confirm that `SearchElastic` naming convention is not snake care like this: `search_elastic` ... I ask because I ran into a similar naming issue with bulk operation where it gave me a validation error due to not having it snake case. 

I am not sure if it will matter directly in the table as the validation might be at the model level but wanted to ask either way.

stevenbarragan: Yes, this is the value on the database.

luisgmetzger (approved):

PR #1874, opened 2025-09-30
SRCH-6203 Create DB model and migration

stevenbarragan:
## Summary
- Created CrawlConfig model with validations, migration and tests.
- Using fugit gem to validate CRON format.
- Store array's of strings as JSON

# NOTES
- Run migrations as
```
bin/rails db:migrate
```

### DIFF
--- a/app/models/crawl_config.rb
+++ b/app/models/crawl_config.rb
@@ -1,0 +1,64 @@
+class CrawlConfig < ApplicationRecord
+  serialize :allowed_domains, coder: JSON
+  serialize :starting_urls, coder: JSON
+  serialize :sitemap_urls, coder: JSON
+  serialize :deny_paths, coder: JSON
+
+  before_validation :normalize_domains_and_urls
+  before_validation :initialize_empty_arrays
+
+  validates :name, presence: true
+  validates :allowed_domains, presence: true
+  validates :starting_urls, presence: true
+  validates :depth_limit, presence: true, numericality: { only_integer: true }
+  validates :schedule, presence: true
+  validates :output_target, presence: true, inclusion: { in: %w[endpoint elasticsearch] }
+
+  validate :validate_schedule_format
+  validate :validate_starting_urls
+  validates :allowed_domains, uniqueness: { scope: :output_target, message: "and output target combination must be unique" }
+
+  private
+
+  def normalize_domains_and_urls
+    # Sort and uniq arrays to ensure consistent representation for uniqueness validation
+    self.allowed_domains.sort!.uniq! if allowed_domains.is_a?(Array)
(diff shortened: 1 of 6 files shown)


app/models/crawl_config.rb
```diff
@@ -0,0 +1,67 @@
+# frozen_string_literal: true
+
+class CrawlConfig < ApplicationRecord
+  serialize :allowed_domains, coder: JSON
+  serialize :deny_paths, coder: JSON
+  serialize :sitemap_urls, coder: JSON
+  serialize :starting_urls, coder: JSON
```
luisgmetzger, line 7: If these columns are JSON/JSONB in DB, you can probably drop `serialize`.

app/models/crawl_config.rb
```diff
@@ -0,0 +1,67 @@
...
+  def normalize_domains_and_urls
+    # Sort and uniq arrays to ensure consistent representation for uniqueness validation
+    self.allowed_domains.sort!.uniq! if allowed_domains.is_a?(Array)
+    self.starting_urls.sort!.uniq! if starting_urls.is_a?(Array)
+  end
+
+  def initialize_empty_arrays
+    self.sitemap_urls ||= []
+    self.deny_paths ||= []
+  end
```
luisgmetzger, line 36: Maybe this can be more like this: 

```
  ARRAY_ATTRS = %i[
    allowed_domains
    deny_paths
    sitemap_urls
    starting_urls
  ].freeze

  # Ensure arrays are present, sorted, and unique for a canonical representation
  def prepare_arrays
    ARRAY_ATTRS.each do |attr|
      value = send(attr)
      value = [] unless value.is_a?(Array)
      value = value.compact.map(&:to_s) # Strings are a safe canonical form
      value.sort!
      value.uniq!
      send("#{attr}=", value)
    end
  end
  ```

app/models/crawl_config.rb
stevenbarragan: I've tried to set this columns as json type from the database, but I still have to specify a serializer in here. Also, when I changed the type I cannot add the index on output_target and allowed_domains that I'm using to validate uniqueness between those two. So, I decided to leave it as type text, supporting the unique index, serializer setup is still needed regardless.

app/models/crawl_config.rb
stevenbarragan: Why would this be better?

In the two methods above is very clear what's going on and why. It's very hard to follow what's happening on your suggested code, and why would it be better. Also we should not be using `send` as a "regular "programming pattern.

app/models/crawl_config.rb
luisgmetzger: That is fine we can keep as is if it working.

luisgmetzger (approved):

PR #1875, opened 2025-10-01
SRCH-5952: Added multi-field keyword option

igoristic:
## Summary
This PR adds a multi-field type for "domain_name" field. We will now be able to do partial-matching/aggregation by doing something like:
```json
GET /our_index_name/_search
{
  "size": 0,
  "aggs": {
    "by_domain": {
      "terms": { "field": "domain_name.keyword" }
    }
  }
}
```

We will also need to apply it to mapping, eg:
```json
PUT our_index_name
{
  "mappings": {
    "properties": {
      "domain_name": {
        "type": "text",
        "analyzer": "domain_name_analyzer",
        "fields": {
          "keyword": {
            "type": "keyword"
          }
        }
      }
    }
  }
}
```

And possibly update the index template as well, eg:
```json
PUT _template/our_template_name
{
  "index_patterns": ["your_index_pattern"],
  "settings": {
    "number_of_shards": 1,
    "number_of_replicas": 1
  },
  "mappings": {
    "properties": {
      "domain_name": {
        "type": "text",
        "analyzer": "domain_name_analyzer",
        "fields": {
          "keyword": {
            "type": "keyword"
          }
        }
      }
    }
  }
}
```

 
### Checklist
 
#### Functionality Checks
 
#### Process Checks

### DIFF
--- a/lib/search_elastic/template.rb
+++ b/lib/search_elastic/template.rb
@@ -222,6 +222,11 @@ class SearchElastic::Template
     json.domain_name do
       json.type('text')
       json.analyzer('domain_name_analyzer')
+      json.fields do
+        json.keyword do
+          json.type('keyword')
+        end
+      end
     end
   end
 


stevenbarragan (dismissed):
LGTM

igoristic:
@stevenbarragan and @selfdanielj 

Sorry for the confusion. I had opensearch-ruby gems installed and had custom Opensearch client code (for debugging). I addressed the error after switching to the built-in OpenSearch client (that's using elasticsearch-ruby gem)

config/initializers/es.rb
```diff
@@ -21,19 +21,56 @@ def self.client
...
+    else
+      Rails.logger.info { "Creating new Elasticsearch index: #{index_name}" }
+      self.client.indices.put_template(
+        body: template_generator.body,
+        include_type_name: false,
+        name: index_name,
+        order: 0
+      )
+      repo = SearchElastic::DocumentRepository.new
+      repo.create_index!(index: index_name, include_type_name: false)
```
selfdanielj: When i run this i don't see the template or the index created in elastic... i also tried running create_indexes on main and i also don't see it... is usasearch:elasticsearch:create_indexes the right command?

app/models/open_search/indexer.rb
```diff
@@ -4,16 +4,55 @@ class OpenSearch::Indexer
...
+          type: 'text',
+          analyzer: 'domain_name_analyzer',
+          fields: {
+            keyword: {
+              type: 'keyword'
+            }
+          }
+        }
+      }
+    }
```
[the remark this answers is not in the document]
igoristic: I chose this approach to keep the code modular and maintainable. By breaking down the template into smaller parts, it becomes easier to understand and modify individual parts of the mapping. It also makes it clear what fields we're trying to update (if the index already exists).

If you want to make another change to the mapping, you can modify the `mapping_update` schema directly. For example, if you want to add a new field to the mapping, you can simply add it to the properties hash:

```ruby
mapping_update = {
  properties: {
    domain_name: {
      type: 'text',
      analyzer: 'domain_name_analyzer',
      fields: {
        keyword: {
          type: 'keyword'
        }
      }
    },
    new_field: {
      type: 'text'
    }
  }
}
```

ps: You would also need to add it the same change to `lib/search_elastic/template.rb`

selfdanielj (dismissed):
Verified functionality on both elastic and opensearch.  Nice work!

config/initializers/es.rb
```diff
@@ -20,20 +21,15 @@ def self.client
...
-    template_generator = SearchElastic::Template.new("*#{index_name}*")
-
-    self.client.indices.put_template(
-      body: template_generator.body,
-      create: true,
-      include_type_name: false,
-      name: index_name,
-      order: 0
-    )
+  @index_creator = SearchElasticIndexCreator.new(
```
stevenbarragan: What do you think about setting up the client directly when creating this object, rather when calling the `create_or_update_index` method?

lib/search_elastic/index_create.rb
```diff
@@ -0,0 +1,94 @@
...
+    @index_name = index_name
+    @shards = shards
+    @replicas = replicas
+  end
+
+  def create_or_update_index(client)
+    @client = client
+    index_pattern = "*#{@index_name}*"
+    template_generator = SearchElastic::Template.new(index_pattern)
+    template_body = JSON.parse(template_generator.body)
```
stevenbarragan: Can you add a comment why this is now necessary?, It seems that it's for setting the shards, and replicas but I don't see it being set before on the template, can we pass this parameters to the template directly?

PR #1879, opened 2025-10-06
SRCH-6204 spider admin UI

stevenbarragan:
## Summary
- Create CRUD interface for Crawl Config records
- Array elements are separated by new lines in texareas
- Fix `endpoint` typo
- Add a CrawlConfig seed record


selfdanielj (commented):
Can it be changed to show different fields by default?

Image

I'm not the main user but I think that seeing the starting urls and sitemap urls by default would be more helpful than the created date and probably the output target (since i think we will eventually deprecate "endpoint" and have them only write to elasticsearch/opensearch

selfdanielj (commented):
When I create a new crawl config record here http://localhost:3000/admin/crawl_configs It does not allow me to set values for: `sitemap_check_hours`, `allow_query_string`, or `handle_javascript`.

Nor does the page show those values or allow me to edit them on existing records.

selfdanielj (approved):
Looks good to me, i tested out the changes.  I can't really speak to the code quality.

igoristic (approved):
Awesome work 🏆

PR #1993, opened 2026-03-19
SRCH-6448: Port i14y Serde serialize_hash into search-gov

luisgmetzger:
## What this does

Makes `Serde.serialize_hash` functional in search-gov by adding the missing `String#extract_array` method. This is the first step toward having the crawler index directly into OpenSearch instead of going through the i14y HTTP API.

The Serde module (`lib/serde.rb`) was already copied from i14y and is identical, but the serialize path was never used because indexing always went through i14y. The only thing missing was `String#extract_array`, a 3-line method that splits comma-separated strings into downcased arrays. Serde calls it when processing `tags` and `searchgov_custom1/2/3` fields.

Also adds spec coverage for the full Serde module, which had zero specs in search-gov until now.

## Changes

- **`config/initializers/extensions/string.rb`** -- added `extract_array` (split on comma, strip, downcase). Same logic already exists as a private method in `SearchElasticEngine` and in i14y's `lib/ext/string.rb`.
- **`spec/lib/serde_spec.rb`** -- new spec file covering `serialize_hash`, `deserialize_hash`, and `uri_params_hash`. Ported from i14y's spec and expanded with additional edge cases (Spanish language fields, special characters, missing optional fields, pre-existing arrays, various URL formats).
- **`spec/models/string_spec.rb`** -- added `#extract_array` examples to the existing String spec.

No changes to `lib/serde.rb` itself -- the code was already correct.

## Testing

1. Run the Serde spec: `bundle exec rspec spec/lib/serde_spec.rb`
2. Run the String spec: `bundle exec rspec spec/models/string_spec.rb`
3. Verify no regressions in existing `deserialize_hash` consumers -- the only caller is `SearchElastic::DocumentSearchResults#deserialized`, which is unchanged.

## Jira

https://gsa-standard.atlassian-us-gov-mod.net/browse/SRCH-6448

### DIFF
--- a/config/initializers/extensions/string.rb
+++ b/config/initializers/extensions/string.rb
@@ -8,4 +8,7 @@ class String
     gsub(/(\b|')[a-z]+/) { |w| NON_CAPITALIZED.include?(w) ? w : w.capitalize }.sub(/^[a-z]/) { |l| l.upcase }
   end
 
+  def extract_array
+    split(',').map(&:strip).map(&:downcase)
+  end
 end
--- a/spec/lib/serde_spec.rb
+++ b/spec/lib/serde_spec.rb
@@ -1,0 +1,335 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe Serde do
+  describe '.serialize_hash' do
+    subject(:serialize_hash) do
+      described_class.serialize_hash(original_hash, 'en')
+    end
+
+    let(:original_hash) do
+      ActiveSupport::HashWithIndifferentAccess.new(
+        { 'title' => 'my title',
+          'description' => 'my description',
+          'content' => 'my content',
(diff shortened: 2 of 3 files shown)


selfdanielj (changes requested):

selfdanielj (changes requested):
This still contains Azhar's commit that we do not want in staging `4acf0bfd05c8ff4afb7f81455a81c7a712fa559f`... All the changes to the `cicd-scripts/`, `docs/` and `config/` `buildspec`, etc..

YaritzaGarcia:
I ran this command to revert the changes: git revert 4acf0bfd05c8ff4afb7f81455a81c7a712fa559f

YaritzaGarcia (approved):
LGTM

PR #1994, opened 2026-03-19
SRCH-6449: Build OpenSearch document indexing service

luisgmetzger:
## What this does

Adds a new `LegacyOpenSearch::DocumentIndexer` service class that can index and delete documents directly in OpenSearch, bypassing the i14y HTTP API entirely. This is the next piece of the Elasticsearch-to-OpenSearch migration (SRCH-6423).

The indexer accepts the same params shape that `SearchgovUrl` already builds for `I14yDocument`, runs the fields through `Serde.serialize_hash` for all the usual transformations (language-suffix renaming, HTML sanitization, array splitting, URI extraction), then writes straight to the OpenSearch cluster via `OPENSEARCH_CLIENT`.

Depends on #1993 (SRCH-6448) for `Serde` and `String#extract_array`.

## Key decisions

- Uses `client.index()` as an upsert instead of separate create/update calls -- this is simpler and matches what `SearchgovUrl` actually needs (re-crawls should overwrite)
- Wraps incoming params in `HashWithIndifferentAccess` to match how i14y's `DocumentRepository` feeds data to `Serde`
- Injects a `created_at` default when not provided, replicating i14y's `Document` model behavior
- Maps `Elasticsearch::Transport` errors to custom exception classes (`DuplicateID`, `DocumentIndexerError`) for clean caller handling
- `delete` accepts a `handle:` kwarg it ignores, for backward compatibility with existing call sites

## Files changed

- `app/models/legacy_opensearch/document_indexer.rb` -- new service class
- `spec/models/legacy_opensearch/document_indexer_spec.rb` -- full spec coverage

## Testing

1. Check out the branch and make sure the SRCH-6448 changes are present (`String#extract_array` and `Serde` specs)
2. Run the new specs:
   ```
   bundle exec rspec spec/models/legacy_opensearch/document_indexer_spec.rb
   ```
3. Verify specs cover:
   - Happy path indexing (correct body shape, language handling, Serde integration)
   - Happy path deletion
   - Error cases: blank document_id, duplicate ID conflict, transport errors, not-found on delete
   - Edge cases: nil language defaults to 'en', missing optional fields, string-keyed params, created_at preservation vs injection, original params not mutated

### DIFF
--- a/app/models/legacy_opensearch/document_indexer.rb
+++ b/app/models/legacy_opensearch/document_indexer.rb
@@ -1,0 +1,58 @@
+# frozen_string_literal: true
+
+class LegacyOpenSearch::DocumentIndexer
+  class DocumentIndexerError < StandardError; end
+  class DuplicateID < DocumentIndexerError; end
+
+  INDEX_NAME = ENV.fetch('LEGACY_OPENSEARCH_INDEX')
+
+  # Indexes a document into OpenSearch. Accepts the same params shape as
+  # SearchgovUrl#i14y_params. Uses Serde.serialize_hash to transform fields
+  # (language-suffix renaming, HTML sanitization, array conversion, etc.)
+  # before writing.
+  #
+  # This is an upsert: it creates the document if it doesn't exist, or
+  # replaces it if it does. The i14y create/update distinction is unnecessary
+  # here because SearchgovUrl#i14y_params always sends the full field set.
+  def self.index(params)
+    params = ActiveSupport::HashWithIndifferentAccess.new(params.deep_dup)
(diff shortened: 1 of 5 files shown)


selfdanielj (changes requested):

YaritzaGarcia:
Hello! I still see the changes from the commit 4acf0bfd05c8ff4afb7f81455a81c7a712fa559f. Maybe you can run `git revert 4acf0bfd05c8ff4afb7f81455a81c7a712fa559f`.

YaritzaGarcia (approved):

PR #1995, opened 2026-03-19
SRCH-6450: Update SearchgovUrl to index directly to OpenSearch

luisgmetzger:
## What this does

Swaps out the `I14yDocument` HTTP API calls in `SearchgovUrl` so the crawler writes directly to OpenSearch via `LegacyOpenSearch::DocumentIndexer`. This is the critical piece that actually gets fresh indexing working again -- i14y's connection to Elasticsearch has been broken in production since March 18, meaning the 36 SearchGov affiliates have been running on stale snapshot data.

Depends on #1993 (SRCH-6448) and #1994 (SRCH-6449).

## Changes

**`app/models/searchgov_url.rb`** (4 lines changed)

- `index_document`: replaced the `indexed? ? I14yDocument.update : I14yDocument.create` ternary with a single `LegacyOpenSearch::DocumentIndexer.index` call. The new service uses `client.index()` which is an upsert, so the create/update distinction goes away.
- `delete_document`: replaced `I14yDocument.delete` with `LegacyOpenSearch::DocumentIndexer.delete`. Updated the log message format since we're no longer going through i14y.
- `i14y_params` is unchanged -- the DocumentIndexer accepts the same param shape.

**`spec/models/searchgov_url_spec.rb`** (68 insertions, 36 deletions)

- Updated ~25 `I14yDocument` mock references to `LegacyOpenSearch::DocumentIndexer`
- `.create`/`.update` expectations become `.index`
- Error classes remapped (DuplicateID, DocumentIndexerError)
- "already indexed" test now expects `.index` (upsert) instead of `.update`

New specs added for gap coverage:
- Params shape contract (verifies all i14y_params keys reach the indexer)
- DocumentIndexerError on index (503/transport error handled gracefully)
- NotFound on delete succeeds without error

## Testing

1. Check out the branch (needs SRCH-6448 and SRCH-6449 commits which are cherry-picked in)
2. Run the searchgov_url specs:
   ```
   bundle exec rspec spec/models/searchgov_url_spec.rb
   ```
3. Key things to verify:
   - All existing tests still pass with the new service mocks
   - The "already indexed" test uses `.index` instead of `.update`
   - DuplicateID from parallel indexing still results in "OK" status
   - Transport errors are caught and recorded as last_crawl_status
   - Destroy callback still deletes the document
   - Delete failures don't prevent record destruction

## Jira

https://gsa-standard.atlassian-us-gov-mod.net/browse/SRCH-6450

### DIFF
--- a/app/models/legacy_opensearch/document_indexer.rb
+++ b/app/models/legacy_opensearch/document_indexer.rb
@@ -1,0 +1,58 @@
+# frozen_string_literal: true
+
+class LegacyOpenSearch::DocumentIndexer
+  class DocumentIndexerError < StandardError; end
+  class DuplicateID < DocumentIndexerError; end
+
+  INDEX_NAME = ENV.fetch('LEGACY_OPENSEARCH_INDEX')
+
+  # Indexes a document into OpenSearch. Accepts the same params shape as
+  # SearchgovUrl#i14y_params. Uses Serde.serialize_hash to transform fields
+  # (language-suffix renaming, HTML sanitization, array conversion, etc.)
+  # before writing.
+  #
+  # This is an upsert: it creates the document if it doesn't exist, or
+  # replaces it if it does. The i14y create/update distinction is unnecessary
+  # here because SearchgovUrl#i14y_params always sends the full field set.
+  def self.index(params)
+    params = ActiveSupport::HashWithIndifferentAccess.new(params.deep_dup)
+    document_id = params.delete(:document_id)
+    language = params.delete(:language) || 'en'
+    params.delete(:handle)
+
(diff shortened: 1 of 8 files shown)


selfdanielj (changes requested):

YaritzaGarcia:
Hello! I still see the changes from the commit https://github.com/GSA/search-gov/commit/4acf0bfd05c8ff4afb7f81455a81c7a712fa559f. Maybe you can run git revert 4acf0bfd05c8ff4afb7f81455a81c7a712fa559f.

luisgmetzger:
@YaritzaGarcia  - Should be removed now

YaritzaGarcia (approved):

PR #2013, opened 2026-04-15
SRCH-6490: Redirect /search/advanced and /search/news to /search and query params persist

YaritzaGarcia:
## Summary
This PR adds redirects for deprecated/unused routes that are still being accessed (as seen in logs). It also removes deprecated tests that are affected by this change.

Instead of allowing these routes to be processed, we now redirect them to the appropriate destination while preserving any query parameters.

## Changes

* Added redirects for:
  * `/search/news` → `/search`
  * `/search/advanced` → `/search`
* Preserved query parameters using `request.query_string`
* Implemented redirects in Rails routing layer
* Remove deprecated tests.

## Example

Request:

```
http://localhost:3000/search/news?affiliate=usagov&query=money
```
and 
```
http://localhost:3000/search/advanced?affiliate=usagov&query=money
```
---
Redirects to:
```
http://localhost:3000/search?affiliate=usagov&query=money
```

## Alternatives Considered

* **CloudFront redirect**: Rejected due to added complexity when preserving query parameters and lack of flexibility
* Rails solution is simpler and easier to maintain

## Testing
* Manually verified redirects in local environment
* Confirmed query parameters are preserved
* Checked logs to ensure correct behavior

Logs for news redirect:
image

Logs for advanced redirect:
image

 
### Checklist
 
#### Functionality Checks
 
#### Process Checks


config/routes.rb
```diff
@@ -7,6 +7,17 @@
   constraints(AdvancedSearchesConstraint.new) do
     get '/search/advanced', to: redirect(path: '/search')
```
luisgmetzger: This existing redirect uses `redirect(path: '/search')` which drops query parameters. The new redirect you added below (line 11) preserves them. But Rails evaluates routes top-down, so SearchGov affiliates will still hit this line first and lose their `?affiliate=...&query=...` params.

Since the new unconditional redirect on line 11 now handles all requests (SearchGov or not), this entire `constraints` block is dead code and can be removed. That also fixes the query-param bug for SearchGov sites.

config/routes.rb
```diff
@@ -7,6 +7,17 @@
...
   end
+
+  get '/search/advanced', to: redirect { |params, request|
+    query_string = request.query_string.present? ? "?#{request.query_string}" : ""
+    "/search#{query_string}"
+  }
+
+  get '/search/news', to: redirect { |params, request|
+    query_string = request.query_string.present? ? "?#{request.query_string}" : ""
+    "/search#{query_string}"
```
luisgmetzger, line 14: These two route definitions (lines 18 and 20) are now unreachable since the redirects above them always match first. They still register named helpers (`advanced_search_path`, `news_search_path`) but those helpers now just generate URLs that redirect. Worth removing to avoid confusion.

spec/requests/redirect_spec.rb
```diff
@@ -30,3 +30,18 @@
     expect(response).to redirect_to('/sites')
   end
 end
+
+describe "/search/news" do
+  it 'redirects to /search' do
```
luisgmetzger, line 35: The main goal of this PR is preserving query params through the redirect, but these tests only check bare URLs without params. Could you add a case like:

```ruby
it 'preserves query parameters' do
  get '/search/news?affiliate=usagov&query=money'
  expect(response).to redirect_to('/search?affiliate=usagov&query=money')
end
```

Same for the `/search/advanced` block below.

spec/requests/searches_spec.rb
```diff
@@ -4,19 +4,24 @@
...
+    end
+
     before do
-      get '/search/news', params:  { query: 'element',
-                                     affiliate: affiliate.name,
-                                     channel: rss_feeds(:white_house_blog).id }
+      get '/search/news', params: news_search_params
     end
 
     it 'sets the format to html' do
```
luisgmetzger: This test was originally verifying the format of an actual page response. Now that `/search/news` returns a 301 redirect, checking `request.format` on a redirect doesn't really prove anything. Consider removing it since the new redirect assertion on line 22 already covers the behavior.

YaritzaGarcia:
Quick note, let’s hold off on merging into staging. Since we’re planning to move changes from staging to main soon, this PR should be retargeted to main instead of staging.

Let’s wait until the current update to main is finished, then we can adjust this PR accordingly.

PR #2045, opened 2026-05-20
[SRCH-6455] - remove i14y code

luisgmetzger:
## Summary

Remove all i14y code from the codebase per SRCH-6455. The i14y integration is retired — all document search now goes through SearchElastic/OpenSearch engines directly.

**Key changes:**
- Extract shared methods from `I14ySearch` into a `DocumentSearchable` concern, reparent `BlendedSearch` and `SearchElasticEngine` to inherit from `FilterableSearch`
- Delete 26 pure i14y files (models, controllers, views, helpers, observer, libs)
- Update ~35 files with i14y references (controllers, models, views, helpers, config, JS, env)
- Add migration to migrate SearchGov affiliates → SearchElastic, drop `i14y_drawers`/`i14y_memberships` tables, remove `gets_i14y_results`/`i14y_date_stamp_enabled` columns
- Delete 17 spec files, 5 fixtures, 3 VCR cassettes; update 25+ specs and Cucumber features
- Remove `SearchGov` from `SEARCH_ENGINES` constant

**120 files changed, ~5,100 lines removed.**

#### Functionality Checks

- [x] Your primary commit message is of the format **SRCH-#### <description>** matching the associated Jira ticket.
- [x] PR title is either of the format **SRCH-#### <description>** matching the associated Jira ticket.
- [ ] Automated checks pass.

#### Process Checks

## How to test

1. **Verify no i14y references remain:**
   ```
   grep -ri i14y app/ lib/ config/ spec/ features/ --include='*.rb' --include='*.haml' --include='*.ts' --include='*.tsx'
   ```
   Should return nothing (VCR cassettes with base64 noise are expected false positives).

2. **Verify routes are clean:**
   ```
   bin/rails routes | grep i14y
   ```
   Should return nothing.

3. **Run the migration:**
   ```
   bin/rails db:migrate
   ```
   Confirm `i14y_drawers` and `i14y_memberships` tables are dropped, and any `search_gov` affiliates are migrated to `search_elastic`.

4. **Run specs:** `bundle exec rspec`

5. **Run Cucumber features:** `bundle exec cucumber`

6. **Smoke test search engines:** Verify that `BlendedSearch`, `SearchElasticEngine`, `OpenSearch::Engine`, and `LegacyOpenSearch::Engine` still return results for an affiliate using each engine type.

Resolves [SRCH-6455](https://gsa-standard.atlassian-us-gov-mod.net/browse/SRCH-6455)

[SRCH-6455]: https://gsa-standard.atlassian-us-gov-mod.net/browse/SRCH-6455


YaritzaGarcia (dismissed):
Noticed that there are two tests failing:

Failed examples:

rspec ./spec/system/click_tracking_spec.rb:26 # A user searches a searchgov site for a best bet the search results have the expected data attributes

rspec ./spec/system/click_tracking_spec.rb:36 # A user searches a searchgov site for a best bet the user clicks a search result js sends in a click event, creating the expected log line.

db/migrate/20260520010100_remove_i14y_tables_and_columns.rb
```diff
@@ -0,0 +1,37 @@
+class RemoveI14yTablesAndColumns < ActiveRecord::Migration[7.0]
+  def up
+    execute "UPDATE affiliates SET search_engine = 'search_elastic' WHERE search_engine = 'search_gov'"
```
selfdanielj: @luisgmetzger i think maybe this should be opensearch or whatever the value in the db is for that option since search elastic is not valid any more

config/routes.rb
```diff
@@ -24,7 +24,6 @@
     namespace :v2 do
       get '/search' => 'searches#blended'
       get '/search/bing' => 'searches#bing'
-      get '/search/i14y' => 'searches#i14y'
```
selfdanielj, line 27: we need to keep this route to support existing API use.  Can it be pointed to code that does what it used to do?  `searches#blended` doesnt appear to work in the same way?

selfdanielj (changes requested):

app/controllers/api/v2/searches_controller.rb
```diff
@@ -16,17 +16,11 @@ def blended
         respond_with(@search)
       end
 
-      def i14y
```
YaritzaGarcia, line 19: I think this is related to what @selfdanielj and @DK1999 are talking in [Slack](https://gsa-tts.slack.com/archives/CCH1V6KSL/p1779285974185239?thread_ts=1779240589.885909&cid=CCH1V6KSL). Maybe we need to undo this? Not a 100% sure but it looks like this function gets called when a GET request is made to the `api/v2/search/i14y` endpoint.

app/controllers/api/v2/searches_controller.rb
```diff
@@ -108,7 +102,6 @@ def validate_search_options
       def search_options_validator_klass
         case action_name.to_sym
         when :blended then Api::NonCommercialSearchOptions
-        when :i14y then Api::I14ySearchOptions
```
YaritzaGarcia, line 111: https://github.com/GSA/search-gov/pull/2045/changes#r3274906089

Same thing with this line of code

db/migrate/20260520010100_remove_i14y_tables_and_columns.rb
luisgmetzger: Good call. `search_elastic` is still a valid engine in the codebase (`SEARCH_ENGINES = %w[BingV7 SearchElastic OpenSearch LegacyOpenSearch]`), but if the production affiliates currently on SearchGov should land on `open_search` instead, happy to update. Let me know which one is the right target and I'll change it.

config/routes.rb
luisgmetzger: Agreed, this needs to stay. Restored in commit a19a11e. The `/search/i14y` route now points to the `i14y` action which uses `selected_engine` under the hood (routes to `ApiSearchElastic`, `OpenSearch::ApiEngine`, or `LegacyOpenSearch::ApiEngine` depending on the affiliate config). Same behavior as before, just no i14y service code behind it.

luisgmetzger:
Thanks for catching those @YaritzaGarcia. The `click_tracking_spec` failures were caused by the `searchgov_affiliate` fixture still having `search_engine: SearchGov` after we removed that engine type. Fixed in 8e0d7e5 by updating the fixture to `SearchElastic`. Should be green now.

db/migrate/20260520010100_remove_i14y_tables_and_columns.rb
selfdanielj: @DK1999 what do you think?  I don't think there are any search elastic affiliates in prod anyway.

selfdanielj: We discussed this at the team meeting today and agreed it should go to open search

luismetzger: Updated in fca8c88 — the migration now targets `open_search` and I've updated the fixtures and Cucumber features to match.

app/controllers/searches_controller.rb
selfdanielj, line 1: @luisgmetzger what are the different template names that we might see in vertical now?

luisgmetzger: @selfdanielj  - The possible vertical values from pick_klass_vertical_template are:

:web — WebSearch
:document — OpenSearch::Engine, LegacyOpenSearch::Engine, or SearchElasticEngine
:blended — BlendedSearch (when gets_blended_results is true)

Before this PR, :i14y was also a possible value (for SearchGov affiliates). Now those engines all map to :document, which renders the searches/document.html.haml template which is a renamed version of the old searches/i14y.html.haml.

selfdanielj (approved):
I reviewed the code and ran locally, trying out searches both web and API and poking around in the admin center. Looks good to me!

YaritzaGarcia (approved):

PR #2053, opened 2026-05-28
fix(deploy): explicitly fetch SEARCH_AWS_BUCKET in fetch_env_vars.sh

amian-gsa:
SEARCH_AWS_BUCKET may not appear in ssm:DescribeParameters results for some instance roles (confirmed on dev cron in us-west-1) but is readable via ssm:GetParameter. This caused AWS_BUCKET to be missing from .env, making upload_assets_to_s3.sh fail with:
  'AWS_S3_BUCKET or AWS_BUCKET environment variable is not set'

Add an explicit get-parameter call for SEARCH_AWS_BUCKET after the describe loop, using the same direct-fetch pattern as LOGIN_DOT_GOV_PEM. Writes AWS_BUCKET to .env unconditionally so asset upload always has the correct bucket name. Deduplicates any existing AWS_BUCKET line written by the loop before appending the explicit value.

Fixes: dev cron (us-west-1) AfterInstall pipeline failure

### DIFF
--- a/cicd-scripts/fetch_env_vars.sh
+++ b/cicd-scripts/fetch_env_vars.sh
@@ -13,6 +13,10 @@ warn() {
   echo "[CODEDEPLOY][BEFORE_INSTALL][fetch_env_vars][WARN] $*"
 }
 
+error() {
+  echo "[CODEDEPLOY][BEFORE_INSTALL][fetch_env_vars][ERROR] $*" >&2
+}
+
 retry() {
   local attempts="$1"
   local delay_seconds="$2"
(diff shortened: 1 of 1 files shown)


cicd-scripts/fetch_env_vars.sh
```diff
@@ -84,6 +84,24 @@ for PARAM in $PARAM_KEYS; do
...
+    --name "SEARCH_AWS_BUCKET" \
+    --region "$REGION" \
+    --query "Parameter.Value" \
+    --output text 2>/dev/null); then
+  log "Fetched SEARCH_AWS_BUCKET → AWS_BUCKET=${AWS_BUCKET_VALUE}"
+  # Remove any existing AWS_BUCKET line written by the loop, then append
+  sed -i '/^AWS_BUCKET=/d' .env
+  echo "AWS_BUCKET=${AWS_BUCKET_VALUE}" >> .env
+else
+  warn "Could not fetch SEARCH_AWS_BUCKET from region $REGION — AWS_BUCKET will not be set"
```
selfdanielj: What happens to the deployment and app behavior if this doesn't get set?

amian-gsa: Deployment will fail which is the desired behaviour because this parameters hold the bucket name where assets etc are stored.

selfdanielj: Is the resulting error message instructive enough?  What do you think about forcing a fail here to make it clear? This just logs a warning and continues right now right?

selfdanielj (approved):

PR #2060, opened 2026-06-01
SRCH-6571: Fix Rails logger conflict with SemanticLogger

luisgmetzger:
## Summary
- Remove `config.logger = ActiveSupport::Logger.new(STDOUT)` from `config/environments/production.rb`. On the Ubuntu 24.04 dev/staging hosts this explicit logger is wrapped as a SemanticLogger appender whose async thread starts at boot, flushes init messages, then stops -- after which request-level JSON logs go nowhere. `rails_semantic_logger` now owns logging natively (`config.rails_semantic_logger.format = :json`). STDOUT output is still available via `RAILS_LOG_TO_STDOUT=1`, which the gem turns into a proper JSON STDOUT appender (already used by the Dockerfile).
- Add a Capistrano `deploy:clean_log_dir` task hooked `before 'deploy:symlink:linked_dirs'`. Because the repo tracks `log/.gitignore`, every git checkout creates a real `release/log/` dir; the current Capistrano flow then nests `current/log/log -> shared/log`, orphaning SemanticLogger's file appender output in the release dir instead of `shared/log` (which the CloudWatch agent tails). Removing the real dir first makes `linked_dirs` create `current/log -> shared/log` correctly.

## Verification (AWS, before fix)
- CloudWatch `searchgov/production.log`: prod (us-east-1) ~30.7 GB live JSON request logs; staging (us-east-2) ~95 KB; dev (us-west-1) ~5.5 KB. Staging/dev contained only plain-text boot output, no request logs.
- SSM on app1 boxes:
  - prod (Ubuntu 20.04): `current/log -> shared/log` (correct), `shared/log/production.log` ~859 MB, actively written.
  - staging/dev (Ubuntu 24.04): `current/log` is a real dir with nested `log -> shared/log`; orphaned `current/log/production.log` (6469 B / 4938 B) ending in `Async: Thread has stopped`; `shared/log/production.log` 939 B / 0 B.
- OS audit: all prod app boxes are still 20.04, so prod is unaffected today but would break on its next Capistrano deploy regardless of OS -- this fix is also preventive for prod.

## Test plan
- [ ] Deploy to dev (merge to `main`): confirm `current/log -> shared/log` and JSON request logs append to `shared/log/production.log`; confirm CloudWatch `searchgov/production.log` (us-west-1) ingests fresh JSON.
- [ ] Deploy to staging (`staging` branch): same SSM + CloudWatch (us-east-2) checks.
- [ ] Deploy to production (`production` branch, manual approval): monitor us-east-1 ingestion, no regression.

Refs: SRCH-6571


selfdanielj (dismissed):
sure we can try this... for these stories were we need to deploy to test it out, if you create the branch in this the upstream repo instead of the fork we can deploy manually before merging into main to make sure it works.  Using the fork we have to merge to get the changes in and then we have to PR & approve, PR & approve, etc... while we work out the kinks, potentially leaving the repo in a state with broken deployment.

config/deploy.rb
selfdanielj, line 1: is this file still used without capistrano?

luismetzger: Deploys go through CodeDeploy now (appspec.yml + cicd-scripts), so config/deploy.rb isn't even in the path. Reverted that change.

The real culprit is in after_install.sh. It makes a real log dir, then symlinks shared/log onto it, so the link ends up nested inside (log/log). Moved the fix there: skip creating log/tmp and rm them right before the symlinks, so release/log points straight at shared/log. Pushed in b692a7a8.

selfdanielj (approved):
i hope this is it because i really didnt like the log/log thing

PR #2069, opened 2026-06-10
SRCH-6598: Reap orphaned geckodriver when Firefox fails to launch

luisgmetzger:
## What this fixes

The other half of SRCH-6598 (dev's geckodriver/MySQL connection leak). When Firefox fails to launch, Selenium raises before `driver` is ever assigned in JsFetcher, so the `ensure driver.quit` never runs. The geckodriver that was already spawned just lives on forever with inherited file descriptors, including MySQL connections. That's how dev ended up with 875 leaked connections on crawl2 and 401 on crawl1.

## What it does now

Wraps driver creation in a rescue that TERMs any geckodriver children of the current process before re-raising. So even if Firefox is broken, we can't accumulate orphans anymore.

Also fixed a latent spec bug while adding the failure path test: the old spec referenced `Selenium::WebDriver::Driver::Manage`, which doesn't exist (the real class is `Selenium::WebDriver::Manager`). It never failed before because it was passed as a string.

## Test plan
- [x] `bundle exec rspec spec/lib/js_fetcher_spec.rb` (2 examples, 0 failures)
- [x] `bundle exec rubocop` clean on changed files
- [ ] On dev (after the ansible fix lands), fetch a js_renderer URL and confirm `pgrep -c geckodriver` stays at 0

Companion infra fix: GSA/searchgov-ansible#362 (installs the real Mozilla deb instead of the snap wrapper)

### DIFF
--- a/lib/js_fetcher.rb
+++ b/lib/js_fetcher.rb
@@ -10,7 +10,15 @@ module JsFetcher
     options.add_preference('browser.tabs.warnOnClose', false)
     options.add_preference('general.useragent.override', DEFAULT_USER_AGENT)
 
-    driver = Selenium::WebDriver.for(:firefox, options:)
+    begin
+      driver = Selenium::WebDriver.for(:firefox, options:)
+    rescue StandardError
+      # If Firefox fails to launch, Selenium raises before `driver` is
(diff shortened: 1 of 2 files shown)


selfdanielj (approved):
makes sense!

PR #2084, opened 2026-07-06
SRCH-6613: Bump mysql2 0.5.6 -> 0.5.7 for MySQL 8.4 compatibility

luisgmetzger:
## Summary
- Bumps `mysql2` from `~> 0.5.5` (locked `0.5.6`) to `~> 0.5.7` ahead of the RDS MySQL 8.0 -> 8.4 upgrade (spike SRCH-6588).
- 0.5.7 adds MySQL 8.4 to its CI and includes Ruby 3.4 / macOS build fixes.
- **Optional / decoupled:** verified in Docker that mysql2 0.5.6 already connects to MySQL 8.4 (`caching_sha2_password`), so this is not a hard prerequisite for the DB upgrade. It can ship anytime, independent of the dev/staging/prod DB work.
- Lockfile change is limited to mysql2 (0.5.7 declares its `bigdecimal` dependency); no other gems moved and `BUNDLED WITH` is unchanged.

#### Functionality Checks

- [x] PR title is of the format **SRCH-#### \<description\>**.
- [ ] Automated checks pass.

#### Process Checks


Gemfile.lock
```diff
@@ -536,7 +536,8 @@ GEM
     mustermann (3.0.3)
       ruby2_keywords (~> 0.0.1)
     mutex_m (0.2.0)
-    mysql2 (0.5.6)
+    mysql2 (0.5.7)
+      bigdecimal
```
selfdanielj, line 540: does this need to be pinned to anything? i see elsewhere in this file it is.  I don't really know the best practices for this type of file.

luisgmetzger: This one is bundler-generated and matches how Gemfile.lock always works. The bare bigdecimal here is just mysql2 0.5.7's dependency as declared in its gemspec (which sets no version constraint). The concrete version is pinned once, in the GEM section at bigdecimal (3.2.2), and the same bare pattern already shows up under other gems (e.g. lines 166 and 293).

Hand-pinning it here would be non-idiomatic and bundler would rewrite it on the next lock. So no change needed.

selfdanielj (approved):
looks good... we will hold off on this until after db upgrade is shown to work with 0.5.5

PR #2085, opened 2026-07-10
feat(SRCH-6631): gate db:migrate by RUN_MIGRATIONS SSM parameter per tier

amian-gsa:
## Summary
`cicd-scripts/after_install.sh` previously ran `rails db:migrate` unconditionally on every CodeDeploy deployment, regardless of tier (app, api, crawler, cron, spider, proxy, letsencrypt). Since all tiers share the same production RDS instance, a deploy to any non-app tier could apply schema changes the main app fleet's currently-running release does not expect.

## What changed
- Looks up the instance's own `terraform_module` EC2 tag via IMDSv2 + `aws ec2 describe-tags` (same pattern already used in `fetch_env_vars.sh` for region detection)
- Reads the `RUN_MIGRATIONS` JSON value from the shared `.env` file (extracted safely via `grep`/`sed`, not sourced wholesale, since `.env` may contain multiline PEM content)
- Uses `jq` to look up this tier's `true`/`false` value from the JSON, defaulting to `false` if the tier is missing or unresolvable
- Only runs `db:migrate` when the tier's value is exactly `"true"`

**Fail-safe by design:** any failure to resolve the tag, `.env` value, or `jq` availability results in migrations being skipped, never run.

## Why
Addresses the 2026-07-10 production outage where an unconditional `db:migrate` on a crawler-green deployment broke the main app fleet. Full incident writeup: `~/Documents/ai/ubuntu_upgrade/prod_outage_20260710_shared_db_migration_incident.md`.

## Companion change
Paired with a `searchgov-tf` change (same branch name) that adds the `RUN_MIGRATIONS` SSM parameter with `app=true`, all other tiers `=false`.

## Verification performed
- Logic tested standalone against mock JSON/`.env` data
- **End-to-end tested against real dev AWS infrastructure**: applied the companion Terraform parameter to dev, then ran the actual tag-lookup + extraction + `jq` logic via SSM on real `app` and `crawler` dev instances — confirmed `app` tier correctly resolves to `true` (migrate runs) and `crawler` tier resolves to `false` (migrate skipped)
- Confirmed IMDSv2 compatibility on legacy Ubuntu 20.04 production instances (live tag lookup succeeded on `app1.production`) — not just the new Ubuntu 24.04 green fleet

## Rollout notes
- Requires the companion `searchgov-tf` PR to be merged and applied to staging/production before this takes effect there (dev already has the parameter applied for testing).

### DIFF
--- a/cicd-scripts/after_install.sh
+++ b/cicd-scripts/after_install.sh
@@ -146,7 +146,7 @@ SECRET_KEY_BASE=placeholder RAILS_ENV=production ./bin/rails assets:precompile
 # Fail-safe: if the tier tag, RUN_MIGRATIONS value, or jq cannot be resolved
 # for any reason, migrations are SKIPPED (never default to running them).
 get_terraform_module_tag() {
-  local imds_token region instance_id tag_value
+  local imds_token instance_id tag_value
 
   imds_token=$(curl -sS --fail --max-time 2 -X PUT \
     "http://169.254.169.254/latest/api/token" \
@@ -157,21 +157,23 @@ get_terraform_module_tag() {
     return
   fi
 
-  region=$(curl -sS --fail --max-time 2 \
-    -H "X-aws-ec2-metadata-token: $imds_token" \
-    "http://169.254.169.254/latest/dynamic/instance-identity/document" 2>/dev/null \
-    | sed -n 's/.*"region"[[:space:]]*:[[:space:]]*"\([^"]*\)".*/\1/p')
   instance_id=$(curl -sS --fail --max-time 2 \
     -H "X-aws-ec2-metadata-token: $imds_token" \
     "http://169.254.169.254/latest/meta-data/instance-id" 2>/dev/null || true)
 
-  if [ -z "$region" ] || [ -z "$instance_id" ]; then
-    warn "Could not determine region/instance-id; terraform_module tag unknown"
+  if [ -z "$instance_id" ]; then
+    warn "Could not determine instance-id; terraform_module tag unknown"
     echo "unknown"
     return
   fi
 
-  tag_value=$(aws ec2 describe-tags --region "$region" \
+  # Deliberately omit --region: AWS CLI v2 automatically resolves the region
+  # via IMDS when no region is set via env var/profile/flag (confirmed via
+  # `aws configure list` -> "region : <value> : imds" on these instances).
+  # CodeDeploy does not inject a region env var into hook scripts, so this
+  # relies on the CLI's own IMDS fallback rather than duplicating that lookup
+  # with a manual instance-identity-document parse.
+  tag_value=$(aws ec2 describe-tags \
     --filters "Name=resource-id,Values=$instance_id" "Name=key,Values=terraform_module" \
     --query "Tags[0].Value" --output text 2>/dev/null || true)
 
@@ -183,6 +185,7 @@ get_terraform_module_tag() {
 
   echo "$tag_value"
 }
+
 
 # Extract only the RUN_MIGRATIONS line from the shared .env file. Deliberately
 # avoids sourcing .env wholesale in bash, since it may contain multiline
--- a/cicd-scripts/after_install.sh
+++ b/cicd-scripts/after_install.sh
@@ -134,89 +134,32 @@ fi
 log "Precompiling assets"
 SECRET_KEY_BASE=placeholder RAILS_ENV=production ./bin/rails assets:precompile
 
-# SRCH-6631: Only run database migrations on the tier(s) designated as the
-# migration owner (RUN_MIGRATIONS SSM parameter, managed in searchgov-tf).
-# This prevents a deploy to crawler/cron/spider/api/proxy/letsencrypt (which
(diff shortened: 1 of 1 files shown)


cicd-scripts/after_install.sh
```diff
@@ -134,9 +134,86 @@ fi
...
+  imds_token=$(curl -sS --fail --max-time 2 -X PUT \
+    "http://169.254.169.254/latest/api/token" \
+    -H "X-aws-ec2-metadata-token-ttl-seconds: 300" 2>/dev/null || true)
+  if [ -z "$imds_token" ]; then
+    warn "Could not retrieve IMDSv2 token; terraform_module tag unknown"
+    echo "unknown"
+    return
+  fi
+
+  region=$(curl -sS --fail --max-time 2 \
```
selfdanielj: isn't region already available in the codedeploy environment?

selfdanielj (commented):
`+80 -3` seems like way too many new lines for this change... we talk about simplifying things a lot but here you are making four curl requests, an aws cli request, parsing responses with both sed and jq all to just check if we should run a db migration.  

If aws changes something in these responses (we are using `/latest/` after all) we will break right?

Was it not feasible to make a env var available in the pipeline?  Or maybe they are only available in the codebuild environment and those are not duplicated? Mabye there could be a green-fleet codebuild environment to go with the pipepine?

amian-gsa:
Good catch — investigated this before changing anything.

Confirmed CodeDeploy does **not** inject a region env var into hook scripts: checked `AWS_REGION`/`AWS_DEFAULT_REGION` (both empty) and `~/.aws/config` (absent) for the `search` user on a live instance.

However, you're right that we don't need to parse it ourselves either way: AWS CLI v2 automatically resolves the region via IMDS internally when no `--region`/env var/profile is set — confirmed via `aws configure list` → `region : <value> : imds` on live instances.

Pushed a follow-up commit (`db0b4a881`) that drops the manual instance-identity-document parsing and removes the explicit `--region` flag from the `describe-tags` call, letting the CLI's own IMDS fallback handle it. Region is still required and still comes from the same IMDS source — this just removes ~10 lines of redundant/fragile sed-parsing.

Re-verified live against real dev instances, running as the actual `search` user (matching `appspec.yml`'s `runas: search`): `app` tier → `app`, `crawler` tier → `crawler`, same correct results as before the simplification.

cicd-scripts/after_install.sh
amian-gsa: Updated code.

amian-gsa:
> `+80 -3` seems like way too many new lines for this change... we talk about simplifying things a lot but here you are making four curl requests, an aws cli request, parsing responses with both sed and jq all to just check if we should run a db migration.
> 
> If aws changes something in these responses (we are using `/latest/` after all) we will break right?
> 
> Was it not feasible to make a env var available in the pipeline? Or maybe they are only available in the codebuild environment and those are not duplicated? Mabye there could be a green-fleet codebuild environment to go with the pipepine?

updated code.

cicd-scripts/after_install.sh
```diff
@@ -134,9 +134,32 @@ fi
...
+# DEPLOYMENT_GROUP_NAME is set directly by the CodeDeploy agent for every
+# hook script (see hook_executor.rb's @child_envs / Open3.popen3 call) --
+# no IMDS calls, no AWS API calls, and no dependency on IMDS response
+# formats ever changing. This intentionally replaces an earlier
+# implementation that queried IMDS for an EC2 tag and cross-referenced it
+# against an SSM parameter; that approach was reviewed as unnecessarily
+# complex (multiple curl calls + an AWS API call + sed/jq parsing) for a
+# fact CodeDeploy already hands the script for free.
+case "${DEPLOYMENT_GROUP_NAME:-}" in
+  dev_searchgov_dg | staging_searchgov_dg | prod_searchgov_dg)
```
selfdanielj, line 155: looks great!  is it worth making the match value a parameter? `DB_MIGRATION_DEPLOYMENT_GROUP` or something that is set to `dev_searchgov_dg` in dev, `staging_searchgov_dg` in staging, etc... so its not a code change to change it?  Or is it safer that its a code change since it makes it harder to mess up?

amian-gsa: It makes sense this way because these names exist architecturally and will never change. Treating them as variable creates cross functional maintenance point which is unnecessary and does not offer any value but introduce more complexity instead.

selfdanielj: makes sense... i guess i was thinking of in dev and staging for instance all the DGs are named "green".  Should those two be green in this PR?

selfdanielj: and, in the horrible case of another EC2 upgrade, if the existing ones are "green" what are the names of the new ones?

amian-gsa: That's extremely unlikely. These instances will be retired before that. That level of future proofing in this case by adding another ssm parameter is not warrented.

selfdanielj (approved):

PR #2086, opened 2026-07-10
feat(SRCH-6631): gate db:migrate by RUN_MIGRATIONS SSM parameter per tier [staging]

amian-gsa:
## Summary
Cherry-pick of `450203bdb` (main PR #2085) onto a staging-based branch, since main and staging have significant bidirectional divergence (34 staging-only commits vs 10 main-only at time of writing) and a direct main-branch PR into staging would have pulled in unrelated changes.

Updates `cicd-scripts/after_install.sh` to gate `db:migrate` by tier instead of running it unconditionally.

## Why
Addresses the 2026-07-10 production outage where an unconditional `db:migrate` on a crawler-green deployment broke the main app fleet. See `~/Documents/ai/ubuntu_upgrade/prod_outage_20260710_shared_db_migration_incident.md`.

## Companion change
Paired with a `searchgov-tf` staging PR (same branch naming) that adds the `RUN_MIGRATIONS` SSM parameter.

## Verification
- Diff against `origin/staging` confirmed clean/isolated — only this change, no incidental pickup from divergence
- Already end-to-end tested against real dev AWS infrastructure (see main PR #2085)

## Rollout notes
- Requires the companion `searchgov-tf` staging PR to be merged and applied to staging before this takes effect there.

## Related
- Main PR: #2085

### DIFF
--- a/cicd-scripts/after_install.sh
+++ b/cicd-scripts/after_install.sh
@@ -199,7 +199,7 @@ TERRAFORM_MODULE="$(get_terraform_module_tag)"
 # troubleshooting/verification once app/crawler/cron share one deployment
 # group and are no longer distinguishable by deployment group name alone.
 case "$TERRAFORM_MODULE" in
-  app)
+  app | app-green)
     log "Running database migrations (terraform_module=$TERRAFORM_MODULE)"
     RAILS_ENV=production bundle exec rails db:migrate
     ;;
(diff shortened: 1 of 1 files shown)


amian-gsa:
Same fix cherry-picked here from the main PR (#2085) — see that thread for full details on the region-parsing simplification (commit `467802ecc` on this branch, corresponds to `db0b4a881` on main).

Summary: dropped the manual IMDS instance-identity-document parsing for region and removed the explicit `--region` flag from the `describe-tags` call, since AWS CLI v2 already auto-resolves region via IMDS when unset. Re-verified live against dev instances as the `search` user.

selfdanielj (approved):
code worked in dev but as mentioned in slack there are some tag differences that need adjustment

amian-gsa:
> code worked in dev but as mentioned in slack there are some tag differences that need adjustment

Updated the script to include app-green in criteria to run migrations.

selfdanielj (approved):
this needs to be adjusted prior to app upgrade

PR #2088, opened 2026-07-15
feat(SRCH-6631): gate db:migrate on environment + terraform_module + Fleet

amian-gsa:
## Summary
Extends the SRCH-6631 migration-safety gate to be environment-aware, using a single combined EC2 `describe-tags` lookup (`terraform_module`, `Fleet`, `environment`) instead of `terraform_module` alone.

## Problem
Exact-matching `terraform_module=app` alone is correct for dev and production today (both have a plain `app` instance with no `Fleet` tag, no coexisting green fleet), but incorrect for staging, which is mid Ubuntu-24.04 green-fleet upgrade — staging's app tier currently only exists as `app-green` (`Fleet=green`). The old exact-match logic meant staging never auto-ran migrations, forcing manual `db:migrate` on every staging deploy.

## Fix
Case-match on `$ENVIRONMENT:$TERRAFORM_MODULE:$FLEET` together, covering only the 3 real, confirmed-live combinations:
- `dev:app:` (no Fleet tag)
- `staging:app-green:green`
- `production:app:` (no Fleet tag)

No other combination (including hypothetical future coexistence scenarios, e.g. production's own eventual `app-green`) is matched — those will be designed later based on real tag state at that time, per `app_fixes_tracking.md` item #10.

Tag lookup uses a single `aws ec2 describe-tags --output json` call (same API cost as before) parsed with `jq` (confirmed already installed on all app-tier instances in dev/staging/prod) rather than delimiter-split text, avoiding any reliance on whitespace/tab characters to distinguish an absent tag from a parsing artifact.

Fail-safe unchanged: any unresolved tag defaults to `unknown`, which always skips migrations.

## Verification
- **Live-tested via SSM on real instances**, running as the actual `search` user:
  - staging `app-green` (`Fleet=green`) → now runs migrations (the fix)
  - dev `app` (no Fleet tag) → still runs migrations (unchanged)
  - production `app` (no Fleet tag) → still runs migrations (unchanged)

## Related
- Companion main-targeted PR to follow separately
- Design/decision history: `app_fixes_tracking.md` item #10

### DIFF
--- a/cicd-scripts/after_install.sh
+++ b/cicd-scripts/after_install.sh
@@ -199,7 +199,7 @@ TERRAFORM_MODULE="$(get_terraform_module_tag)"
 # troubleshooting/verification once app/crawler/cron share one deployment
 # group and are no longer distinguishable by deployment group name alone.
 case "$TERRAFORM_MODULE" in
-  app)
+  app | app-green)
     log "Running database migrations (terraform_module=$TERRAFORM_MODULE)"
     RAILS_ENV=production bundle exec rails db:migrate
     ;;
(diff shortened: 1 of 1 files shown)


selfdanielj (approved):
👍

PR #2090, opened 2026-07-15
SRCH-6631: Reconcile production CodeDeploy hooks with main + gate legacy Capistrano tiers

amian-gsa:
## Problem

`production`'s `appspec.yml`/`cicd-scripts/` never received the CodeDeploy deploy-hook rework from the Ubuntu 24.04 upgrade (PR #1990 and everything built on top of it since). This blocks landing the SRCH-6631 db:migrate fix in prod, and is a gap in the Ubuntu 24.04 upgrade itself.

Bringing those files over as-is would introduce a second problem: production's `app`/`cron` tiers (`terraform_module=app`/`cron`, no `Fleet` tag — not yet cut over to green) are still deployed today via Capistrano's `cap deploy` (a separate CodeBuild stage in the same pipeline). CodeDeploy's full hook lifecycle already runs against these hosts today, but only `fetch_env_vars.sh`/asset-copy currently do anything meaningful there. The newer hooks create releases, promote `current`, run `bundle install`, and conditionally run `db:migrate` — if that ran unconditionally on these hosts, it would race with Capistrano doing the same job independently on the same release state.

## Changes

1. **`3307b1e59`** — Brings `appspec.yml` and all `cicd-scripts/` files up to `main`'s current state (verified byte-identical to `staging` for these files).
2. **`b341df203`** — Adds `cicd-scripts/lib/tier_gate.sh`, a shared helper sourced by `after_install.sh`, `application_start.sh`, and `validate_service.sh`. Each script resolves `environment`/`terraform_module`/`Fleet` tags once and exits early (no-op) when `environment:terraform_module` is `production:app` or `production:cron` — mirroring the existing SRCH-6631 `db:migrate` gate's pattern rather than a blanket `Fleet=green` check (which would have incorrectly broken dev's untagged `app` tier, already correctly CodeDeploy-owned).

## Verified

- `bash -n` syntax check passed on all 4 changed/added files.
- Isolated functional test of `is_legacy_capistrano_tier()` across 7 tag combinations — `production:app`/`production:cron` skip; `dev:app`, `staging:app-green`, `production:crawler-green`/`app-green` all proceed unaffected.
- Live AWS check (`aws deploy get-deployment-instance`, `aws ec2 describe-instances`, `aws deploy get-deployment-group`) confirms `prod_searchgov_dg` tag-filters on `app`/`cron`/`crawler` together; on merge, the 6 running `app` + 1 `cron` instances will skip the new release/service logic and continue exactly as today under Capistrano. `crawler-green` (already gated-in) is unaffected. Legacy `crawler` instances are currently stopped.

## Net effect

Zero behavior change anywhere on merge. This lands the files safely now so the later `app`/`cron` green cutover doesn't require this file-sync step to be redone, and unblocks landing the SRCH-6631 migration-gating fix in production.

## Follow-up (not in this PR)

When production's `app`/`cron` tiers cut over to `app-green`/`cron-green`, the case statement in `is_legacy_capistrano_tier()` must be updated/retired — same class of follow-up already documented for the `db:migrate` gate in `app_fixes_tracking.md` section 10.

NOTES:

### Additional commits beyond the CodeDeploy hook reconciliation

Since the appspec.yml/cicd-scripts reconciliation was drafted, four more commits landed on this branch:

__`513faf2cc` — Remove Capistrano `resque_worker`/`resque_scheduler` wiring__ `DEPLOY_RESQUE_SERVER_ADDRESSES` (the SSM parameter behind these Capistrano roles) currently resolves to `["10.150.12.43","10.150.12.24"]` — the two legacy crawler instance IPs, stopped since crawler's cutover to the green fleet. Confirmed live on a prod `app` instance: no resque systemd units, no resque processes, no crontab — nothing resque-related runs on `app`/`cron` themselves. These roles never ran resque *on* app/cron; they pointed app/cron's own Capistrano deploy *at the crawler boxes*, to restart resque there as a side effect. Since crawler's SSM address was never updated post-cutover, every app/cron deploy has been silently trying to restart resque on dead instances. Removing this is a no-op for current behavior. Resque today is systemd-managed on `crawler`/`crawler-green` only, via `RESQUE_WORKER_SERVICE`/`RESQUE_SCHEDULER_SERVICE` in `application_start.sh`/`application_stop.sh`/`validate_service.sh`, gated by `tier_gate.sh`. `main`/`staging` still carry the old wiring — tracked separately via Jira for their own cleanup, since their situation differs (no tier there is still Capistrano-managed, so their removal is a bigger, different-shaped change).

__`b9d10f7fa` — Remove dead resque parameter-store entries and `cap resque:*` tasks from `buildspec_searchgov.yml`__ Follow-up to `513faf2cc`: this branch's `buildspec_searchgov.yml` still referenced the two retired SSM parameters via CodeBuild's `parameter-store` mapping, and called `cap $SEARCH_ENV resque:restart`/`resque:scheduler:restart` — which would fail outright once `capistrano-resque` was removed from `Capfile`. Left `cap $SEARCH_ENV deploy`/`puma:*` tasks intact.

__`a5e3ca7` (searchgov-tf) — Fix undeclared `aws_instance.spider_servers_green` reference__ Unrelated pre-existing bug found while validating the `searchgov-tf` side of the resque SSM parameter removal: `terraform/modules/spider/cloudwatch.tf` referenced a resource that was never declared, blocking `terraform validate`/`plan` across all three environments. Fixed in `searchgov-tf` PR #349, not this repo — noted here since it was found in the course of this work.

__`e3fffdff5` — Gate `upload_assets_to_s3.sh` behind `tier_gate.sh`__ Found via review: this hook wasn't gated like the other three, so it would've run unconditionally on prod's legacy Capistrano app tier for the first time (it didn't exist in prod's pre-reconciliation appspec.yml). Confirmed assets currently reach S3 via whichever CodeDeploy-hook-driven tier last deployed (crawler-green today), not via any Capistrano task. Gated identically to `after_install.sh`/`application_start.sh`/`validate_service.sh`.


luisgmetzger:
@amian-gsa - The CodeDeploy and appspec side looks solid. I diffed it and the appspec.yml plus cicd-scripts/ are byte identical to staging, so that whole chunk is basically already proven out. Tier gate does exactly what we want on prod app and cron too.

Got one question before you take it out of draft though. There's a third commit in here (513faf2cc) that rips out the Capistrano resque wiring, the capistrano-resque require, the resque_worker/resque_scheduler roles, :workers, all that. It's not called out in the PR description, and both main and staging still have that wiring, so this is actually pulling prod away from them instead of catching it up.

What's nagging me: prod app/cron are still deploying through Capistrano, and now that the CodeDeploy hooks are gated off on those boxes, I don't think anything would kick resque on deploy anymore, unless resque is already running as a systemd/Ansible service there. If capistrano-resque is still the thing restarting workers, they'd just keep chugging on old code after every deploy, which kind of bumps into the "zero behavior change" line.

So mostly just wondering, how's resque actually managed on prod app/cron these days?

If it's still Capistrano, I'd probably pull this commit out for now and do it after the systemd resque cutover.

If it's already on systemd/Ansible, could you toss a note in the description saying so, and maybe why main/staging are still holding onto the wiring? Honestly might be cleanest to do the removal on main first so prod isn't the weird one out.

None of this blocks the actual deploy hook reconciliation, that part's good to go. Just want to get the resque thing straight first. Thanks!

selfdanielj:
> So mostly just wondering, how's resque actually managed on prod app/cron these days?

Resque isn't running on prod/app cron right?  Its systemd processes on the green crawler that you setup and included in app_start right?

amian-gsa:
> @amian-gsa - The CodeDeploy and appspec side looks solid. I diffed it and the appspec.yml plus cicd-scripts/ are byte identical to staging, so that whole chunk is basically already proven out. Tier gate does exactly what we want on prod app and cron too.
> 
> Got one question before you take it out of draft though. There's a third commit in here ([513faf2](https://github.com/GSA/search-gov/commit/513faf2cca1da7d546572186a3d2909dfc6e756b)) that rips out the Capistrano resque wiring, the capistrano-resque require, the resque_worker/resque_scheduler roles, :workers, all that. It's not called out in the PR description, and both main and staging still have that wiring, so this is actually pulling prod away from them instead of catching it up.
> 
> What's nagging me: prod app/cron are still deploying through Capistrano, and now that the CodeDeploy hooks are gated off on those boxes, I don't think anything would kick resque on deploy anymore, unless resque is already running as a systemd/Ansible service there. If capistrano-resque is still the thing restarting workers, they'd just keep chugging on old code after every deploy, which kind of bumps into the "zero behavior change" line.
> 
> So mostly just wondering, how's resque actually managed on prod app/cron these days?
> 
> If it's still Capistrano, I'd probably pull this commit out for now and do it after the systemd resque cutover.
> 
> If it's already on systemd/Ansible, could you toss a note in the description saying so, and maybe why main/staging are still holding onto the wiring? Honestly might be cleanest to do the removal on main first so prod isn't the weird one out.
> 
> None of this blocks the actual deploy hook reconciliation, that part's good to go. Just want to get the resque thing straight first. Thanks!

- We have couple of JIRA stories to clean main and staging so the concern of drift from main and staging will be addressed through those stories.

- resque_worker/resque_scheduler in config/deploy.rb were never resque running on app/cron — they were roles pointing app/cron's own Capistrano deploy at the crawler boxes, to restart resque there as a side effect of an app/cron deploy. Since crawler cut over to green and its SSM address was never updated (we diverged to different path due to db migration issues and we're still on that path to get to resolve issues to get back to previous path of reconciliation), every app/cron deploy was silently trying to restart resque on stopped, dead instances. Removing this is a no-op for current behavior, not a behavior change — the wiring was already pointing at nothing live.

amian-gsa:
> > So mostly just wondering, how's resque actually managed on prod app/cron these days?
> 
> Resque isn't running on prod/app cron right? Its systemd processes on the green crawler that you setup and included in app_start right?

- resque-worker@.service / resque-worker.target / resque-scheduler.service are systemd-managed (enabled).
resque-worker.target: active, live worker processes confirmed running the searchgov queue set.
- resque-scheduler.service: inactive.
- This is exactly the systemd management built this session in application_start.sh/application_stop.sh/validate_service.sh (RESQUE_WORKER_SERVICE/RESQUE_SCHEDULER_SERVICE), gated to only fire on crawler/crawler-green via tier_gate.sh.

appspec.yml
```diff
@@ -1,20 +1,48 @@
...
   AfterInstall:
+    - location: cicd-scripts/after_install.sh
+      timeout: 1200
+      runas: search
+
     - location: cicd-scripts/copy_non_fingerprinted_assets.sh
       timeout: 300
       runas: search
 
+    - location: cicd-scripts/upload_assets_to_s3.sh
```
luisgmetzger, line 34: This one isn't tier gated the way after_install is, so it actually runs on the prod app boxes that are still on Capistrano. That means every prod app deploy newly kicks off an `aws s3 sync ... --delete` it doesn't do today, so it nudges past the zero behavior change line. Two things worth confirming: is uploading from the legacy app tier intended (how do prod app assets reach S3 today?), and is `--delete` safe with all 6 app instances running this at once against the same bucket? Probably fine since they build identical fingerprinted assets so there's nothing to delete, but worth a quick look before merge.

cicd-scripts/lib/tier_gate.sh
```diff
@@ -0,0 +1,125 @@
...
+    _tier_gate_warn "environment tag not found on this instance; defaulting to unknown"
+  fi
+
+  export TERRAFORM_MODULE FLEET ENVIRONMENT
+}
+
+# Returns success (0) if this instance is a legacy, Capistrano-managed tier
+# that the new release-management/service hooks must not act on.
+# Requires resolve_deployment_tags to have been called first.
+is_legacy_capistrano_tier() {
```
luisgmetzger, line 167: Heads up that this fails open on prod. If IMDS or ec2:DescribeTags hiccups, ENVIRONMENT comes back unknown, this returns false, and after_install/application_start/validate_service all run on a Capistrano box and race the real deploy (creating releases/, flipping current, bundle install). The db:migrate case is separately fail safe so migrations still skip, but the release and symlink logic isn't. Since none of these hooks exist on prod today, it's a brand new failure mode there. Fine if the app/cron instance role definitely has ec2:DescribeTags (looks like you verified live), just flagging the residual risk. If you wanted belt and suspenders you could fail closed on the unknown case, e.g. treat unresolved tags as legacy so the release logic skips rather than runs.

cicd-scripts/fetch_env_vars.sh
```diff
@@ -1,73 +1,180 @@
...
+    --region "$REGION" \
+    --query "Parameter.Value" \
+    --output text 2>/dev/null); then
+  log "Fetched SEARCH_AWS_BUCKET → AWS_BUCKET=${AWS_BUCKET_VALUE}"
+  # Remove any existing AWS_BUCKET line written by the loop, then append
+  sed -i '/^AWS_BUCKET=/d' .env
+  echo "AWS_BUCKET=${AWS_BUCKET_VALUE}" >> .env
+else
+  error "Could not fetch SEARCH_AWS_BUCKET from region $REGION — aborting deployment"
+  exit 1
```
luisgmetzger, line 139: This hook isn't gated either, so it runs on the prod app/cron Capistrano boxes, and this new `exit 1` is a hard abort the old script never had. Low risk since it's identical to staging and `aws ssm get-parameter` succeeds even when describe-parameters hides the key, just noting it's another spot where the legacy tier behavior genuinely changes. Nice side win in this rewrite though: dropping `set -x` and the old `cat .env` stops leaking decrypted secrets into the CodeDeploy logs.

luisgmetzger (commented):
Did a pass on this. Core reconciliation looks great: I diffed it and appspec.yml plus all of cicd-scripts/ are byte identical to main and staging, and tier_gate.sh matches what already merged in #2092/#2093, so that whole surface is well proven. Left a few inline notes. The two I actually care about: upload_assets_to_s3.sh runs ungated on the prod app tier, and the tier gate fails open on prod. Resque removal makes sense given your explanation (confirmed the systemd role lives under the crawler playbook in ansible, pointing at now stopped instances), just worth adding those two commits and the dead instances rationale to the description so the diff isn't a surprise. CI is green.

amian-gsa:
> Did a pass on this. Core reconciliation looks great: I diffed it and appspec.yml plus all of cicd-scripts/ are byte identical to main and staging, and tier_gate.sh matches what already merged in #2092/#2093, so that whole surface is well proven. Left a few inline notes. The two I actually care about: upload_assets_to_s3.sh runs ungated on the prod app tier, and the tier gate fails open on prod. Resque removal makes sense given your explanation (confirmed the systemd role lives under the crawler playbook in ansible, pointing at now stopped instances), just worth adding those two commits and the dead instances rationale to the description so the diff isn't a surprise. CI is green.

Updated PR description with additonal notes.

appspec.yml
amian-gsa: > This one isn't tier gated the way after_install is, so it actually runs on the prod app boxes that are still on Capistrano. That means every prod app deploy newly kicks off an `aws s3 sync ... --delete` it doesn't do today, so it nudges past the zero behavior change line. Two things worth confirming: is uploading from the legacy app tier intended (how do prod app assets reach S3 today?), and is `--delete` safe with all 6 app instances running this at once against the same bucket? Probably fine since they build identical fingerprinted assets so there's nothing to delete, but worth a quick look before merge.

Good catch!

- Confirmed this was a real gap, not just a hygiene nit. Traced how prod app assets currently reach S3: verified live S3 object timestamps in `dgsearch-images/assets/` line up with a `crawler-green` CodeDeploy deployment completion, not any Capistrano task — confirmed no S3 sync exists anywhere in `config/deploy.rb`/`Capfile`. So assets reach S3 today via whichever CodeDeploy-hook-driven tier last deployed, not via the legacy app tier at all. Running this hook ungated on all 6 legacy app instances would've added a new, redundant `--delete` sync per instance on every prod app deploy.

- Fixed in `e3fffdff5` — gated identically to the other three hooks. On the `--delete` safety question: agreed it's likely safe since app/crawler-green build from the same codebase/commit and produce identical fingerprinted output, but I haven't independently verified that with a live multi-instance test, so I'm not overstating confidence there — the gate makes it moot for the legacy tier regardless.

cicd-scripts/lib/tier_gate.sh
amian-gsa: > Heads up that this fails open on prod. If IMDS or ec2:DescribeTags hiccups, ENVIRONMENT comes back unknown, this returns false, and after_install/application_start/validate_service all run on a Capistrano box and race the real deploy (creating releases/, flipping current, bundle install). The db:migrate case is separately fail safe so migrations still skip, but the release and symlink logic isn't. Since none of these hooks exist on prod today, it's a brand new failure mode there. Fine if the app/cron instance role definitely has ec2:DescribeTags (looks like you verified live), just flagging the residual risk. If you wanted belt and suspenders you could fail closed on the unknown case, e.g. treat unresolved tags as legacy so the release logic skips rather than runs.

Another good catch. You're right, and on further discussion this direction is actually backwards — proceeding on an actual legacy host risks two release-management systems racing over the same `current` symlink/release state (real corruption), while skipping on a healthy host is a self-contained, visible failure that resolves on the next deploy. Changed in `f82c2525`: added retry/backoff to the tag lookup (3 attempts, 2s apart) to absorb genuine transient hiccups, then hard-`exit 1` if tags still can't be resolved — fails the deployment loudly on that instance rather than guessing either direction. Verified with 4 scenario tests. Also caught a real bug while writing those tests: the warning logger was writing to stdout, which corrupted the tag JSON on any retry via command substitution — fixed as part of the same commit.

cicd-scripts/fetch_env_vars.sh
amian-gsa: > This hook isn't gated either, so it runs on the prod app/cron Capistrano boxes, and this new `exit 1` is a hard abort the old script never had. Low risk since it's identical to staging and `aws ssm get-parameter` succeeds even when describe-parameters hides the key, just noting it's another spot where the legacy tier behavior genuinely changes. Nice side win in this rewrite though: dropping `set -x` and the old `cat .env` stops leaking decrypted secrets into the CodeDeploy logs.

Agreed this is a real, new behavior on the legacy tier — and worth calling it a deliberate improvement rather than just a low-risk side effect. Failing fast here (`BeforeInstall`, first hook, clear error message) is easier to diagnose than the old behavior of silently proceeding without a required value and having something fail more confusingly downstream. Same fail-loud principle we just applied to `tier_gate.sh` per your other comment. Appreciate the callout on the secrets-leak fix too — that one was a genuine unprompted improvement from the rewrite.

selfdanielj (approved):
🎉

PR #2093, opened 2026-07-15
SRCH-6631: Gate release/service hooks against legacy Capistrano tiers

amian-gsa:
## Problem

Companion PR to #2090 (production) and #2092 (main). Lands the same `tier_gate.sh` gate on `staging` so all three branches converge on identical hook logic.

Production's `app`/`cron` tiers (`terraform_module=app`/`cron`, no `Fleet` tag) are still deployed via Capistrano's `cap deploy`. The CodeDeploy release-management hooks (`after_install.sh`, `application_start.sh`, `validate_service.sh`) must not act on those hosts, since doing so unconditionally would race with Capistrano over release/current/migrate state.

## Changes

Cherry-picked from `b341df203` (search-gov `feature/SRCH-6631-db-migration-flag-azhar-prod`, already in PR #2090):

- Adds `cicd-scripts/lib/tier_gate.sh` — shared helper resolving `environment`/`terraform_module`/`Fleet` tags once and exposing `is_legacy_capistrano_tier()`.
- Wires the gate into `after_install.sh`, `application_start.sh`, `validate_service.sh` — each exits early (no-op) when `environment:terraform_module` is `production:app` or `production:cron`.

## Impact on staging

**None.** `is_legacy_capistrano_tier()` only matches `environment=production`; staging (`environment=staging`) never matches, so the gate is inert here — verified via the same isolated test cases used in #2090 (7 tag combinations, all pass, including `staging:app-green` proceeding unaffected). This PR exercises the sourcing/tag-resolution plumbing on staging's next deploy without ever hitting the skip branch.

## Verified

- `bash -n` syntax check passed on all 4 files.
- Clean cherry-pick onto `staging`, no conflicts.

## Related

- #2090 — same gate, targeting `production`.
- #2092 — same gate, targeting `main`.

### DIFF
--- a/cicd-scripts/after_install.sh
+++ b/cicd-scripts/after_install.sh
@@ -1,6 +1,10 @@
 #!/bin/bash
 set -euo pipefail
 
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]:-$0}")" && pwd)"
+# shellcheck source=lib/tier_gate.sh
+source "$SCRIPT_DIR/lib/tier_gate.sh"
+
 log() {
   echo "[CODEDEPLOY][AFTER_INSTALL] $*"
 }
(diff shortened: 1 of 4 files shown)


selfdanielj (approved):
looks like it will work... not able to test this locally of course

luisgmetzger (approved):

PR #2095, opened 2026-07-20
Fix upload_assets_to_s3.sh gating gap and tier_gate.sh fail-open risk

amian-gsa:
## Background

Two bug fixes originally found and fixed while reviewing PR #2090 (production's CodeDeploy hook reconciliation), cherry-picked here since both bugs exist identically on `main` — this repo's `cicd-scripts/lib/tier_gate.sh` and `cicd-scripts/upload_assets_to_s3.sh` are byte-identical to production's pre-fix versions.

### 1. Gate `upload_assets_to_s3.sh` behind `tier_gate.sh`

This hook was not sourcing `tier_gate.sh`, so it ran unconditionally on every instance — including any legacy Capistrano-managed tier that never had this hook before the Ubuntu 24.04 CodeDeploy migration. Confirmed live: prod app assets currently reach S3 via whichever CodeDeploy-hook-driven tier last deployed, not via any Capistrano task — no S3 sync exists anywhere in `config/deploy.rb`/`Capfile`. Running this hook ungated on a legacy tier would add a new, redundant `aws s3 sync --delete` per instance on every deploy that doesn't happen today.

Gated identically to `after_install.sh`/`application_start.sh`/`validate_service.sh`.

### 2. `tier_gate.sh`: fail closed with retry, instead of fail-open

The previous fail-safe — treating unresolved instance tags as "not a legacy tier" so hook logic proceeds — has an asymmetric risk: proceeding on an actual legacy Capistrano host risks two independent release-management systems racing over the same `releases/`/`current` state (real corruption, the exact scenario this gate exists to prevent). Skipping on a healthy host is a self-contained, visible failure that resolves cleanly on the next deploy attempt.

Changed to fail loudly instead of guessing either direction:
- `get_instance_tags()` now retries the IMDS/describe-tags lookup up to 3 times (2s apart) to absorb a genuinely transient hiccup.
- `resolve_deployment_tags()` now hard-exits (`exit 1`) if tags still can't be resolved after retries, rather than silently defaulting to "unknown" and letting callers proceed.

**Bug caught while writing tests for this change:** `_tier_gate_warn` was writing to stdout, and `get_instance_tags()` is always invoked via command substitution — so warning lines were being captured as part of the JSON return value on any retry, corrupting tag resolution on the very hiccup-recovery path this change was meant to harden. Fixed by moving `_tier_gate_warn` to stderr, matching `_tier_gate_error`. Verified via 4 scenario tests (immediate success x2, retry-then-succeed, persistent-failure → clean exit 1).

## Verification

- `bash -n` syntax check passes on both changed files.
- Verified live via SSM that `ec2:DescribeTags` succeeds on both a dev `app` instance and a staging `app-green` instance, confirming the new fail-closed retry logic has no permission-related risk on either environment.
- 4 scenario tests run locally against `tier_gate.sh`'s retry/fail-closed logic (immediate success, crawler-green not-legacy case, retry-then-succeed, persistent-failure exit code + clean stdout).

## Related

- `search-gov` PR #2090 (original findings, production-scoped fixes)

### DIFF
--- a/cicd-scripts/lib/tier_gate.sh
+++ b/cicd-scripts/lib/tier_gate.sh
@@ -25,11 +25,24 @@
 # use one consistent, explicit definition rather than relying on incidental
 # no-op behavior (e.g. unit-name mismatches) to stay safe on those hosts.
 #
-# Fail-safe: if instance tags cannot be resolved for any reason, this treats
-# the host as NOT a legacy tier (i.e. hook logic proceeds). Silently skipping
-# hook logic on a host we can't identify is a bigger risk than proceeding,
-# since dev/staging/other green hosts must not be accidentally skipped just
-# because of a transient IMDS/AWS CLI hiccup.
+# Fail-safe: if instance tags cannot be resolved after retries, this ABORTS
+# the deployment on that instance (hard exit 1) rather than guessing either
+# direction. An earlier version of this fail-safe treated unresolved tags as
+# "not a legacy tier" (i.e. let hook logic proceed), reasoning that silently
+# skipping on a healthy green host was the bigger risk. On reflection that
+# is backwards: proceeding on an actual legacy Capistrano host risks two
+# independent release-management systems racing over the same releases/
+# current state -- creating releases, flipping symlinks, restarting
+# services -- which is exactly the corruption scenario this gate exists to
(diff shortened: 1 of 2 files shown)


luisgmetzger (approved):

selfdanielj (approved):
👍

PR #2061, opened 2026-06-02
SRCH-6468: Remove Elasticsearch-specific code from search-gov

luisgmetzger:
## Summary

Collapses the dual Elasticsearch/OpenSearch stack in search-gov down to **OpenSearch only**, per SRCH-6468.

- Removes the `SearchElastic` engine, `ApiSearchElastic`, `lib/search_elastic/*`, and the `search_elastic` affiliate option; adds a migration moving existing `SearchElastic` affiliates to the `OpenSearch` engine.
- Removes X-Pack Watchers, `ElasticsearchDeleteByQueryJob`, the Elasticsearch health check, and the analytics data migrator/tooling.
- Makes `Es::ELK`, `Indexable`, and the custom indices OpenSearch-only; drops `Es::CustomIndices`, the ES initializers, `elasticsearch_client.yml`, and the `OPENSEARCH_APP_ENABLED` flag.
- Strips `ES_*` env vars, removes `elasticsearch-xpack`/`-persistence`/`-model` gems (keeps `elasticsearch` as the OpenSearch transport), and removes the Elasticsearch container/tasks from CircleCI.

## OpenSearch-only fixes surfaced by the test suite

These paths were Elasticsearch-only before the collapse and broke once routed to OpenSearch:

- `ElasticBlendedResults` now resolves the source model from `hit['_index']` (OpenSearch no longer returns `_type`).
- `LogstashDeduper` filters by the `type` field in the query body instead of the removed ES `type` URL param.
- The `an indexable` shared example targets OpenSearch instead of a hardcoded ES host.

## Test plan

- [x] Full `rspec` run locally against OpenSearch 3.2 (search-services) — all OpenSearch/ES-related specs green.
- [x] Exercised search (`/api/v2/search` blended + i14y) and analytics paths.
- [x] CI green (rspec + cucumber).
- [ ] Confirm affiliate data migration runs cleanly in staging.

_Note: remaining local spec failures are environmental only (Tika under arm64 emulation, headless browser driver) and pass in CI._

PR #2100, opened 2026-07-23
SRCH-6653 Gate custom Elasticsearch indices behind CUSTOM_INDICES_ENABLED

luisgmetzger:
## Summary
Addresses SRCH-6653, a broken-out, minimal slice of SRCH-6468 (full ES removal, tracked in #2061, which is blocked on the server + MySQL upgrade). This just stops the production log noise now.

**Problem:** The legacy Elasticsearch cluster is decommissioned, and prod `ES_HOSTS` has been repointed at a placeholder:

```
$ aws ssm get-parameter --name ES_HOSTS
"Value": "https://es-is-off-1.prod.infr.search.usa.gov:9200,https://es-is-off-2.prod.infr.search.usa.gov:9200"
```

`Es.custom_indices_enabled?` used `ENV['ES_HOSTS'].present?`, which is still `true` for that placeholder, so every related-searches / custom-indices read still tries to connect to the dead cluster. The calls are rescued (`Faraday::ConnectionFailed` -> empty results), so users are unaffected, but they spam `production.log` with thousands of connection failures per minute and do wasted work on nearly every i14y search.

**Fix:** Stop inferring this from the contents of `ES_HOSTS` and gate it on an explicit `CUSTOM_INDICES_ENABLED` flag, following the existing `OPENSEARCH_APP_ENABLED` / `OpenSearchConfig.enabled?` convention:

```ruby
def self.custom_indices_enabled?
  ENV['CUSTOM_INDICES_ENABLED'].to_s.casecmp?('true')
end
```

The flag is off unless a deployment explicitly opts in, which short-circuits the four legacy custom-indices code paths (`SaytSuggestion.related_search`, `NewsSearch#search`, `ApiVideoSearch#search`, `Indexable#search_for`) before they hit the dead cluster. `ES_HOSTS` is left alone and can hold whatever it holds.

**Behavior impact:** None user-facing. Those paths already returned empty (rescued) results in prod; now they return empty without the failed network call and log spam.

### Per-environment configuration

| Environment | Flag | Notes |
| --- | --- | --- |
| Production | unset -> disabled | No parameter-store change required; goes quiet on deploy |
| Staging | unset -> disabled | Already the case before this PR — staging Terraform never defined `ES_HOSTS` |
| Development | `CUSTOM_INDICES_ENABLED=true` | Local ES container via searchgov-services |
| Test | `CUSTOM_INDICES_ENABLED=true` | CircleCI runs a real ES 7.17.7 and the suite exercises these paths |

Dev and test stay opted in deliberately: CircleCI stands up Elasticsearch 7.17.7, runs `rake usasearch:elasticsearch:create_indexes`, and a large share of the rspec + cucumber suite covers the custom-indices paths against it. Turning those off means removing that coverage and the ES service from CI, which is SRCH-6468's scope.

### Review notes

Two questions came up in review, both resolved by dropping the `ES_HOSTS` dependency:

- **Is `ES_HOSTS` a JSON array needing `.value`?** No. `{ value = [...] }` is Terraform's `aws_ssm_parameter` resource syntax, not the runtime value. Separately, `terraform/modules/parameters/paramter.tf` sets `lifecycle { ignore_changes = [value] }`, so the value in `variables.tf` is only used at creation time and the live parameter (version 5, modified 2026-03-31) has since diverged.
- **Isn't the existence of `ES_HOSTS` enough?** Existence-only is exactly what `main` does today, and it's the bug — the placeholder is present, so the check always returns `true`.

### Checklist
- [x] PR title is of the format **SRCH-#### \<description\>**.
- [ ] Automated checks pass.

Note: opened as draft. Part of Azhar's ES decommission plan (SRCH-6468).


lib/es.rb
```diff
@@ -10,8 +10,17 @@ module Es
...
+  # been decommissioned (e.g. "es-is-off-1.prod...."). The hostnames are kept in
+  # place instead of being removed from the parameter store, so we detect the
+  # marker here and treat custom indices as disabled. This stops the app from
+  # repeatedly attempting (and logging) failed connections to the dead cluster.
+  # Full Elasticsearch removal is tracked in SRCH-6468.
+  DISABLED_HOSTS_MARKER = 'es-is-off'
+
   def self.custom_indices_enabled?
-    ENV['ES_HOSTS'].present?
+    hosts = ENV['ES_HOSTS'].to_s
```
jayandra: I see the terraform variable as [{ value = [...]}](https://github.com/GSA/searchgov-tf/blob/main/terraform/environments/production/variables.tf#L631). Do we need to do `ENV['ES_HOSTS'].value` to read the array?

lib/es.rb
```diff
@@ -10,8 +10,17 @@ module Es
...
+    hosts.present? && hosts.exclude?(DISABLED_HOSTS_MARKER)
```
selfdanielj: why does it matter what it says in there?  i think its pretty clear that we don't use these values or ES_HOSTS.  Could its existence be enough for this check at this point?

jayandra: So basically always returning true? I am saying so because the list of servers are hard-coded into terraform variable and will always be there, right?

selfdanielj:
Also why wouldn't we set this to behave like we want in dev/staging? Instead of saying "no-op..." I think we should config lower environments to not do this as well.

lib/es.rb
luisgmetzger: No, `ENV['ES_HOSTS']` is a plain string — the `{ value = [...] }` shape is Terraform's `aws_ssm_parameter` resource syntax, not part of the runtime value the container sees.

Worth flagging that the Terraform file is also stale here. `terraform/modules/parameters/paramter.tf` sets `lifecycle { ignore_changes = [value] }`, so `variables.tf` only supplies the value at parameter *creation* and anything changed afterward in AWS is invisible to Terraform. The live parameter today is:

```
$ aws ssm get-parameter --name ES_HOSTS
"Value": "https://es-is-off-1.prod.infr.search.usa.gov:9200,https://es-is-off-2.prod.infr.search.usa.gov:9200"
"Version": 5, "LastModifiedDate": "2026-03-31"
```

Plain comma-separated, which is what `config/elasticsearch_client.yml` expects (it does `ENV['ES_HOSTS']&.split(',')`).

That said, I've dropped this line entirely — see the reply on Daniel's thread. The check no longer looks at `ES_HOSTS` at all.

lib/es.rb
luisgmetzger: Existence-only is what `main` does today — `ENV['ES_HOSTS'].present?` — and that's precisely the bug. The decommission placeholder is still *present*, so it returns `true` and every related-searches / custom-indices read keeps dialing the dead cluster. So yes, @jayandra, presence-only means always `true`.

But the underlying objection is fair: sniffing a magic substring out of a hostname list is fragile and puts config meaning in the wrong place. Pushed bfb9fba to drop it and use an explicit flag instead, following the existing `OPENSEARCH_APP_ENABLED` / `OpenSearchConfig.enabled?` convention:

```ruby
def self.custom_indices_enabled?
  ENV['CUSTOM_INDICES_ENABLED'].to_s.casecmp?('true')
end
```

Off unless a deployment explicitly opts in, so `ES_HOSTS` can hold whatever it holds and prod stops dialing the dead cluster on deploy with no parameter-store change needed.

luisgmetzger:
@selfdanielj on configuring the lower environments rather than leaving it a code no-op — agreed, and that's now the shape of the change. Rundown of where each environment lands:

**Staging** already behaves the way you want, and did before this PR. `terraform/environments/staging/variables.tf` never defines `ES_HOSTS` at all, so `custom_indices_enabled?` has always been `false` there. Nothing to change.

**Production** is off by default now. Since the flag defaults to disabled when unset, and no `CUSTOM_INDICES_ENABLED` parameter exists, prod goes quiet on deploy with no parameter-store change required.

**Dev and test** are explicitly `CUSTOM_INDICES_ENABLED=true` in `.env.development` / `.env.test`. That's a deliberate difference rather than an accident of string contents: CircleCI stands up a real Elasticsearch 7.17.7 container, runs `rake usasearch:elasticsearch:create_indexes`, and a large share of the rspec + cucumber suite exercises the custom-indices paths against it (`SaytSuggestion.related_search`, `NewsSearch#search`, `ApiVideoSearch#search`, `Indexable#search_for`). Flipping those envs off doesn't just change a flag, it means deleting or skipping that coverage and tearing the ES service out of the CI config — which is the work SRCH-6468 / #2061 is scoped for, and it's blocked on the server + MySQL upgrade.

So the config is now per-environment and explicit in all four places; dev/test are the only ones opted in, and they're the only ones with a live cluster to talk to. Happy to fold the dev/test teardown into this PR instead if you'd rather not wait for SRCH-6468, but it'd grow this well past the "stop the log spam now" slice.

jayandra (approved):
Relying upon new env var `CUSTOM_INDICES_ENABLED` as feature flag to enable/disable Elasticsearch LGTM. Unless the flag is explicitly set to `true`; it defaults to `false` 🚀

PR #2106, opened 2026-07-24
fix(deploy): gate ApplicationStop against legacy Capistrano tiers

amian-gsa:
## Problem

application_start.sh, after_install.sh, upload_assets_to_s3.sh, and validate_service.sh already skip on production's legacy Capistrano-managed tiers (`app`/`cron`) via `is_legacy_capistrano_tier()`. `application_stop.sh` was the one hook missing this gate — it unconditionally stopped Puma on every instance regardless of tier.

That asymmetry meant CodeDeploy unconditionally stopped Puma on legacy app/cron instances during every deploy, but `application_start.sh` correctly no-op'd and never restarted it — leaving Puma down until Capistrano's separate, later pipeline stage (`DeployAppWithCapistrano`) got around to restarting it via its own `Capistrano::Puma::Systemd` mechanism.

## Evidence

Confirmed in production on 2026-07-24: all 6 app instances went unhealthy simultaneously during a deploy (ALB `UnHealthyHostCount` peaked at 6), causing a ~4,000-request 5XX spike over roughly 10 minutes. The same-day deployment to staging (no legacy tier at all — all instances tagged app-green/crawler-green/cron-green) showed only brief, single-instance, non-overlapping outages (peak `UnHealthyHostCount` of 1, ~774 total 5XX), confirming the legacy-tier gate asymmetry as the root cause rather than anything code/deploy-artifact specific.

## Fix

Gating `ApplicationStop` here too makes Capistrano the sole owner of Puma's lifecycle on the legacy tier — CodeDeploy no longer touches it at all there, matching the pattern already used by the other three hooks.

## Verification

`bash -n cicd-scripts/application_stop.sh` — syntax valid.

Note: superseded PRs #2104/#2105 (which contained an earlier, separate fix for this branch) were already merged before this commit was pushed, so this is a fresh PR.

### DIFF
--- a/cicd-scripts/application_stop.sh
+++ b/cicd-scripts/application_stop.sh
@@ -1,6 +1,10 @@
 #!/bin/bash
 set -euo pipefail
 
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]:-$0}")" && pwd)"
+# shellcheck source=lib/tier_gate.sh
+source "$SCRIPT_DIR/lib/tier_gate.sh"
+
 if [ -f /home/search/.config/searchgov-codedeploy.env ]; then
   set -a
   # shellcheck disable=SC1090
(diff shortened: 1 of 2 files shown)


selfdanielj:
@amian-gsa these changes seem ok but should we also be gating `before_install.sh`??  i guess its not causing a problem but it is the only "new" one of these that isn't gated any longer, after this change

jayandra:
Good Callout @selfdanielj 👍 
I second on adding necessary gates to`before_install.sh` such that cap deployment and pipeline deployment would have different paths entirely.

amian-gsa:
> @amian-gsa these changes seem ok but should we also be gating `before_install.sh`?? i guess its not causing a problem but it is the only "new" one of these that isn't gated any longer, after this change

Good catch. Fixed.

jayandra (approved):
LGTM 🚀

PR #2125, opened 2026-08-13
MtoS: SRCH-6653

luisgmetzger:
## Summary
Cherry pick of [SRCH-6653](https://github.com/GSA/search-gov/pull/2100) / [SRCH-6653](https://gsa-standard.atlassian-us-gov-mod.net/browse/SRCH-6653), which has been QAed and accepted in Dev.

Gates legacy custom Elasticsearch indices behind `CUSTOM_INDICES_ENABLED` (off unless explicitly `true`). Staging still uses `ENV['ES_HOSTS'].present?`, and staging SSM still has `ES_HOSTS` set to `es-100/200/300.staging.infr.search.usa.gov`, so those hosts still get dialed today. The flag is unset in SSM in prod, staging, and dev — no parameter-store change required; custom indices stay disabled after deploy.

Source PR: https://github.com/GSA/search-gov/pull/2100

`.env.development` / `.env.test` from #2100 (`CUSTOM_INDICES_ENABLED=true` for local/CI) are not in this MtoS. Staging and prod do not use those files. If CircleCI on this branch fails custom-indices specs, we can add that one line to `.env.test` to match main.

### Checklist
- [x] PR title is either of the format **SRCH-#### \<description\>** matching the associated Jira ticket, or **Release - SRCH-####, SRCH-####, SRCH-####** matching the Jira ticket numbers in the release.
- [ ] Automated checks pass.

[SRCH-6653]: https://gsa-standard.atlassian-us-gov-mod.net/browse/SRCH-6653
[SRCH-6653]: https://gsa-standard.atlassian-us-gov-mod.net/browse/SRCH-6653


jayandra:
@luisgmetzger  The `Es.custom_indices_enabled?` flag is checked in the `Indexable#search_for()` method, which is used by six classes:
```
ElasticBlended
ElasticBoostedContent
ElasticFeaturedCollection
ElasticFederalRegisterDocument
ElasticIndexedDocument
ElasticSaytSuggestion
```
With `CUSTOM_INDICES_ENABLED` not being set in production, searches from all these classes would return empty results unless they have a `use_opensearch?` override. Currently, only `ElasticBoostedContent` has this override and will continue to search (via OpenSearch).

This means searches from the other five classes are effectively disabled in production by silently returning no results. This is the intended behavior we wanted right? Just wanted to validate my understanding.

selfdanielj:
why does this need to go to staging now?

luisgmetzger:
@jayandra Yes — that reading is correct, and it is the intended behavior.

`Indexable#search_for` returns empty `NO_HITS` unless `use_opensearch?` or `Es.custom_indices_enabled?`. Only `ElasticBoostedContent` overrides `use_opensearch?` (text best bets, already on OpenSearch). The other five stay on the legacy custom-indices path:

- `ElasticBlended`
- `ElasticFeaturedCollection`
- `ElasticFederalRegisterDocument`
- `ElasticIndexedDocument`
- `ElasticSaytSuggestion`

In production those five already return empty today: `ES_HOSTS` is the `es-is-off` placeholder, the connection fails, and Faraday is rescued into empty results. Users are unaffected; the only change here is we skip the failed network call so we stop spamming `production.log`.

Boosted content continues to search via OpenSearch. Migrating the remaining five off Elasticsearch is [SRCH-6468](https://gsa-standard.atlassian-us-gov-mod.net/browse/SRCH-6468), not this slice.

[SRCH-6468]: https://gsa-standard.atlassian-us-gov-mod.net/browse/SRCH-6468

luisgmetzger:
@selfdanielj This is the MtoS of [#2100](https://github.com/GSA/search-gov/pull/2100) so the already-QAed Dev change can promote staging → prod.

Staging still dials Elasticsearch today. Staging `lib/es.rb` still uses `ENV['ES_HOSTS'].present?`, and staging SSM still has `ES_HOSTS` set to `es-100/200/300.staging.infr.search.usa.gov`. Terraform never defined that parameter for staging, but `lifecycle { ignore_changes = [value] }` left the old value in place, so those hosts still get hit on every custom-indices read.

`CUSTOM_INDICES_ENABLED` is unset in staging SSM, so after this deploy staging matches prod: flag off, no ES dial, empty results from the five legacy classes, boosted content still on OpenSearch. No parameter-store change required.

That is the same lower-env behavior you asked for on #2100 — staging stops talking to ES rather than depending on host-string contents. Prod log noise is the user-facing reason to promote now; the path is main → staging → prod.

luisgmetzger:
Follow-up from Jayandra’s review: with the flag off, `Indexable#search_for` returns `NO_HITS` (no aggregations). `ElasticFederalRegisterDocumentResults` assumed aggregations were present and raised `undefined method \`map' for nil` instead of returning empty results.

That would have been a staging/prod regression for the Federal Register govbox. Pushed a nil-safe constructor plus a spec so the five legacy classes all return empty, and `ElasticBoostedContent` still searches via OpenSearch.

app/models/elastic_federal_register_document_results.rb
```diff
@@ -5,7 +5,7 @@ class ElasticFederalRegisterDocumentResults < ElasticResults
   def initialize(result)
     @offset = result['hits']['offset']
     @aggregations = extract_aggregations(result['aggregations']) if result['aggregations']
-    hits = @aggregations.map { |bucket| bucket.rows.map(&:value) }.flatten
+    hits = Array(@aggregations).map { |bucket| bucket.rows.map(&:value) }.flatten
```
jayandra, line 8: Good catch 👍 This ensure map doesn't get called on nil and error out.

Does similar `nil.map` issue get raised on other `*_document_results.rb` as well where `search_for(options)` tries to call ElasticSearch? Maybe we can address it separately.

jayandra (approved):

PR #2126, opened 2026-08-18
SRCH-6600: Admin Center analytics are driven only by human-logstash-*

jayandra:
## Summary
- Hard-code logstash_prefix to "human-logstash-"
- Removes `sees_filtered_totals`
- Remove the “Stop filtering bot traffic” Admin Center toggle
- Monthly, Yearly emails and CSV downloads of reports in Monthly Reports tab have their values (clicks, queries etc) only from "human-logstash-*"

## Testing Steps
- Verify that in `/sites/:id`; the filter icon at top with hover text "Stop filtering bot traffic" is gone
- Verify the counts in Analytics tab (Queries, Clicks, Monthly Report) with bot-filtering on before deploy matches to the count after deploy
- Verify the CSV report generated under "Download Top Queries" in Monthly-Report tab doesn't have bot related headers and the human-traffic traffic count before and after deploy matches.
 
#### Functionality Checks
 
#### Process Checks


luisgmetzger:
The rspec/cucumber failures were not the Elasticsearch `AccessControlException` / exit 143. That is CircleCI sending SIGTERM to the ES sidecar after the job already failed.

The real error in `bundle exec rake db:setup` was:

```
You have 1 pending migration:
  20260817220000 RemoveBotTrafficAnalyticsToggle
Run `bin/rails db:migrate` to update your database then try again.
```

CI loads `db/schema.rb` rather than running migrations, so the new `sees_filtered_totals` drop needed to be reflected in schema. I pushed that update to this branch (`1473b1fd7`) so CircleCI can rerun.

app/models/rtu_date_range.rb
```diff
@@ -26,7 +27,7 @@ def default_end
 
   def search(query_body)
     Es::ELK.client_reader.search(
-      index: 'logstash-*',
+      index: "#{logstash_prefix}*",
```
luisgmetzger, line 30: This is the one read path that previously always used unfiltered `logstash-*`, including for the Analytics date picker via `RtuAnalyticsRequestable`.

If `human-logstash-*` is an alias on every daily `logstash-*` index, available dates should match. If some days only have bot traffic (or the human alias is missing), the month picker and default range will shrink after deploy.

Worth confirming against production aliases before merge.

app/models/rtu_popular_raw_human_array.rb
```diff
@@ -16,17 +16,10 @@ def most_popular
...
-    return INSUFFICIENT_DATA if human_cnt_hash.empty?
+    cnt_arr = query_class.new(date_range_top_n_query.body).top_n
+    return INSUFFICIENT_DATA if cnt_arr.empty?
 
-    raw_human_arr = raw_cnt_arr.map do |popular_term, raw_count|
-      human_count = human_cnt_hash[popular_term] || 0
-      [popular_term, raw_count, human_count]
-    end
-    raw_human_arr.sort_by { |a| -a.last }.first(@num_results)
+    cnt_arr.sort_by { |a| -a.last }.first(@num_results)
```
luisgmetzger, line 22: This no longer merges raw + human counts; it returns a single `[term, count]` from `human-logstash-*`. That matches the product change.

The `RawHuman` class names (`RtuPopularRawHumanArray`, `RtuQueryRawHumanArray`, `RtuClickRawHumanArray`) now describe an old shape. Fine to leave for this PR, but a follow-up rename would make the mailers/dashboard easier to read.

db/migrate/20260817220000_remove_bot_traffic_analytics_toggle.rb
```diff
@@ -0,0 +1,11 @@
+class RemoveBotTrafficAnalyticsToggle < ActiveRecord::Migration[7.1]
+  def up
+    remove_column :users, :sees_filtered_totals
```
luisgmetzger: Two deploy notes:

1. Ship **code before (or with) this migration**, not migration-first. Old code still calls `sees_filtered_totals?` and will 500 if the column is already gone.
2. `down` reloads the current template files, which are the new 2-column ERB. A rollback restores the column but not the old bot/human email HTML.

Also, `EmailTemplate.load_default_templates` does not depend on this column, so loading templates before `remove_column` would make a failed template reload less likely to leave analytics in a half-migrated state.

luisgmetzger (commented):
The bot-filter removal is thorough and internally consistent: `logstash_prefix` is hard-coded to `human-logstash-`, `sees_filtered_totals` and the Admin Center toggle are gone, and the call sites (analytics controllers, mailers, CSV, SAYT discovery) were updated together with specs and email templates.

CI was failing on `db:setup` because `schema.rb` still had the old column/version; that is now in this branch.

A few things to confirm before merge:

- **`RtuDateRange`** previously always queried unfiltered `logstash-*` (date picker / default range). It now uses `human-logstash-*`. Confirm production has human aliases for the same days.
- **Deploy order:** run this migration *after* (or with) the code deploy. Old code still calls `sees_filtered_totals?`. `down` also cannot restore the old 3-column email HTML because those templates are no longer in the repo.
- **History:** blank commit + revert + re-add of the migration. Squash before merge if possible.
- Optional follow-up: `Rtu*RawHumanArray` names now describe a shape that no longer exists; a cucumber check that the filter toggle is gone would lock the UI change.

Not blocking on those. The application-code sweep looks complete.

app/models/rtu_date_range.rb
jayandra: Ok, will look into it

jayandra: I couldn't find an easy way to look for `human-logstash-*` alias which have 0 docs in them when the original index has some bot-entries. 

However, running `GET _cat/aliases/human-logstash-*?h=alias,index&s=index` to list all aliases and `GET _cat/indices/logstash-*?h=index,docs.count&s=index` to list all indices returned the same number of entries. There were a total of **567** indices listed with date ranges from **2025.02.01** to **2026.08.21**. So, I think it should be safe to assume that we don't see days with 100% bot queries and for any `logstash` created, an corresponding `human-logstash` does get created.

app/models/rtu_popular_raw_human_array.rb
jayandra: Agree, lets do another rename separately as this PR is already getting long.

db/migrate/20260817220000_remove_bot_traffic_analytics_toggle.rb
jayandra: I agree on moving the template loading step before the `remove_column`. However, for the rollback template loading, I still don't have a clear solution. Running it would only make sense only after the code has been reverted.

I am now leaning towards deleting the template reloading line from rollback. I guess, after code rollback if we don't reload the template, it would still be functional but would be missing the column showing `total_unfiltered_queries`, would you agree?

luisgmetzger (approved):

PR #2143, opened 2026-09-04
Added some docs

jayandra:


docs/boosted_content_guide.md
```diff
@@ -0,0 +1,351 @@
...
+- **URL** - Link to the content
+- **Keywords** - Search terms that trigger this result (optional)
+
+---
+
+## Data Flow Architecture
+
+BoostedContent follows a **two-stage data flow**:
+
+```
```
selfdanielj, line 22: i think these would look better as a mermaid.js diagram... not sure if you are familiar but here is an example for the stage one diagram

```mermaid
flowchart TD
    subgraph Stage1["Stage 1: Data Indexing (MySQL ➔ OpenSearch)"]
        direction TB
        A[("BoostedContent (MySQL)")] 
        --> B["ElasticBoostedContentData (transforms data)"]
        --> C["ElasticResqueIndexer (batch processor via Resque)"]
        --> D["ElasticBoostedContent.index() (sends to OpenSearch)"]
        --> E[("OpenSearch Index")]
    end
```

jayandra: The ascii has been replaced with mermaid diagrams

selfdanielj (approved):
✍️

PR #2129, opened 2026-08-24
SRCH-6637: Search non-english pdf for English affiliates

jayandra:
## Summary
- English affiliates were dropping PDFs whose language detector stored text on a non-`en` suffix (`title_es`, `content_de`, etc.).
- For English searches, keep HTML on `*_en`, but also match PDFs on other mapped language suffixes. Filter is `(language: en) OR (mime_type: application/pdf)`.
- Request highlights and `_source` (`title_*`) for those suffixes so a German/Spanish PDF can return a title and snippet. Non-English affiliates are unchanged (strict language filter, no PDF exception).
- `DocumentSearchResults` still deserializes title/highlights from the **document** language, not the affiliate language.

## Localhost Test:
To test it locally, I basically updated an existing text to be Spanish equivalent with "jayandra" added in the title/description/content. Upon searching for "jayandra", I did see it in the results page. If I switch out of the feature branch to `main` and search for "jayandra" again, I would see no results.
```
POST development-i14y-documents-searchgov-legacy/_update/42b7afd3d232f6a6ca9e06fcfac2e4b10a32a3766285899993c167ece31a9949
{
  "script": {
    "lang": "painless",
    "source": """
      ctx._source.title_es = params.title_es;
      ctx._source.content_es = params.content_es;
      ctx._source.description_es = params.description_es;
      ctx._source.language = "es";
      ctx._source.mime_type = "application/pdf";
      ctx._source.remove('title_en');
      ctx._source.remove('content_en');
      ctx._source.remove('description_en');
    """,
    "params": {
      "title_es": "Track your Spanish jayandra p VR&E benefits | Veterans Affairs",
      "content_es": "In this section Track your jayandra Spanish VR&E benefits Check the status of your Veteran Readiness and Employment (VR&E) application and benefits. Track your VR&E benefit status online now What you can do when you sign in Here’s what you can do when you sign in to track your VR&E benefit status online: Get updates on the status of your application Learn how to prepare for your initial evaluation with your counselor Explore career trackers, resources, and tools Who can track their VR&E benefit status online You can track your VR&E benefit status online if you’ve submitted an Application for Veteran Readiness and Employment for Claimants with Service-Connected Disabilities (VA Form 28-1900). If you haven’t submitted your application yet, you can learn more about how to apply for VR&E. Learn how to apply for Veteran Readiness and Employment More resources about VR&E benefits Last updated: July 2, 2026",
[...]

PR #2173, opened 2026-09-17
SRCH-6637: Added multi-match instead of common search for English language search

jayandra:
#### Problem:
- English affiliates enumerated a `common` clause per locale and text field, which is expensive and failed on search term  more than 5 words
#### Solution:
- English (unquoted) searches now use a single `multi_match` on `title_*`, `description_*`, and `content_*` with `minimum_should_match: '3<90%'`.
 
### Localhost QA:
Reran the localhost test mentioned in #2129 searching for "_veteran jayandra readiness Employment Spanish application_", and it did give me matching results without erroring out.


DK1999:
@selfdanielj Take a look!

selfdanielj:
@jayandra any idea what this empty `should` is doing there?  its in both `en` and `es` queries.  

Is it something that could be filled in by some other query option like domain_name? or is it always empty?
image

selfdanielj:
@jayandra  what do you think about making the multi_match the standard way of doing it for all languages?  The difference would be on english we search `["title_*", "description_*", "content_*"]` while for other languages we use the localized versions, e.g. `["title_es", "description_es", "content_es"]`. 

Or do you think thats out of scope and should be added as another story?

selfdanielj (commented):
Confirmed no errors locally with a query like `this is a very very very long query and I just want to see how long I can make it wow it can be really long`... left a few separte questions.  Otherwise looks good to me

DK1999:
Glad to see that very long quite long query working well. Butting in on the Dan's scope question... since we're already cracking into this query, let's just see how much we can improve it now.

jayandra:
@selfdanielj Thank you for digging into the blank should block and discovering that it’s blank when the affiliate doesn’t have any domains listed in `/sites/<affiliate>/domains`. In such cases, the search results aren’t restricted to that domain and instead look for matches everywhere, which is very similar to setting `Gets results from all domains` in the admin panel.

I’ve replaced the `common` query with `multi_match`. For English affiliates, it does a wildcard search across all fields, whereas for non-English affiliates, it continues to scope the search to their locale-specific fields only.

DK1999:
Thanks, Jayandra.
Dan, do you know why the "Gets results from all domains" selection needed to exist and what it does?
Is there any reason why it's not ok to allow affiliates with no domain value to search everything?

selfdanielj (approved):
Looks good to me!  Verified `es` and `en` both use `multi-match` now

selfdanielj:
> Thanks, Jayandra. Dan, do you know why the "Gets results from all domains" selection needed to exist and what it does? Is there any reason why it's not ok to allow affiliates with no domain value to search everything?

@DK1999 I think we talked about this in slack right?  Do you have any more outstanding questions on this?

PR #2178, opened 2026-09-18
SRCH-6782: Removed primary and secondary footer section when no Footer Link has been added

jayandra:
## Summary
When no footer link has been added in Admin Center → Manage Display → Visual Design, the primary and secondary footer sections are no longer displayed. Only the Return to top link appears.

Displaying footer links when they exist is unchanged.

## QA in Localhost:
Following screenshot shows old and new HTML structure. The primary and secondary footer sections are missing when there is no footer link.
Screenshot 2026-09-18 at 3 20 10 PM


selfdanielj (approved):
Confirmed on updated SERP that the extra divs are not present when no footer is selected and that one or more footers do display when put into the customer admin center

